PR #2, opened 2019-10-11
Basic Blacklight install

awead:
Install Blacklight and get a basic page layout rendering. This includes solr, but we don't have to run it in order for the application to work. The `catalog` page will throw an error if solr isn't running, but that's safe to ignore for now.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,3 +13,8 @@ AllCops:
     - 'vendor/**/*'
     - 'bin/**/*'
     - 'node_modules/**/*'
+
+Metrics/BlockLength:
+  Exclude:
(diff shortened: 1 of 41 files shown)


awead:
@rschenk Right, so, pipeline for sass/css stylesheets, and webpacker for javascript. It's actually what Rails 6 is doing by default. The asset pipeline is still there, it just only only has images and stylesheets.

I'm not sure how I feel about this, mainly because there are going to be javascript dependencies, like Uppy, for example, that will have their own stylesheets and might ought to go in webpack along with the js code. For now, we can use this pattern and switch it around if we need to later.

rschenk:
@awead I think that's a good approach.

PR #17, opened 2019-10-14
Specify Node version 12.x

rschenk:
When I tried to `yarn install` I got the following error:

> error eslint@6.5.1: The engine "node" is incompatible with this module. Expected version "^8.10.0 || ^10.13.0 || >=11.10.1". Got "11.8.0"

Looks to me like we need a `.node-version` file in the same way that we have a `.ruby-version` etc. There exists a node version manager for [asdf](https://github.com/asdf-vm/asdf-nodejs) (what @awead uses) as well as [fnm](https://github.com/Schniz/fnm) (what @rschenk uses). I dunno what @whereismyjetpack uses, but I'm guessing it involves epic and mystical sed commands piped through each other like the unix equivalent of a Persian rug. 

For starters, I just popped the latest major version of Node (that is, 12) into the `.node-version` file and that was fine for both fnm and the yarn install. Could one of you test this out with your respective version manager / sed-foo and see if that's acceptable to you?


whereismyjetpack:
looks good to me, I tend to do all my node stuff in docker so it's whatever is in the Dockerfile :)

awead:
Nice! @rschenk fixes #17

PR #18, opened 2019-10-15
WIP: db model ports

rschenk:
Can you check over what I've got so far with factory bot and if it's in-line with how y'all normally do things? Specifically linting the factories vs putting an expectation in the spec? I prefer the latter because linting all the factories before the test run can prevent the tests from running if a factory is invalid—which sometimes is _undesirable_ if you're actively working on validations and associations


rschenk:
Adam and I talked about this over chat, no need for this PR anymore

PR #19, opened 2019-10-15
Adding Code Climate reporting to the CI build

awead:
This replaces the existing rspec command in Drone CI with a `bin/ci` command that wraps the rspec test with Code Climate executeable. It is downloaded from Code Climate and then used to send coverage reports which can be viewed in the context if a PR (using a browser plugin) and with Github integration, it will block any PRs that have a significant drop in coverage.


.drone.yml
```diff
@@ -52,7 +54,7 @@ steps:
   <<: *test_common
   commands:
     - cd /app
-    - bundle exec rspec
+    - bin/ci
```
whereismyjetpack: I like this pattern, but would almost think we should rename this to ./bin/ci-rspec so we can perhaps drop more complex things in ./bin for testing, or building perhaps?

PR #21, opened 2019-10-16
Solr configurator

### LINKED ISSUE #9: Solr Configs
At this point going to stick with Solr because of it's easier integration with Blacklight

@whereismyjetpack  is going to start adding solr manifests to understand how we want to deploy Solr
-initialization (should be a core)
-APIs or config files -- for schema and configuration -- how do we want to manage those.

There is a new Blacklight version to look into as well. We are likely going to use the Blacklight XML config so we aren't off on our own doing something. Since version 7 of Solr, they (solr) are pushing the config API to use, but online, people using Solr are suggesting to turn on legacy mode and upload the XML file instead and not using the API for configs. Start with the Blacklight config and tweak as needed.

whereismyjetpack:
Fixes #9 

Adds solr into docker-compose and runs in cloud mode. 

Introduces new configurables `SOLR_HOST`, `SOLR_PORT`, `SOLR_COLLECTION`

Introduces a rake task for initalizing solr. 
`bundle exec solr:init`

The rake task will create a collection if it doesn't exist, upload the configset if it doesn't exist, and bind that configset to the collection. 

We must be running solr in cloud mode for these apis to work. `solr-foreground -DzkRun` the docker-compose file is setup to do this for us. We also assume that solr API is unauthenticated, and don't handle use cases where we would need to authenticate to the api. historically we have relied on host based firewalls to protect that endpoint.


lib/solr_configurator.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    def raw_data
+      @raw_data ||= zip_file.read
+    end
+
+    # Gets a response object, if it's status code is not 200, we emit the body and bail
+    # This can be better! PRs accepted.
+    def check_resp(resp)
+      if resp.status == 200
```
awead: raise StandardError unless resp.status == 200

Dockerfile
```diff
@@ -31,22 +31,25 @@ RUN curl -Lo /tmp/envconsul.zip https://releases.hashicorp.com/envconsul/0.9.0/e
...
 COPY package.json yarn.lock /app/
 RUN gem install bundler
+RUN yarn
 
 RUN useradd -u 10000 app -d /app
 RUN chown -R app /app
 USER app
+ENV BUNDLE_PATH=vendor/bundle
 
+# RUN --mount=type=cache,id=gem,sharing=locked,uid=10000,target=/app/vendor bundle package
```
awead: @whereismyjetpack can you remove any extraneous comments? It looks like these were part of some testing process.

lib/solr_configurator.rb
```diff
@@ -0,0 +1,129 @@
...
+  def initialize
+    connection
+  end
+
+  def zip_file
+    File.open(tempfile)
+  end
+
+  def configset_exists?
+    resp = @connection.get('/solr/admin/configs?action=LIST&omitHeader=true')
```
rschenk: I would avoid using `@connection` entirely in this class, and instead call the private method `connection`. That way there's no chance of a dependency or race condition between the private method being called first (in this case in the constructor), setting the instance variable, and then subsequent methods calling it. Plus what if in the future you needed to do something fancy in that method beyond just setting a single variable? By using the private method you're basically abstracting the implementation away.

rschenk: Faraday can accept a hash argument to build the URL parameters for you. This has an advantage of making sure that everything is properly URL escaped. Ditto goes for headers. This will also probably make tests easier to write with webmock or similar.

Personally I would change
```
resp = @connection.get('/solr/admin/configs?action=LIST&omitHeader=true')
```

To
```
resp = @connection.get('/solr/admin/configs', {action: 'LIST', omitHeader: true})
```

lib/solr_configurator.rb
```diff
@@ -0,0 +1,129 @@
...
+      return true
+    end
+
+    resp = @connection.get("/solr/admin/collections?action=CREATE&name=#{collection_name}&numShards=#{num_shards}")
+    check_resp(resp)
+  end
+
+  def modify_collection
+    puts 'Modifying Collection to use Configset'
+    uri = %W(
```
rschenk: If using faraday's hash-as-get-params feature, this multi-line string goes away and becomes much more readable

lib/solr_configurator.rb
```diff
@@ -0,0 +1,129 @@
...
+      /solr/admin/collections
+      ?action=MODIFYCOLLECTION
+      &collection=#{collection_name}
+      &collection.configName=#{@configset_name}
+    ).join
+    resp = @connection.get(uri)
+    check_resp(resp)
+  end
+
+  def upload_config
```
rschenk: To me, all these puts statements below seem to mix the concerns of this class with the concerns of the script that will be running it. What if I wanted to use this class outside the context of a rake task / script? What if this were part of some kind of web UI for configuring Solr instances instead?

I could see an argument for moving most of this logic into the rake task itself instead, since `configset_exists?` is a private method, and that would be a more appropriate . The main plus for keeping this logic together in the method is, it makes it easier to test the idempotency...except that there currently is no test for that.

lib/solr_configurator.rb
```diff
@@ -0,0 +1,129 @@
...
+      @configset_name ||= "#{collection_name}-#{md5}"
+    end
+
+    def raw_data
+      @raw_data ||= zip_file.read
+    end
+
+    # Gets a response object, if it's status code is not 200, we emit the body and bail
+    # This can be better! PRs accepted.
+    def check_resp(resp)
```
rschenk:
```
raise resp.body unless resp.status == 200
```

lib/solr_configurator.rb
whereismyjetpack: doh! the lightbulb just went off. ok. I can certainly fix that.

rschenk: That's a pattern I learned from Sandi Metz 👍

rschenk: Exactly. Your `#initialize` method can go away

lib/solr_configurator.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+require 'zip'
+
+class SolrConfigurator
+  def initialize; end
```
rschenk: You can actually remove this empty initialize method entirely, it'll be inherited from `Object`. It's fine to leave it here, it won't do any harm, but it's not necessary and the Ruby idiom would be to remove it

config/blacklight.yml
```diff
@@ -1,9 +1,23 @@
 development:
```
whereismyjetpack: blacklight config loader doesn't support YAML anchors. we should maybe try to fix that upstream.

whereismyjetpack:
ITS GREEEEEN?!??!

awead:
LGTM. Pending @rschenk's approval, you can do a "squash and merge"

PR #23, opened 2019-10-16
Bring over database models from ss-json

rschenk:
Ports over the db models and specs from the ss-json repo, and also adds FactoryBot factories along the way. Having FactoryBot will make it easier to flesh out some of the tests, which are lacking—I suspect our code coverage metrics on this PR will be a bit low.

### DIFF
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -4,7 +4,7 @@ FactoryBot.define do
   factory :work_version do
     work
     version_name { '1' }
-    aasm_state { 'draft' }
+    aasm_state { WorkVersion::STATE_DRAFT }
     title { generate(:work_title) }
   end
 end
--- a/spec/factories/works.rb
+++ b/spec/factories/works.rb
@@ -13,5 +13,12 @@ FactoryBot.define do
     end
   end
 
-  sequence(:work_title) { |n| "Work Version #{n}" }
+  sequence(:work_title) do |n|
+    a = Faker::Lorem.words(rand(1..5)).join(' ').capitalize
+    b = Faker::Lorem.words.join(', ')
+    c = Faker::Lorem.words.join("'s ")
+    id = n.ordinalize
+
+    %(#{a}: "#{b}; #{id} #{c}")
+  end
 end


app/models/work.rb
```diff
@@ -0,0 +1,57 @@
...
+class Work < ApplicationRecord
+  belongs_to :depositor,
+             class_name: 'User',
+             foreign_key: 'depositor_id',
+             inverse_of: 'works'
+
+  has_many :work_creations,
+           dependent: :restrict_with_exception
+
+  has_many :aliases,
```
awead: @rschenk seeing your other comment in the factory, it occurred to me that alias/aliases could pose problems. Do we want to rename this?

spec/factories/work_versions.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :work_version do
+    work
+    version_name { '1' }
+    aasm_state { 'draft' }
+    title { generate(:work_title) }
```
awead: @rschenk is this a FactoryBot thing? I was half-hoping for a custom method that creates a rando title via Faker. If you need inspiration https://github.com/psu-libraries/cho/blob/master/spec/support/metadata_factory.rb#L7

Gemfile
```diff
@@ -5,21 +5,26 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
 gem 'rails', '~> 6.0.0'
 gem 'rsolr', '>= 1.0', '< 3'
 gem 'sass-rails', '~> 5'
+gem 'shrine', '~> 3.0'
 gem 'webpacker', '~> 4.0'
 
 group :development, :test do
   gem 'byebug', platform: :mri
+  gem 'factory_bot_rails'
+  gem 'faker'
```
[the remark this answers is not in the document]
rschenk: Oh oops, this must be a mistake. I'll fix this up before merging, I was intending to switch over from `ffaker` to normal `faker`

spec/factories/work_versions.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :work_version do
+    work
+    version_name { '1' }
+    aasm_state { 'draft' }
```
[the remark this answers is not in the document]
rschenk: Oh right! Forgot about that!

app/models/work.rb
rschenk: It's only a problem in the factory, where `alias` could not be used as a bare word. I think we're good here and no need to rename

spec/factories/work_versions.rb
rschenk: This is a FactoryBot thing that generates a title that's guaranteed to be unique, but it's true that the titles will be super generic "Work Title 1", "Work Title 2" etc. I very much like them to be guaranteed to be unique for integration/feature tests, but I could combine the approaches...rando Faker title that includes a unique id somewhere in it

PR #24, opened 2019-10-17
CI Improvements

whereismyjetpack:
- switches to multi-stage builds for baking test targets
- run test targets on those stages
- parallelize testing stages

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -62,14 +62,13 @@ steps:
     - export MINIO_HOST=$(dig +short minio)
     - docker run
         --net host
-        -e POSTGRES_PASSWORD=scholarsphere
-        -e AWS_BUCKET=scholarsphere
         -e S3_ENDPOINT=http://$MINIO_HOST:9000
         -e AWS_BUCKET=scholarsphere
         -e AWS_ACCESS_KEY_ID=scholarsphere
         -e AWS_SECRET_ACCESS_KEY=scholarsphere
         -e AWS_REGION=us-east-1
         -e POSTGRES_USER=scholarsphere
+        -e POSTGRES_PASSWORD=scholarsphere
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -22,41 +27,82 @@ steps:
...
+    - export DOCKER_HOST=unix:///var/run/host.sock
+    - apk add bind-tools
+    - export POSTGRES_HOST=$(dig +short postgres)
+    - export MINIO_HOST=$(dig +short minio)
+    - docker run
+        --net host
+        -e POSTGRES_PASSWORD=scholarsphere
+        -e AWS_BUCKET=scholarsphere
+        -e S3_ENDPOINT=http://$MINIO_HOST:9000
+        -e AWS_BUCKET=scholarsphere
```
awead: duplicate line.

.drone.yml
```diff
@@ -22,41 +27,82 @@ steps:
...
-    - bundle exec rails db:create
-    - bundle exec rails db:migrate
-    - bin/ci-rspec
+    - export DOCKER_HOST=unix:///var/run/host.sock
+    - apk add bind-tools
+    - export POSTGRES_HOST=$(dig +short postgres)
+    - export MINIO_HOST=$(dig +short minio)
+    - docker run
+        --net host
+        -e POSTGRES_PASSWORD=scholarsphere
```
awead: should this be hidden or do we care? I'm not sure I do :)

.drone.yml
whereismyjetpack: good eye, sniper.

.drone.yml
whereismyjetpack: it's ephemerial, i don't think we care.

awead: since the host name is hidden this probably doesn't matter?

PR #31, opened 2019-10-24
Oauth integration

rschenk:
This was mostly paired with @whereismyjetpack and myself. It integrates our app with the PSU-OAuth service that we also paired on

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -32,8 +32,6 @@ module Scholarsphere
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
-    config.eager_load_paths << Rails.root.join('lib')
-
     # Don't generate system test files.
     config.generators.system_tests = nil
   end
--- a/config/initializers/shrine.rb
+++ b/config/initializers/shrine.rb
@@ -4,6 +4,8 @@ require 'shrine'
 require 'shrine/storage/s3'
 require 'uppy/s3_multipart'
 
+require 'scholarsphere/shrine_config'
+
 Shrine.storages = Scholarsphere::ShrineConfig.storages
 
 Shrine.plugin :activerecord
--- a/lib/omniauth/strategies/psu.rb
+++ b/lib/omniauth/strategies/psu.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'omniauth/strategies/oauth2'
+
 module OmniAuth
   module Strategies
     class Psu < OmniAuth::Strategies::OAuth2
--- a/spec/lib/scholarsphere/shrine_config_spec.rb
+++ b/spec/lib/scholarsphere/shrine_config_spec.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
+require 'spec_helper'
+
+require_relative '../../../lib/scholarsphere/shrine_config'
 
 RSpec.describe Scholarsphere::ShrineConfig do
   subject { described_class }
--- a/spec/lib/solr_configurator_spec.rb
+++ b/spec/lib/solr_configurator_spec.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
+require 'spec_helper'
+
+require_relative '../../lib/solr_configurator'
 
 RSpec.describe SolrConfigurator do
   describe '#zip_file' do


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,17 @@
...
+    new_user_session_path
+  end
+
+  # TODO Blacklight's default configuration/views require this path helper to be
+  # present. Since we are using OAuth exclusively and are not currently using
+  # Devise database_authenticatable, this path helper does not exist, causing
+  # Blacklight to freak out.
+  def edit_user_registration_path
+    '/'
+  end
```
awead: @rschenk @whereismyjetpack this is used in Blacklight's included view at `app/views/shared/_user_util_links.html.erb`. I would override that partial locally to remove the method call. We're going to do this anyway, so now's an appropriate time to do it.

app/models/user.rb
```diff
@@ -13,4 +15,24 @@ class User < ApplicationRecord
...
+    []
+  end
+
+  # TODO blacklight wants to use this for the User's name in the header??
+  def to_s
+    access_id
+  end
+
+  # TODO Which fields do we want updated each time a user logs in?
+  #   - Group memberships
```
awead: @rschenk @whereismyjetpack We update a user's groups in Scholarsphere every 24 hours. We'll need to ticket this unless you want to go ahead and add it now.

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# https://github.com/plataformatec/devise/wiki/OmniAuth:-Overview
+#
+class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
```
awead: @rschenk @whereismyjetpack None of this class is showing as tested. Do you have unit tests to cover the cases?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  def psu
+    @user = User.from_omniauth(request.env['omniauth.auth'])
+
+    if @user.persisted?
+      sign_in_and_redirect @user, event: :authentication
+      set_flash_message(:notice, :success, kind: 'Penn State') if is_navigational_format?
+    else
+      # TODO if OAuth fails, where do we went users?
```
awead: @rschenk @whereismyjetpack ticket?

spec/helpers/application_helper_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  # Required by Devise when not using database_authenticatable
+  describe '#new_session_path' do
+    subject { helper.new_session_path }
+
+    it { is_expected.to eq(new_user_session_path) }
+  end
+
+  # Required to exist by default Blacklight. Does not exist because we're not
+  # using database_authenticatable
+  describe '#edit_user_registration_path' do
```
awead: @rschenk @whereismyjetpack we won't need the test if override the view (see above)

spec/models/user_spec.rb
```diff
@@ -24,6 +24,76 @@
...
+    subject { described_class.new.bookmarks }
+
+    it { is_expected.to eq [] }
+  end
+
+  # Blacklight uses this method in the navigation header. I don't like this
+  describe '#to_s' do
+    subject { described_class.new(access_id: 'abc123').to_s }
+
+    it { is_expected.to eq 'abc123' }
```
awead: @rschenk @whereismyjetpack we shouldn't need these two tests if just run the generator that adds the Blacklight user module. We could have a `expect(user).to respond_to(:bookmarks) }` 🤷‍♂

app/controllers/users/omniauth_callbacks_controller.rb
rschenk: We didn't write this code, it came straight from Devise and was hard af to test, so I didn't. But I can try to figure it out!

app/models/user.rb
rschenk: Does that mean that Scholarsphere will need to talk to LDAP itself, outside of the OAuth flow? It seems that way, from the above comment. Maybe we could somehow use the OAuth server and re-ping the user resource? We should ticket this

app/controllers/users/omniauth_callbacks_controller.rb
rschenk: Ticketed as Issue #33

PR #39, opened 2019-10-29
Work factory builds versions better

rschenk:
I decided to remove the code from the `Work` model that automatically builds a `WorkVersion` if none exist. This functionality would only be used in one place anyways, and it was causing more confusion than I thought it was worth. I updated the work factory accordingly, which now easily lets you specify the number of versions you'd like it to build, and whether it should include a draft.


app/models/work.rb
```diff
@@ -48,6 +46,9 @@ def self.options_for_select_box
             presence: true,
             inclusion: { in: Types.all }
 
+  validates :versions,
+            presence: true
+
```
awead: @rschenk So I assume that the UI will create the initial work and its version? If we do this outside of the UI, we'd just do `work.versions.build` or some such?

PR #42, opened 2019-10-29
Indexing Service

awead:
Creates an indexing service and uses it to index WorkVersion resources into Solr.

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -40,6 +40,8 @@ class WorkVersion < ApplicationRecord
             acceptance: true,
             if: :published?
 
+  after_save :update_index
+
   aasm do
     state :draft, intial: true
     state :published, :withdrawn, :removed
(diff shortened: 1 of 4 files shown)


spec/services/indexing_service_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      it 'indexes a minimal amount of metadata in Solr' do
+        service
+        document = SolrDocument.find(resource.uuid)
+        expect(document.id).to eq(resource.uuid)
+        expect(document['model_ssi']).to eq('IndexedResource')
+      end
+    end
+
+    context 'with an improperly identified resource' do
+      let(:resource) { Struct.new('UnsupportedResource') }
```
rschenk: Just FYI, `Struct.new` returns back a constant, not an instance.

Ideas to make this look more like a "resource" object...

```
Struct.new('UnsupportedResource').new
BasicObject.new
mock('UnsupportedResource')
```

PR #44, opened 2019-10-30
Use bin/minio for starting up minio

awead:
Creates a minio executable to start the minio server and hooks into our application.yml file for the config options.

### DIFF
--- a/bin/minio
+++ b/bin/minio
@@ -1,0 +1,36 @@
+#!/usr/bin/env ruby
+require 'fileutils'
+require 'figaro'
+
+APP_ROOT = File.expand_path('..', __dir__)
+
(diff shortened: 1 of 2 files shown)


rschenk:
Ah, I see, I didn't realize the docker-compose one didn't use figaro/env-variables. In that case :shipit:

whereismyjetpack:
docker-compose has it in the environment variables that it sends to each of the services; they should all have sane defaults and "just work" binstubs might be a good idea; but each developer may want a slightly different thing though. for example my workflow is 

`docker-compose up -d solr db minio` and then `bundle exec rails s` in another window. others may do something like `docker-compose up -d solr`, or whatever. I'd be more than happy to talk about all and anything too!

PR #45, opened 2019-10-31
[WIP] Start dependencies

whereismyjetpack:
This is a work in progress, a small wrapper around `docker-compose` that pulls in config from `figaro` and then runs the docker compose up phase for _only_ minio and solr. 

TODO:

- a developer has to run `docker-compose down` in `APP_ROOT` when they are done working. I don't think or know if this should be another binstub or not

- if a developer wants to run database or rails server within docker this script isn't very useful to them. they can, however, run the commands by hand

- we don't check if docker or docker-compose is installed. I don't know how clever this script should be, with cleverness comes complexity and what not.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -1,12 +1,41 @@
 #!/usr/bin/env ruby
+# frozen_string_literal: true.
 require 'fileutils'
+require 'figaro'
+require 'net/http'
+require 'uri'
 
 # path to your application root.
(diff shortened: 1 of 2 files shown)


awead:
@whereismyjetpack what's the status on this PR? Do you still need it?

whereismyjetpack:
we can throw it away

PR #47, opened 2019-10-31
Removing profile option

awead:
The profile option was added to our config back when we were having really slow test problems. Now that it isn't an issue anymore, the tag is generating a lot of extra output each time tests are run. This creates unnecessary background "noise" so let's turn it off and add it when we need to really profile the test suite.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,4 +1,3 @@
 --color
 --format documentation
---profile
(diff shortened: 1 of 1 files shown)


rschenk:
And there was much rejoicing

PR #49, opened 2019-11-04
Docker fixes

whereismyjetpack:
Lints the `Dockerfile` 
Adds missing configurations to web container
remove server.pid as part of container startup if it exists.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,12 +6,14 @@ ENV TZ=America/New_York
 
 # Install system deps
 RUN apt-get update && \
-    apt-get install ffmpeg -y 
+    apt-get install --no-install-recommends ffmpeg=7:4.1.4-1~deb10u1 -y && \
+    rm -rf /var/lib/apt/lists/* 
 
 ## NodeJS
 ENV NODE_VERSION 12.9.1
 RUN mkdir /usr/local/nvm
 ENV NVM_DIR /usr/local/nvm
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
 ENV NODE_PATH $NVM_DIR/v$NODE_VERSION/lib/node_modules
 ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -6,12 +6,14 @@ ENV TZ=America/New_York
 
 # Install system deps
 RUN apt-get update && \
-    apt-get install ffmpeg -y 
+    apt-get install --no-install-recommends ffmpeg=7:4.1.4-1~deb10u1 -y && \
```
awead: Why are we installing ffmpeg? We're not relying on it for anything at the moment, as far as I know.

Dockerfile
```diff
@@ -22,7 +24,7 @@ RUN . $NVM_DIR/nvm.sh \
     && nvm alias default $NODE_VERSION \
     && nvm use default
 
-RUN npm install -g yarn
+RUN npm install -g yarn@1.19.1
```
awead: What happens as our local machines' yarn versions get out of sync with this one?

Dockerfile
```diff
@@ -34,8 +36,8 @@ RUN chown -R app /app
 USER app
 
 COPY Gemfile Gemfile.lock /app/
-RUN gem install bundler
-RUN bundle install --frozen --path vendor/bundle
+RUN gem install bundler:2.0.2
```
awead: Ditto, as per my yarn comment. The Gemfile.lock could specify a new version of bundler and the install might fail. Is there away to pick the bundler version from the lockfile?

start.sh
```diff
@@ -5,6 +5,9 @@ if [ ${APP_ROLE:-app} == "sidekiq" ]; then
     bundle exec sidekiq
 else
     echo "starting rails"
+    if [ -f tmp/pids/server.pid ]; then
+      rm tmp/pids/server.pid
```
awead: `rm -f tmp/pids/server.pid` will silently fail if the file doesn't exist, then you don't need the `if`

Dockerfile
whereismyjetpack: I wasn't sure if we'd need it for the background jobs stuff, its a holdover from the ss-shrine repo

Dockerfile
whereismyjetpack: This gives us a kind of chicken and egg situation, where we need bundler to install from the lock file. If bundler fails, we should issue a PR to the Dockerfile to fix it?

Dockerfile
whereismyjetpack: I think we issue a PR to update the Dockerfile when our versions change. 

Motivation comes from 12 factor app (https://12factor.net/dependencies) and we should explicitly mark our versions in the docker file so that stale versions don't get cached, or updated out from under us.

Dockerfile
whereismyjetpack: I agree. I'll remove that. The docker optimizations are ever-evolving right now 🚀

PR #55, opened 2019-11-06
CRUD for WorkVersions

rschenk:
@awead Can you please check out `app/views/dashboard/work_versions/_status_badge.html.erb`. Do you reckon it needs a presenter? Maybe? Eh?


app/views/dashboard/work_versions/_show_files.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if work_version.file_resources.present? %>
```
awead: @rschenk You could move the if statement into show.html.erb: `render 'show_files' [...] if work_version.file_resources.present?` and eliminate the nesting.

app/assets/stylesheets/work_versions.scss
```diff
@@ -0,0 +1,7 @@
+@import 'bootstrap';
+@import 'bootstrap/variables';
```
awead: @rschenk having to re-import these each time seems bad, but I'm guessing that ultimately it doesn't matter since it gets compiled down into one file anyway. Having different files named according to the partial they're styling is nice, but I guess the trade off is these duplicated import statements.

app/views/dashboard/work_versions/_show_fields.html.erb
```diff
@@ -0,0 +1,74 @@
+<p>
+  <strong><%= WorkVersion.human_attribute_name('version_name') %>:</strong>
+  <%= work_version.version_name %>
+</p>
```
awead: @rschenk extra credit if you want to reformat these in a more bootstrap-ish way, such as a definition list. Up to you though...

app/views/dashboard/work_versions/show.html.erb
```diff
@@ -0,0 +1,23 @@
+<nav aria-label="breadcrumb">
+  <ol class="breadcrumb">
+    <li class="breadcrumb-item"><%= link_to 'Your ScholarSphere Deposits', dashboard_works_path %></li>
```
awead: @rschenk we should i18n-ify the link text.

app/assets/stylesheets/work_versions.scss
rschenk: Yeah I have no idea if Sass is smart enough to compile these all down. I'll have a look

app/views/dashboard/work_versions/_show_fields.html.erb
rschenk: OnE gIaNt ViM mAcRo?!?!?!?

app/assets/stylesheets/work_versions.scss
awead: I think it is... at least when we precompile the assets in production mode.

awead:
@rschenk 
> @awead Can you please check out `app/views/dashboard/work_versions/_status_badge.html.erb`. Do you reckon it needs a presenter? Maybe? Eh?

Meh. I think a two-line view is good for now. We could refactor later if there are different kinds of badges, exc.

PR #74, opened 2019-11-14
Refactor JS feature specs

### LINKED ISSUE #67: Testing browser integration in DroneCI
Feature tests can use a headless browser for javascript and other integration testing. Currently, this is not implemented in our CI environment. Tests such as `publish_new_work_spec.rb` will only run in the local testing environment. We need to implement a reliable way of running these tests using a complete browser environment.

What Done Looks Like:
Remove the `unless: ci_build?` from any existing feature tests and the DroneCI build will pass.

rschenk:
Capybara has the notion of a javascript driver, and it's possible to set the default one to whatever you want. By registering our selenium_chrome_headless driver as the default one, then you can tag your specs with `js: true` as opposed to `with_driver: :selenium_chrome_headless` which is a heck of a lot easier to remember.

This also allowed me to conditionally switch between the `transaction` and `truncation` DatabaseCleaner strategy. Transaction is way faster but doesn't play nice with headless chrome; truncation is way slower but does play nice with headless chrome. By Swapping the two of them out, we get a big speed boost in the spec runtimes (_especially_ for non-feature specs, which comprise most of our tests!). See the lengthy commit message for more deets

Fixes #67

### DIFF
--- a/spec/support/capybara_drivers.rb
+++ b/spec/support/capybara_drivers.rb
@@ -1,46 +1,29 @@
 # frozen_string_literal: true
 
-module CapybaraDrivers
-  def sign_in(user: nil, driver:)
-    assign_current_driver(driver)
-    OmniAuth.config.test_mode = true
-    OmniAuth.config.mock_auth[:psu] = mock_auth_hash(user) if user.present?
-  end
-
-  private
-
-    def assign_current_driver(driver)
-      register_driver(driver) unless Capybara.drivers.include?(driver)
-      Capybara.current_driver = driver
-    end
-
-    def register_driver(driver)
-      Capybara.register_driver driver do |app|
-        capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(chromeOptions: chrome_options)
-        Capybara::Selenium::Driver.new(app, browser: :chrome, desired_capabilities: capabilities)
-      end
-    end
-
-    def chrome_options
-      { args: ['no-sandbox', 'disable-gpu', 'window-size=1024,768', 'single-process'] }
-    end
-
-    def mock_auth_hash(user)
-      build :psu_oauth_response, access_id: user.access_id
-    end
+RSpec.configure do |_config|
+  Capybara.javascript_driver = if ci_build?
+                                 :selenium_remote
+                               else
+                                 :selenium_chrome_headless
+                                 # :selenium_chrome # For debugging in dev
+                               end
 end
 
-RSpec.configure do |config|
-  config.before(type: :feature) do |example|
-    driver = example.metadata.fetch(:with_driver, Capybara.current_driver)
-
-    if example.metadata.key?(:with_user)
-      sign_in(user: send(example.metadata.fetch(:with_user)), driver: driver)
-    else
-      sign_in(driver: driver)
-    end
-  end
-
-  config.include CapybaraDrivers, type: :feature
-  config.include ActionView::RecordIdentifier, type: :feature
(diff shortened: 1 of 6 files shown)


whereismyjetpack:
I think this may conflict with some work I was doing to get selenium to work in CI. see #75. maybe we can pair tomorrow to get both methods working?

rschenk:
@whereismyjetpack I see what you did in #75, and this PR may actually make it easier. If we combine my PR with your PR, then all of the test can just use `js: true` and then we can swap the driver out in the Capybara config under the hood

```
Capybara.javascript_driver = if ci_build?
                               :crazy_docker_headless_mayhem
                             else
                               :selenium_chrome_headless 
                             end
```

awead:
@rschenk one of the things you could do in earlier setup was to use `with_driver: :selenium_chrome` and you'd get the non-headless version. Capybara includes four drivers by default: selenium, selenium_headless, selenium_chrome, and selenium_chrome_headless. It's helpful when debugging to add the non-headless one in there when you need it. How would you do that in this updated setup?

rschenk:
@awead When debugging, you could change it in your `Capybara.javascript_driver` line in the config

spec/support/capybara_drivers.rb
```diff
@@ -1,46 +1,36 @@
...
-      sign_in(driver: driver)
+      sign_in
     end
   end
 
   config.include CapybaraDrivers, type: :feature
   config.include ActionView::RecordIdentifier, type: :feature
 end
+
+Capybara.register_driver :selenium_chrome_headless do |app|
```
awead: @rschenk I'm not sure this block is even necessary. `: selenium_chrome_headless` is already present in the list of available drivers, so the only line that may be needed is line 36. This might change if we need to specify a window size. We did need to do this in some other tests to verify that UI elements were visible in a tablet and phone-sized page, but I'd rather see us put this in when they're needed. I'm also not convinced the other options actually do anything.

rschenk: Cool. In that case, I suppose we could always register our own custom drivers, like `:selenium_chrome_tablet`, `:selenium_chrome_phone`, etc etc

awead:
@rschenk we can simplify the driver registration even more. See https://github.com/psu-stewardship/scholarsphere-4/blob/capybara/spec/support/capybara_drivers.rb

spec/support/feature_helpers.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+end
+
+RSpec.configure do |config|
+  Capybara.javascript_driver = if ci_build?
+                                 :selenium_remote
+                               else
+                                 :selenium_chrome_headless
+                                 # :selenium_chrome # For debugging in dev
+                               end
```
awead: @rschenk I think this block could just get moved to to the CapybaraDrivers module, and then you wouldn't need the `require_relative` anymore.

rschenk: We don't actually need the `require_relative` at all, I just put it there as a little hint to where `:selenium_remote` was coming from

I guess I figured that the capybara_drivers file would only contain driver definitions, whereas the feature_helpers file would be for configuring how you wanted the feature specs to work?

awead: @rschenk I get that. The ontologist in me is saying put it in CapybaraDrivers because it _says_ Capybara 😄 I can go either way...

awead:
@rschenk @whereismyjetpack Alright, we've futzed with this enough. It's good. Thanks for the work!

PR #75, opened 2019-11-14
Adds Selenium to CI

whereismyjetpack:
allows user to optionally specify SELENIUM_URL, and APP_HOST variables to use a selenium_remote, will default to the default headless chrome if unset.

### DIFF
--- a/spec/features/publish_new_work_spec.rb
+++ b/spec/features/publish_new_work_spec.rb
@@ -2,12 +2,12 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Publishing a new work', unless: ci_build? do
+RSpec.describe 'Publishing a new work' do
   let(:user) { create(:user) }
   let(:metadata) { MetadataFactory.new.work_version }
   let(:updated_metadata) { MetadataFactory.new.work_version }
 
-  it 'routes the user through the workflow', with_user: :user, with_driver: :selenium_chrome_headless do
+  it 'routes the user through the workflow', with_user: :user, with_driver: :selenium_remote do
     visit(new_dashboard_work_path)
     fill_in('Title', with: metadata[:title])
     click_button('Create Work')
--- a/spec/features/publish_new_work_spec.rb
+++ b/spec/features/publish_new_work_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe 'Publishing a new work' do
   let(:metadata) { MetadataFactory.new.work_version }
   let(:updated_metadata) { MetadataFactory.new.work_version }
 
-  it 'routes the user through the workflow', with_user: :user, js: true  do
+  it 'routes the user through the workflow', with_user: :user, js: true do
     visit(new_dashboard_work_path)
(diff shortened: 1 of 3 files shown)


spec/features/publish_new_work_spec.rb
```diff
@@ -2,12 +2,12 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Publishing a new work', unless: ci_build? do
+RSpec.describe 'Publishing a new work' do
   let(:user) { create(:user) }
   let(:metadata) { MetadataFactory.new.work_version }
   let(:updated_metadata) { MetadataFactory.new.work_version }
 
-  it 'routes the user through the workflow', with_user: :user, with_driver: :selenium_chrome_headless do
```
whereismyjetpack: I had to rename this, since `:selenium_chrome_headless` was already registered, but with only the default options

spec/support/capybara_drivers.rb
```diff
@@ -17,12 +17,24 @@ def assign_current_driver(driver)
...
+          Capybara::Selenium::Driver.new(app,
+                                         browser: :chrome,
+                                         desired_capabilities: capabilities)
+        end
       end
     end
 
     def chrome_options
-      { args: ['no-sandbox', 'disable-gpu', 'window-size=1024,768', 'single-process'] }
+      { args: ['headless', 'no-sandbox', 'disable-gpu', 'window-size=1024,768', 'single-process'] }
```
awead: @whereismyjetpack does 'headless' actually work? If you take it out, does it _not_ run headless?

spec/features/publish_new_work_spec.rb
```diff
@@ -2,12 +2,12 @@
...
+  it 'routes the user through the workflow', with_user: :user, with_driver: :selenium_remote do
```
awead: @whereismyjetpack see @rschenk's other comment, but the option here should be that we use a local driver, and ideally have the option to specific different kinds of drivers for local debugging, etc. But when we're in a ci build, we'd have the test use `:selenium_remote`

spec/support/capybara_drivers.rb
whereismyjetpack: I seem to remember needing to put this in. My memory is a bit hazy right now though :)

spec/features/publish_new_work_spec.rb
whereismyjetpack: this is a bad name. i should mend that. i think what would happen is it would register a local driver if you don't have the `SELENIUM_URL` set. I think we should maybe wrap that all up in the `ci_build?` boolean and 🎂

spec/support/capybara_drivers.rb
whereismyjetpack: https://drone-test.dsrd.libraries.psu.edu/psu-stewardship/scholarsphere-4/392/1/9 taking headless out does, indeed fail it.

whereismyjetpack:
closing in favor of #74 ; and then I'll add `:selenium_remote`

PR #76, opened 2019-11-18
Making fields multiple

### LINKED ISSUE #77: Enable multiple field inputs
In matching with the existing Scholarsphere application, there should be multiple fields on our work versions according to the API specification.

awead:
Updates our work version schema to define fields as multiple according to the current Scholarsphere API schema.

Includes javascript features for adding and removing additional fields in the form, and refactors our form to render fields using different partials. This is butting up against a form presenter pattern, so we could discuss switching over to SimpleForm or some other type of form management solution.

Fixes #77 
Screen Shot 2019-11-19 at 8 49 15 AM


rschenk:
@awead Looks good, but I pushed some changes for you to review.

1. I reformatted some syntactic vinegar from the WorkVersionsController
2. I changed the naming of your field partials... 
    * `_default` didn't tell me anything about what the field did, so I renamed it to `_multi_text`
    * `_text_area` was actually a multiple text area. If we ever needed to show a non-multiple text area, we'd need to rename this partial, so I went ahead and named it `_multi_text_area`

PR #79, opened 2019-11-20
Complete Catalog Fields

awead:
Adds the rest of the fields to the CatalogController and refactors our metadata factories.


awead:
@rschenk one of the reasons is so we can create a different batch of fields and update an existing record with them. See https://github.com/psu-stewardship/scholarsphere-4/blob/catalog-fields/spec/features/edit_multiple_fields_spec.rb#L8-L9 I need to have an object with the metadata. But, I see your point. Is there a way to put the methods that generate the random metadata in the factory so that we can call them directly, and get just the metadata back, without having to build a whole new object?

rschenk:
@awead FactoryBot has a method for such a purpose: `attributes_for`

PR #84, opened 2019-11-21
Audit works

rschenk:
Shows a timeline of the edits to a particular Work. 

So far we only deal with WorkVersion's metadata. Edits to files are forthcoming and not in this PR.

Implements a large chunk of #7


app/controllers/dashboard/work_histories_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Dashboard
+  class WorkHistoriesController < BaseController
+    def show
+      @work = current_user.works.find(params[:work_id])
+      @work_history = WorkHistoryPresenter.new(@work)
+      @latest_work_version = WorkVersionDecorator.new(@work.latest_version)
```
awead: @rschenk I don't know if this matters, but `@work` doesn't seem to be used in the views anywhere. I'm also wondering if the WorkHistoryPresenter might have a `latest` method that produces the `@latest_work_version`. That would cut down the number of instance variables the controller is having to set.

app/presenters/work_history_presenter.rb
```diff
@@ -0,0 +1,68 @@
...
+    def find_user(user_id)
+      User.find(user_id)
+    rescue ActiveRecord::RecordNotFound
+      null_user
+    end
+
+    # Null Object Pattern here used to prevent "undefined method x for nil" in
+    # the view
+    def null_user
+      User.new
```
awead: @rschenk This is a purely academic remark: what about a user class that cannot be saved, such as a "readonly" instance of user to guard against the really-never-gonna-happen-so-why-even-bother case where somehow `null_user` tries to get saved.

app/presenters/work_history_presenter.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    attr_reader :user_lookup_cache
```
awead: @rschenk I'm not used to seeing private attr_readers. Any reason why it's at the very bottom? I wonder if it's more legible to put it right after `private`.

app/presenters/work_version_change_presenter.rb
```diff
@@ -0,0 +1,74 @@
...
+    paper_trail_version.event == 'create'
+  end
+
+  def update?
+    paper_trail_version.event == 'update' && !publish?
+  end
+
+  def publish?
+    paper_trail_version.event == 'update' &&
+      paper_trail_version.object_changes.fetch('aasm_state', []).last.to_s == WorkVersion::STATE_PUBLISHED.to_s
```
awead: @rschenk I wonder if the `to_s`s are really necessary here.

app/services/work_version_change_diff.rb
```diff
@@ -0,0 +1,31 @@
...
+# change in a PaperTrail::Version object. Passing that PaperTrail::Version to
+# this class will return the differences in metadata of that change.
+#
+# PaperTrail _does_ have the ability to perform its own diffs of changes, and
+# therefore one might think that this class should be unecessary. Unfortunately
+# that feature of Papertrail is thwarted by our use of a jsonb column to store
+# the metadata.
+#
+# @example Given a PaperTrail::Version of a WorkVersion where title was updated:
+#   >  MetadataDiff(paper_trail_version)
```
awead: @rschenk don't you mean WorkVersionChangeDiff ?

app/controllers/dashboard/work_histories_controller.rb
rschenk: Good thinking on all points!

app/presenters/work_history_presenter.rb
rschenk: You're right it should be at the top, right underneath `private`, I made a goof

app/presenters/work_version_change_presenter.rb
rschenk: `"yeah probably not".to_s`

PR #90, opened 2019-11-22
Work indexing

### LINKED ISSUE #81: Index works instead of work versions
Currently all versions of a work are being indexed. Instead, we should only index the current one, and search results should return the current work only.

## TODO
* add uuid to a work
* index the work's uuid using metadata from it's current _published_ version

Note: Draft versions will not be indexed.

awead:
Changes the work and work version indexing process so that only works will be searched when searching the catalog. Work versions are still indexed in Solr so that they can be rendered by the CatalogController. However, a new boolean field is added to their documents indicating if the particular version is the latest published version or not. Also, draft versions are _not_ indexed.

A work is indexed into solr by copying the metadata fields from the latest published version and adding it to its own. This has the side-effect of creating two, almost identical, records for a given latest published version.

Fixes #81

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class WorkIndexer
+  def self.call(work, commit: false)
+    work.versions.published.map do |version|
+      IndexingService.add_document(version.to_solr, commit: false)
+    end
+    IndexingService.add_document(work.to_solr, commit: commit)
+  end
+end
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -56,6 +56,11 @@ class Work < ApplicationRecord
     work
   end
 
+  def self.reindex_all
+    find_each { |work| WorkIndexer.call(work, commit: false) }
+    IndexingService.commit
+  end
+
   def latest_version
     versions.last
   end
@@ -66,5 +71,11 @@ class Work < ApplicationRecord
 
   def draft_version
     versions.draft.last
+  end
+
+  def to_solr
+    SolrDocumentBuilder.call(resource: latest_published_version)
+      .merge(SolrDocumentBuilder.call(resource: self))
+      .except(:latest_version_bsi)
   end
 end
--- a/app/services/indexing_service.rb
+++ b/app/services/indexing_service.rb
@@ -1,49 +1,15 @@
 # frozen_string_literal: true
 
 class IndexingService
-  class Error < StandardError; end
-
-  # @param [ActiveRecord::Base]
-  # @param [Schema, Hash, nil]
+  # @param [Hash]
   # @param [Boolean]
-  def self.call(resource:, schema: nil, commit: false)
-    new(resource, schema, commit).index
+  def self.add_document(document, commit: false)
+    connection = Blacklight.default_index.connection
+    connection.add(document)
+    connection.commit if commit
   end
 
-  attr_reader :resource, :schema
-
-  def initialize(resource, schema, commit)
-    @resource = resource
-    @schema = schema || DefaultSchema.new(resource.class).schema
-    @commit = commit
+  def self.commit
+    Blacklight.default_index.connection.commit
   end
-
-  def index
-    raise Error, "#{resource.class} has no uuid defined" unless resource.respond_to?(:uuid)
-
-    connection.add(document.merge(id: resource.uuid, model_ssi: resource.class.to_s))
-    connection.commit if commit?
-  end
-
-  def document
-    schema.keys.map do |solr_field|
-      [solr_field.to_s, document_values(solr_field)]
-    end.to_h
-  end
-
-  private
-
-    def document_values(field)
-      schema[field].map do |attribute|
-        resource.send(attribute) if resource.respond_to?(attribute)
-      end
-    end
-
-    def commit?
-      @commit
-    end
-
-    def connection
-      @connection ||= Blacklight.default_index.connection
-    end
 end
--- a/app/services/solr_document_builder.rb
+++ b/app/services/solr_document_builder.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class SolrDocumentBuilder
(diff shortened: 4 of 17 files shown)


app/indexers/work_indexer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class WorkIndexer
+  def self.call(work, commit: false)
+    work.versions.published.map do |version|
```
awead: @rschenk I suppose we could exclude the latest published version here and that would avoid the duplicate record issue. There'd always be x number of records in Solr, 1 work record and x - 1 work version records.

app/models/work.rb
```diff
@@ -56,6 +56,11 @@ def self.build_with_empty_version(*args)
     work
   end
 
+  def self.reindex_all
+    all.map { |work| WorkIndexer.call(work) }
```
rschenk: You should use `find_each` here instead of `all`, because under the hood it will automatically load the records out in batches of 1,000 keeping memory usage light. Using `all` is dangerous when tables get large because it can chew through memory and make the server start swapping

app/models/work.rb
```diff
@@ -56,6 +56,11 @@ def self.build_with_empty_version(*args)
     work
   end
 
+  def self.reindex_all
+    find_each.map { |work| WorkIndexer.call(work) }
```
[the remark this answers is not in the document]
awead: @rschenk ah, no, we don't. As to the the commit, it defaults to false. The next line calls commit.

app/services/solr_document_builder.rb
```diff
@@ -0,0 +1,40 @@
...
+  def self.call(resource:, schema: nil)
+    return {} if resource.nil?
+
+    new(resource, schema).document
+  end
+
+  attr_reader :schema, :resource
+
+  def initialize(resource, schema)
+    raise Error, "#{resource.class} has no uuid defined" unless resource.respond_to?(:uuid)
```
rschenk: Personally I like `ArgumentError` for this kind of stuff

spec/factories/work_versions.rb
```diff
@@ -7,6 +7,11 @@
     title { generate(:work_title) }
     sequence(:version_number) { |n| work.present? ? work.versions.length + 1 : n }
 
+    after(:build) do |work_version|
+      # Postgres does this for us, but for testing, we can do it here to save having to call create/reload.
+      work_version.uuid = SecureRandom.uuid
+    end
+
```
rschenk: Rather than have this in a callback, could we just put this into a normal factorybot attribute?

```ruby
uuid { SecureRandom.uuid }
```

spec/factories/works.rb
```diff
@@ -10,6 +10,11 @@
       versions_count { 1 }
     end
 
+    after(:build) do |work|
+      # Postgres does this for us, but for testing, we can do it here to save having to call create/reload.
+      work.uuid = SecureRandom.uuid
+    end
+
```
rschenk: See comment on `spec/factories/work_version/rb`

app/indexers/work_indexer.rb
rschenk: Would that somehow cause mayhem if we wanted to show the latest published version via a blacklight Catalog#show page?

app/services/indexing_service.rb
```diff
@@ -1,49 +1,15 @@
...
-    new(resource, schema, commit).index
-  end
-
-  attr_reader :resource, :schema
-
-  def initialize(resource, schema, commit)
-    @resource = resource
-    @schema = schema || DefaultSchema.new(resource.class).schema
-    @commit = commit
+  def self.call(document, commit: false)
```
rschenk: Because this class has _two_ public methods, both of which are used, to me that makes `call` not a very descriptive method name. What if we named it something like `add_document`?

app/models/work.rb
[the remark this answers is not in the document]
rschenk: I think my point was, it's not super obvious in this context that it defaults to false

app/indexers/work_indexer.rb
awead: @rschenk maybe? We don't really know how the catalog is going to figure into all of this yet, so I'm inclined to just leave it alone. One thing that might cause some weirdness is that if you search for both works and work versions, the most recent version's result weight in the  relevance ranking might get skewed because there are two documents with the same metadata. I don't think this would happen when you're searching for just works _or_ versions. In either case, I don't think this is something to be concerned about now, since we'll only be searching for works until another use case comes up.

rschenk: Yah I agree, let's punt. For all we know, we could end up showing the Postgres version on that page

spec/factories/work_versions.rb
awead: @rschenk Sure. I had put in the `after(:update)` block to distinguish between `create` but I don't think that matters since calling "create" is going to execute the same things that `build` does up until the record is actually saved. I'm not 100% thrilled with this whole idea since we're bypassing Postgres, and making our test operations different than our dev and prod. operations, but I couldn't see another way, and I really didn't want to have to save/reload these test records all the time.

app/services/solr_document_builder.rb
awead: @rschenk are you saying just use `ArgumentError` or `SolrDocumentBuilder::Error` extends argument error?

rschenk: When I initially wrote that comment, I assumed `SolrDocumentBuilder::Error` was used in more places than just here, although I see now that it's not. The reason why I personally like `ArgumentError` is because it's quite specific so it makes it easy to rescue from, or track down the problem in a stack trace. I had imagined that `SolrDocumentBuilder::Error` would be used other places, like if a schema couldn't be built or something. I guess I felt like the situations where you would have an incorrect argument would be encountered by developers when developing, whereas situations raising `SolrDocumentBuilder::Error` might happen at runtime, when the correct type of object was given to the class but for some reason it failed. I guess it's really just a personal thing.

PR #95, opened 2019-11-27
Displaying file diffs between two versions

awead:
Shows the changes to files between two work versions. The versions may be of any type, either draft or published. Changed file names are shown using the same method as metadata changes. Any removed or added files are show as differently-colored rows in the table.
Screen Shot 2019-11-27 at 2 02 51 PM

### DIFF
--- a/app/services/file_version_membership_diff.rb
+++ b/app/services/file_version_membership_diff.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+# @abstract Returns a hash with the differences between the file memberships of two WorkVersions. Renamed files
+# is array of tuples, each with a FileVersionMembership, representating the two changed memberships. Added and deleted
+# files are returned as arrays of FileVersionMembership objects.
+# @example Given two objects with two different titles
+#   >  FileVersionMembershipDiff.call(obj1, obj2)
+#   => {
+#        renamed:
+#          [
+#            [FileVersionMembership, FileVersionMembership],
+#            [FileVersionMembership, FileVersionMembership]
+#          ]
+#        added: [FileVersionMembership, FileVersionMembership],
+#        deleted: [FileVersionMembership]
+#      }
+#
+class FileVersionMembershipDiff
+  # @param obj1 [WorkVersion]
+  # @param obj2 [WorkVersion]
+  def self.call(*args)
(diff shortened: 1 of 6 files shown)


awead:
@rschenk the test failure is unrelated to the PR. I should be able to fix that up later today.

app/services/file_version_membership_diff.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def diff
+    { renamed: renamed, deleted: deleted, added: added }
+  end
+
+  # @return [Array<[FileVersionMembership, FileVersionMembership]>]
+  def renamed
+    renamed_file_resource_ids.map do |id|
+      original_file = base_version_memberships.where(file_resource_id: id).first
+      renamed_file = comparison_version_memberships.where(file_resource_id: id).first
```
rschenk: The above two lines query the database unnecessarily with the `where`. We can change this to `Enumerable#find` (not to be confused with the ActiveRecord method of the same name) to do this finding in memory rather than another round trip to the DB

app/services/file_version_membership_diff.rb
```diff
@@ -0,0 +1,74 @@
...
+    def base_version_memberships
+      @base_version_memberships ||= base_version.file_version_memberships
+    end
+
+    def comparison_version_memberships
+      @comparison_version_memberships ||= comparison_version.file_version_memberships
+    end
+
+    def renamed_file_resource_ids
+      @renamed_file_resource_ids ||= base_version.file_resources.map(&:id) & comparison_version.file_resources.map(&:id)
```
rschenk: This line queries the database for all the `file_resources`, then loads their ActiveRecords into memory, just to pull out their `id`. Since the `FileResourceMembership` contains a foreign key to the FileResource, we can change this to just mapping the `:file_resource_id` on the `file_version_memberships`, this avoiding two database queries. 

Also storing the result of this intersection as some kind of indexed enumerable (Hash or Set) rather than an Array will give us a totally negligible speed boost when looking up the membership in `#added` and `#deleted` since we'd be hitting an index (O(log(n))) rather than iterating through an array (O(n)). Doesn't actually matter for lists this small but will make computer scientists smile.

PR #97, opened 2019-12-02
Store user's given- and surname separately

### LINKED ISSUE #66: User model should store first and last name instead of just "name"
Our authorization application is returning first and last name. We should store these returned values as given name and surname in our local database for future use.

rschenk:
Fixes #66

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -25,7 +25,12 @@ class User < ApplicationRecord
     User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |new_user|
       new_user.access_id = auth.info.access_id
       new_user.email = auth.info.email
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -25,7 +25,12 @@ def self.from_omniauth(auth)
     User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |new_user|
       new_user.access_id = auth.info.access_id
       new_user.email = auth.info.email
-      new_user.name = "#{auth.info.first_name} #{auth.info.last_name}"
+      new_user.given_name = auth.info.given_name
+      new_user.surname = auth.info.surname
```
rschenk: @awead this is were the names come across from the Auth-proxy

awead: @rschenk gotcha... I did miss this. 😉

Issue #37, opened 2019-10-28
Tie Members into groups


DanCoughlin:
We have groups, but no way of knowing what group a user is a member of. We can get this information from LDAP, via auth proxy.

## TODO

* create user_group_memberships join table to link users to their groups
* update the group membership of a user at login/session-renewal time
* update all additional LDAP information such as given and surname
* find or update by UID (ex. agw13)
* ~~distinguish between PSU-managed groups (i.e. umgs) and application-managed groups (i.e. "public" and "registered")~~ This can be done in #132

awead:
After discussion at stand up today, the consensus at this point is to _not_ store groups in a table, but put them in the user table as a jsonb column. There would be an array of group names, updated by the auth proxy service when the user logs into Scholarsphere. This is how we've been doing in Scholarsphere 3, except there we use a semi-colon-delimited text field. Using jsonb enables us to have indexes on the group names for better performance.

## Todo

* add jsonb column to users table
* remove existing groups table and any associations
* modify auth proxy user registration process to update the user's groups

rschenk:
@awead What specifically do we need to do for the fourth bullet point

> distinguish between PSU-managed groups (i.e. umgs) and application-managed groups (i.e. "public" and "registered")

Will individual users ever be a member of these groups? If so, what are the conditions? Will everybody always be a member of these two groups? Are these groups only used on the ACL side? We are not sure.

awead:
@rschenk they're only used on the ACL side, but when determining access, a user's groups must be cross-referenced with a work's groups specified by its ACLs. A logged in user will always be a member of "public" and "registered", whereas a non-logged in, or guest users, would only be a member of "public."

PR #107, opened 2019-12-04
Sidekiq

### LINKED ISSUE #104: Install and deploy Sidekiq and Redis
We will need Sidekiq in conjunction with Redis to manage and execute async jobs. This will be needed for services associated with DOI integration.

## WDLL

* install and configure sidekiq gem
* deploy and configure Redis
* includes Docker and k8s changes
* includes access to the default admin web interface in Sidekiq

whereismyjetpack:
Fixes #104 

- adds sidekiq
- adds redis to docker-compose
- mounts sidekiq/web at `/sidekiq` for authenticated users
- conditionally sets the active job adapter to `:sidekiq`

TODO:
Once #37  is closed, we will need to add 
```
authenticate :user, lambda { |u| u.is_admin? } do
  mount Sidekiq::Web => '/sidekiq'
end
```

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'boot'
+require_relative '../lib/scholarsphere/redis_config'
 
 require 'rails'
 # Pick the frameworks you want:
@@ -33,9 +34,13 @@ module Scholarsphere
     # the framework and any gems in your application.
 
     # Active Job Configurations
-    if ENV['REDIS_HOST']
+    redis_config = Scholarsphere::RedisConfig.new
+
+    if redis_config.valid?
       config.active_job.queue_adapter = :sidekiq
       # config.active_job.queue_name_prefix = "scholarsphere_production"
+    else
+      config.active_job.queue_adapter = :async
     end
 
     # Don't generate system test files.
--- a/config/application.rb
+++ b/config/application.rb
@@ -36,12 +36,12 @@ module Scholarsphere
     # Active Job Configurations
     redis_config = Scholarsphere::RedisConfig.new
 
-    if redis_config.valid?
-      config.active_job.queue_adapter = :sidekiq
-      # config.active_job.queue_name_prefix = "scholarsphere_production"
-    else
-      config.active_job.queue_adapter = :async
-    end
+    # config.active_job.queue_name_prefix = "scholarsphere_production"
+    config.active_job.queue_adapter = if redis_config.valid?
+                                        :sidekiq
+                                      else
+                                        :async
+                                      end
 
     # Don't generate system test files.
     config.generators.system_tests = nil
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative 'boot'
-require_relative '../lib/scholarsphere/redis_config'
 
 require 'rails'
 # Pick the frameworks you want:
@@ -24,6 +23,8 @@ Bundler.require(*Rails.groups)
 
 module Scholarsphere
   class Application < Rails::Application
+    require 'scholarsphere/redis_config'
+
     config.generators { |generator| generator.test_framework :rspec }
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
--- a/config/application.rb
+++ b/config/application.rb
@@ -36,7 +36,6 @@ module Scholarsphere
     # Active Job Configurations
     redis_config = Scholarsphere::RedisConfig.new
 
-    # config.active_job.queue_name_prefix = "scholarsphere_production"
     config.active_job.queue_adapter = if redis_config.valid?
                                         :sidekiq
                                       else
--- a/config/application.rb
+++ b/config/application.rb
@@ -32,6 +32,12 @@ module Scholarsphere
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Active Job Configurations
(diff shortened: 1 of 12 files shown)


config/application.rb
```diff
@@ -32,6 +32,12 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Active Job Configurations
+    if ENV['REDIS_HOST']
+      config.active_job.queue_adapter = :sidekiq
+      # config.active_job.queue_name_prefix = "scholarsphere_production"
```
awead: @whereismyjetpack do we need to have a separate queue name for each environment (dev, test, prod) or should this be configured per server environment (qa, stage, production) via ENV? Also, the `else` should be `:async` so we can have a queue adapter that mimics services like sidekiq without having to explicitly be one.

config/initializers/okcomputer.rb
```diff
@@ -1,5 +1,11 @@
...
 
+require 'scholarsphere/redis_config'
+
+redis_config = Scholarsphere::RedisConfig.new
+
 OkComputer.mount_at = false
 
 OkComputer::Registry.register 'solr', OkComputer::SolrCheck.new(Rails.application.config_for(:blacklight)[:url])
+
+OkComputer::Registry.register 'redis', OkComputer::RedisCheck.new(redis_config.call) if redis_config.redis_host
```
awead: @whereismyjetpack I think `call` isn't the right name here. Maybe just `config`? Also, you could make a method like `valid?` which returns true or false and checks for the presence of a host, and possibly other things, if needed.

config/application.rb
```diff
@@ -32,6 +32,12 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Active Job Configurations
+    if ENV['REDIS_HOST']
```
awead: @whereismyjetpack would `if redis_config.valid?` be better here?

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'scholarsphere/redis_config'
+
+redis_config = Scholarsphere::RedisConfig.new.call
```
awead: @whereismyjetpack see  my early comments, but I think just `Scholarsphere::RedisConfig.new` here and then redis_config.config or redis_config.to_hash or something...

lib/scholarsphere/redis_config.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Scholarsphere
+  class RedisConfig
+    def call
+      redis_config
+    end
```
awead: @whereismyjetpack don't use the call method and just use `def config`

lib/scholarsphere/redis_config.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def redis_host
```
awead: @whereismyjetpack I think all the `redis_` bits are unnecessary since it's the RedisConfig class, right?

lib/scholarsphere/redis_config.rb
```diff
@@ -0,0 +1,33 @@
...
+    def redis_database
+      ENV.fetch('REDIS_DATABASE', 0)
+    end
+
+    def redis_password
+      ENV.fetch('REDIS_PASSWORD', nil)
+    end
+
+    def redis_config
+      config = {
```
awead: @whereismyjetpack `config` isn't required here, just do:
``` ruby
def awesome_hash
  {
    awesome: true
  }
end

File: config/application.rb (hunk 1 above)
Comment (awead) [reply config/application.rb]: Another thought, if this is per server env. should it be the domain of RedisConfig? Don't know...

File: config/application.rb (hunk 3 above)
Comment (whereismyjetpack) [reply config/application.rb]: yup, i like that much better.

File: config/initializers/okcomputer.rb (hunk 2 above)
Comment (whereismyjetpack) [reply config/initializers/okcomputer.rb]: i like the idea of valid?

File: config/application.rb (hunk 1 above)
Comment (whereismyjetpack) [reply config/application.rb]: I think the queue name prefix should probably be part of application in the SIDEKIQ_ namespace. could be a single env var..

File: config/application.rb [hunk 8]
```diff
@@ -32,6 +33,16 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Active Job Configurations
+    redis_config = Scholarsphere::RedisConfig.new
+
+    # config.active_job.queue_name_prefix = "scholarsphere_production"
```
awead: @whereismyjetpack is the queue name still relevant? If not, I'd remove this line. Otherwise, we could do something like `config.active_job.queue_name_prefix = "scholarsphere_#{Rails.env}"`

config/application.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'boot'
+require_relative '../lib/scholarsphere/redis_config'
```
awead: @whereismyjetpack this is just `require 'scholarpshere/redis_config'`

config/application.rb
whereismyjetpack: I'm not sure that it is. we can always add it later if we decide it's needed.

config/application.rb
whereismyjetpack: I think i'm missing something as part of initialization, i tried that and got ```LoadError: cannot load such file -- scholarsphere/redis_config```

PR #111, opened 2019-12-05
DataCite API Client

### LINKED ISSUE #102: Datacite client service
Write a service that sends requests to the Datacite endpoint.

## WDLL

* register method for creating new DOIs that have incomplete metadata or are for unpublished works
* publish method that updates or creates a DOI in a published state for a published work
* update method that updates metadata for an existing DOI
* any errors will generate exceptions with the Datacite response in the message
* successes will return the DOI string, and datacite response as a json hash

rschenk:
Client class to manage DOIs via DataCite. Also installs the VCR gem for
mocking of network calls.

Fixes #102

### DIFF
--- a/lib/data_cite/client.rb
+++ b/lib/data_cite/client.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+require 'faraday'
+
+module DataCite
+  class Client
+    class Error < StandardError; end
+
+    def initialize(prefix: nil, publisher: nil)
+      @prefix = prefix
+      @publisher = publisher
+    end
+
+    def register(suffix = nil)
+      attributes = if suffix.present?
+                     { doi: "#{prefix}/#{suffix}" }
+                   else
+                     { prefix: prefix }
+                   end
+
+      data = {
+        type: 'dois',
+        attributes: attributes
+      }
+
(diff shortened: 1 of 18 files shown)


lib/data_cite/client.rb
```diff
@@ -0,0 +1,126 @@
...
+          publisher: publisher
+        )
+      }
+
+      if doi.blank?
+        data[:attributes][:prefix] = prefix
+        process_response post(data: data)
+      else
+        data[:id] = doi
+        process_response put(doi: doi, data: data)
```
awead: @rschenk we're not testing this else case.

lib/data_cite/client.rb
```diff
@@ -0,0 +1,126 @@
...
+          errors = parsed_body.fetch('errors', [])
+          stringified_errors = errors.inspect
+          message = stringified_errors.presence || "DataCite response returned #{response.status}"
+
+          raise Error.new(message)
+        end
+
+        doi = parsed_body.dig('data', 'id')
+
+        [doi, parsed_body]
```
awead: @rschenk According to Code Climate, we don't seem to be testing this method. 🤷‍♂

PR #117, opened 2019-12-09
Tie members into groups

### LINKED ISSUE #37: Tie Members into groups

whereismyjetpack:
A list of groups that the user belongs to will be coming out of our auth proxy. We take this list and sync it into the Groups table which maintains a relationship with the user's record.

With the user's group membership maintained locally, we can then use the groups to contruct access controls for enforcing group-based access on resources.

Fixes #37

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,19 +15,34 @@ class User < ApplicationRecord
            as: :agent,
            dependent: :destroy
 
+  has_many :user_group_memberships,
+           dependent: :destroy
+
+  has_many :groups,
+           through: :user_group_memberships
+
   validates :email,
             presence: true,
             uniqueness: true
 
-  # TODO Which fields do we want updated each time a user logs in?
-  #   - Group memberships
   def self.from_omniauth(auth)
-    User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |new_user|
+    user = User.find_or_initialize_by(provider: auth.provider, uid: auth.uid) do |new_user|
+      # Written once, when the user is created, then never again
       new_user.access_id = auth.info.access_id
-      new_user.email = auth.info.email
-      new_user.given_name = auth.info.given_name
-      new_user.surname = auth.info.surname
     end
+
+    # Updated every login
+    user.email = auth.info.email
+    user.given_name = auth.info.given_name
+    user.surname = auth.info.surname
+
+    user.groups = auth.info.groups.map do |group_name|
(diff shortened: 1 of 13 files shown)


app/models/user.rb
```diff
@@ -22,12 +22,17 @@ class User < ApplicationRecord
   # TODO Which fields do we want updated each time a user logs in?
   #   - Group memberships
   def self.from_omniauth(auth)
-    User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |new_user|
+    user = User.find_or_create_by(provider: auth.provider, uid: auth.uid) do |new_user|
```
whereismyjetpack: This is most certainly wrong, but I wasn't sure the best way to _only_ mutate ldap groups.

app/models/user.rb
```diff
@@ -22,12 +22,17 @@ class User < ApplicationRecord
...
       new_user.access_id = auth.info.access_id
       new_user.email = auth.info.email
+      new_user.ldap_groups = auth.info.groups
       new_user.given_name = auth.info.given_name
       new_user.surname = auth.info.surname
     end
+
+    user.update(ldap_groups: auth.info.groups)
```
whereismyjetpack: I'm curious if we maybe only want to run this operation if authproxy returns groups && they differ from what's already in the database? @rschenk Thoughts on this one?

spec/factories/psu_oauth_responses.rb
```diff
@@ -11,6 +11,7 @@
       given_name { Faker::Name.first_name }
       surname { Faker::Name.last_name }
       email { "#{access_id}@psu.edu" }
+      groups { Array.new(3) { Faker::Coffee.blend_name } }
```
awead: 🤣 🤣 🤣 🤣 🤣 🤣 🤣 🤣

whereismyjetpack:
I think we're ok removing those two migrations, since they never landed in mainline. ok to merge?

awead:
@whereismyjetpack yeah, try it out with a deploy to see if it's doing what we want.

PR #127, opened 2019-12-11
Use patern matching for created and updated dates

awead:
Matching exact dates was leading to weird time differences. Instead, we'll match on the day only and the time.


awead:
@rschenk I'm not positive it isn't a timezone bug, so we'll see, I guess!

Issue #101, opened 2019-12-03
Mapping metadata to Datacite


awead:
Which fields are _required_ for datacite and which additional fields do we want required for Datacite's record?

## WDLL

* json object that represents the object's metadata to be sent to Datacite for either updating or creating
* creating metadata objects with incomplete fields should be prohibited

awead:
@srerickson will provide a mapping scholarphere resources types and datacite resource types.

rschenk:
Some questions for @srerickson as I'm coding up an initial sketch of this, and I will probably update this comment as I encounter more issues:

### DataCite's publicationYear
1. `publicationYear` is a required attribute in the DataCite metadata, but our `WorkVersion#published_date` field is not required for publication in our data model. We should probably change this
1. `publicationYear` in DataCite is a single-value field, but `WorkVersion#published_date` is a multiple-value field. What are the rules for reducing those multiple values to a single value?
1. `WorkVersion#published_date` is a string field in our data model, but DataCite's `publicationYear` wants an integer. This means we will need to parse our string values into date objects. This is a good opportunity to discuss what date formats we will accept in our forms, and we should write a ticket for validating and parsing that format on the data-entry side.

srerickson:
### Resource Type Mappings

The mapping of ScholarSphere Resource Types to DataCite Resource Types is [here](https://docs.google.com/spreadsheets/d/1wGVdGFIfsJOC-nVHPzYEAdYgfpadSPgRwEQ7xsBOFQo/edit#gid=372862448). 

### DataCite's publicationYear 

Responding to @rschenk:

1. I agree that `published_date` should be required (not just because it's required by DataCite but also because it's required by Google Scholar). The problem with making it required is that a lot of existing ScholarSphere Works would become invalid. Maybe that's OK, it's just something to note. (One of our metadata clean-up projects is to add publication years for all Works).

2. I'm not sure why `publication_date` (along with other `WorkVersion` fields) is a multivalue field. I expect this is for compatibility with some other part of the stack (@awead ?). I don't think the UI has or needs to support multiple publication dates, so just choosing the first one seems fine. 

[...]

PR #139, opened 2019-12-17
DataCite metadata mapper

rschenk:
Takes a WorkVersion and maps it to the minimum required set of metadata for DataCite

References #101

### DIFF
--- a/lib/data_cite/metadata.rb
+++ b/lib/data_cite/metadata.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module DataCite
+  class Metadata
+    class Error < StandardError; end
+    class ValidationError < Error; end
+
+    attr_reader :work_version
+
+    RESOURCE_TYPES = {
+      Work::Types::DATASET => 'Dataset'
+    }.freeze
+
+    def initialize(work_version:)
+      @work_version = work_version
+      @work = work_version.work
+    end
+
+    def attributes
+      {
+        titles: [{ title: work_version.title }],
+        creators: [creator],
+        publicationYear: publication_year,
+        types: {
+          resourceTypeGeneral: resource_type
+        }
+      }
+    end
+
+    def validate!
+      raise ValidationError.new("title can't be blank") if attributes[:titles].map { |t| t[:title] }.all?(&:blank?)
+
+      raise ValidationError.new("publicationYear can't be blank") if attributes[:publicationYear].blank?
+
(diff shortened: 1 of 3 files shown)


lib/data_cite/metadata.rb
```diff
@@ -0,0 +1,72 @@
...
+        }
+      end
+
+      def publication_year
+        date = parsed_publication_date || work_version.created_at
+        date&.year
+      end
+
+      def parsed_publication_date
+        Time.zone.parse(work_version.published_date.first)
```
awead: @rschenk Any reason why we wouldn't use DateTime?

lib/data_cite/metadata.rb
```diff
@@ -0,0 +1,72 @@
...
+        titles: [{ title: work_version.title }],
+        creators: [creator],
+        publicationYear: publication_year,
+        types: {
+          resourceTypeGeneral: resource_type
+        }
+      }
+    end
+
+    def validate!
```
awead: @rschenk nit picky: should this be a private method?

spec/lib/data_cite/metadata_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        before { allow(work).to receive(:work_type).and_return(Work::Types::DATASET) }
+
+        it 'maps to the correct resource type' do
+          expect(attributes.dig(:types, :resourceTypeGeneral)).to eq 'Dataset'
+        end
+      end
+    end
+  end
+
+  describe '#validate! and #valid?' do
```
awead: Seeing this test make me think that `validate!` ought to be a private method and we access it through `valid?` unless there's a legitimate use case for the former. It doesn't matter too much to me, though.

lib/data_cite/metadata.rb
rschenk: `Time.zone` is how you hook into Rail's timezone support, `DateTime` doesn't have an equivalent feature. Also since Ruby doesn't actually have a time-of-day class (as you'd assume Time would be), the difference between the two is fuzzy at best

spec/lib/data_cite/metadata_spec.rb
rschenk: I'm following the `ActiveModel`/`ActiveRecord` API here, which has both methods public. [validate!](https://api.rubyonrails.org/classes/ActiveModel/Validations.html#method-i-validate-21) raises an error vs [valid?](https://api.rubyonrails.org/classes/ActiveModel/Validations.html#method-i-valid-3F) returns a boolean

Issue #135, opened 2019-12-11
Implement permissions


awead:
Given that we have users, groups, works, and ACLs, we can combine them to form a set of permissions that govern a resource. There are currently only two agents (user and group) as well as two levels (read and edit), depending on the outcome of #133. Combining those into a matrix gives us four dimensions of permissions: read users, read groups, edit users, and edit groups.

## WDLL

* methods for returning a list of users or groups for each dimension
* methods for adding ACLs to a work

Issue #133, opened 2019-12-11
Define access control levels


awead:
Levels for ACLs can be very simple. Currently, there are only two: read and edit. However, additional levels can be defined as needed such as discover, download, or destroy.

Note: Application-level policies will supersede abilities granted by ACLs. For example, if an ACL grants access to a withdrawn resource, restrictions imposed on the resource by the "withdrawn" status will override any access granted by the ACL. Likewise, an ACL that restricts access to a resource will be ignored if the user is an admin, which is an application-level permission.

# Resolutions

Here are the access levels for Scholarsphere:

## Discover

* **can** search and view any metadata in the UI for any work, work version, or collection
* **cannot** download any binary content
* **cannot** create, update, and delete any content

## Read

* **can** search and view any metadata in the UI for any work, work version, or collection
* **can** download all available binary content from the UI
* **cannot** create, update, and delete any content

## Edit

* **can** search and view any metadata in the UI for any work, work version, or collection
* **can** download all available binary content from the UI
* **can** create, update, and delete all aspects of a work, work version, or collection:
    * edit all metadata presented in the UI
    * delete draft resources on existing works
    * create new work versions for existing works
    * publish work versions for existing works
    * add, delete, and edit any files on draft work versions
    * add and remove works in collections

awead:
ping @srerickson Feel free to discuss this here, async, but I'm also happy to talk about it in real time.

PR #141, opened 2019-12-17
Define permissions based on access controls

### LINKED ISSUE #132: Visibility groups
Scholarsphere has the notion of "visibility" which is a application-wide type of permission that can encompass multiple types of permission, but is currently used for two main kinds of access: public, and registered. Public access is identical to Open Access, meaning anyone can access a record's metadata and download its content. Registered access means that the user must be "registered" with the application, or logged in. This translates to institutional access, or anyone with valid Penn State credentials.

These two kinds of permissions are implemented as groups: "public" for open access, and "registered" for Penn State access.

We also have "private" visibility to denote a work that is only accessible to the depositor, but this has been disabled as a UI option because of open access considerations. We could continue to use it, or remove it entirely.

## WDLL

* define in the application two groups for public and registered access
* ensure that these groups are seeded in the groups table
* groups should be notated in such a way that their values, ex. public and registered, are inconsequential, i.e. no "magic strings"
* a group's name should be able to be communicated to the user in a different way, ex. the "public" group name could be display as "Open Access"

## Questions

Note to @srerickson 

* remove the notion of "private" visibility?
* would there be a need for additional types of visibility?
* should we remove the notion of visibility entirely and solely rely on ACLs to communicate open access, PSU-access, restricted access, etc?

awead:
Using a set of "levels" for defining access rights to work resources, an agent, such as a user or group, is granted a level of access that affords them a set of abilities with a given resource.  These levels include:

  * discover access, for reading metadata but not downloading
content
  * read access, for both metadata and binary content access
  * edit access, for all other rights, including the ability to create,
edit, and delete content, as well as create new versions of works

Two application-wide groups are provided for granting open access, and access to anyone at Penn State. A Permissions module manages the agents who are granted different levels of access and ensures that each combination of agent and level is unique to a particular resource. The module builder pattern dynamically adds the methods for creating different permission dimensions. See https://dejimata.com/2017/5/20/the-ruby-module-builder-pattern

Fixes #132, #133, #135

### DIFF
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -1,8 +1,26 @@
 # frozen_string_literal: true
 
 class AccessControl < ApplicationRecord
+  class Level
+    DISCOVER = 'discover'
+    READ = 'read'
+    EDIT = 'edit'
+
+    def self.all
+      [DISCOVER, READ, EDIT]
+    end
+
+    def self.default
+      DISCOVER
+    end
+  end
+
   belongs_to :agent,
              polymorphic: true
   belongs_to :resource,
              polymorphic: true
+
+  validates :access_level,
+            uniqueness: { scope: [:agent, :resource] },
+            inclusion: { in: AccessControl::Level.all }
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,3 +35,7 @@ RSpec/MultipleExpectations:
 RSpec/EmptyLineAfterSubject:
   Exclude:
     - 'spec/factories/work_versions.rb'
+
+RSpec/LeakyConstantDeclaration:
+  Exclude:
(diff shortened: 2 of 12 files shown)


app/models/access_control.rb
```diff
@@ -1,8 +1,17 @@
...
+    EDIT = 'edit'
+  end
+
   belongs_to :agent,
              polymorphic: true
   belongs_to :resource,
              polymorphic: true
+
+  validates :access_level,
+            uniqueness: { scope: [:agent, :resource] }
```
rschenk: nitpick: we should also probably validate that `access_level` is one of our Level constants, similar to how we do `Work#work_type`

awead: @rschenk yes, good catch! I will add that in...

PR #144, opened 2019-12-19
Mint and/or update a DOI for WorkVersions and Works

### LINKED ISSUE #103: DOI Service
Service that does all the work for creating and updating DOIs on works and work versions.

## WDLL

* accepts works or work verisons
* generates the DOI metadata for the the given work or work version
* calls the DOI client (in this case Datacite)
* handles the response by either capturing exceptions or letting the halt the process
* updates the given work or work version with the generate DOI returned from the client

rschenk:
This is a service object that ties together the DataCite client that Adam and I wrote, along with the DataCite metadata mapper class. 

In order to pull this off, I needed to add a few things besides just make a service:
* Verify that `DataCite::Client#publish` is idempotent on DataCite's side. That is, we can call it multiple times and it'll do the right thing (i.e. update the metadata on the second call, without having a freakout that it's already been published.)
* Add a method `DataCite::Client#get` to retrieve the metadata of a particular DOI. This is used for testing
* Add a public URL to the Datacite::Metadata mapper. **_Please note that I did not know what a particular resource's public URL should actually be, and I remember there being some discussion around this. So in the mean time I put in a bogus value, and then put in a (currently failing) pending spec to ensure that it's something reasonable once we figure that out._**

Currently the integration test hits the real DataCite API, and _only_ runs on CI. Dann and I had thought of a few ways we could make this fancier, but ultimately `skip: !ci_build?` is simple and effective.

Fixes #103

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,7 @@ Metrics/BlockLength:
 RSpec/DescribeClass:
   Exclude:
     - 'spec/features/**/*'
+    - 'spec/integration/**/*'
 
 RSpec/ExampleLength:
   Enabled:
--- a/app/services/doi_service.rb
+++ b/app/services/doi_service.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require 'data_cite'
+
+class DoiService
+  class Error < StandardError; end
+
+  def self.call(resource, client_source: nil, metadata_source: nil)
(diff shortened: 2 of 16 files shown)


rschenk:
@whereismyjetpack ☝️

whereismyjetpack:
Cool beans!

PR #147, opened 2019-12-20
exit if any command exits non-zero

whereismyjetpack:
We should not start a deployment if any of the tasks in `start.sh` fails. this allows the script to exit early, allowing the old replica to continue to run until an administrator can mitigate the issue.


awead:
@whereismyjetpack do you still need this?

PR #151, opened 2020-01-08
Set the visibility of a work in order to publish

### LINKED ISSUE #146: Setting visibility permissions is required to publish
Unless otherwise specified by the user, a work will be open access when it is published. The user will have the opportunity to select another form of access, such as Penn State only, but the work cannot be published unless one is chosen.

## WDLL

* works cannot be published unless they are open access or Penn State only
* provide UI selector to choose the visibility at publish time
* visibility defaults to open access
* ~~create UI badges similar to Scholarsphere to denote visibility~~ See #152

awead:
Visibility is now set when the work is first created, and any works may not be published unless the work is either available to the public, i.e. open access or open visibility, or is available within Penn State, i.e. authorized access or Penn State visibility.

The major change here is that the user must select the visibility when first creating the work, along with its title and datatype, as opposed to the _end_ of the process when it is published.

Screen Shot 2020-01-08 at 10 12 22 AM

Fixes #146

### DIFF
--- a/app/controllers/dashboard/works_controller.rb
+++ b/app/controllers/dashboard/works_controller.rb
@@ -54,6 +54,7 @@ module Dashboard
           .require(:work)
           .permit(
             :work_type,
+            :visibility,
             versions_attributes: [
               :title
             ]
--- a/app/models/concerns/permissions.rb
+++ b/app/models/concerns/permissions.rb
@@ -15,6 +15,13 @@ module Permissions
     def self.default
       OPEN
     end
+
+    def self.options_for_select_box
+      {
+        'Open Access' => OPEN,
+        'Penn State' => AUTHORIZED
+      }
+    end
   end
 
   included do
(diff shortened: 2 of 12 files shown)


awead:
@srerickson Note the new UI change. Since visibility and permissions can't change between work versions, putting this on the work creation step makes sense.

srerickson:
@awead sorry for not approving this yesterday.

srerickson:
A few thoughts on this:

- PSU-only visibility is important to provide but also discouraged. Will having this setting up front encourage more people to use it?
- We may want to allow a Work that is PSU-only to be changed to Public after it is Published. (But not allow Public work to be changed to PSU-only).

awead:
@srerickson We can discourage PSU-only settings with UI language, or we can just make works public by default, and let users change it prior to publishing. I'm actually inclined toward the latter because it would simply things greatly. Setting works to PSU-only would be a separate UI and would require the user to really know that they wanted to do it.

Regarding your second point, the separate UI for permissions would allow this. We can control how things happen via our policy methods in Pundit. Something like:

* if it's unpublished, you can move between PSU-only and public, as well as add any additional permissions related to read or edit
* if it's published, you can only make public, or add additional editors?

PR #154, opened 2020-01-10
Index ACLs and enforce them in search results

### LINKED ISSUE #128: Enforce access controls in Solr searches
Use the Blackglight acl gem to enforce access controls when searching a retrieving works and work versions from Solr.

## Steps

* install blacklight-acl gem
* index ACLs into works and work versions
* configure via SearchBuilder to add the access control restrictions to searches

## WDLL

I can search for works, but only see ones that I have read access to, which includes any public work, registered work, or work that has access via name or associated group. As a public user, I should only see works that public.

awead:
Enforces ACLs in search results by indexing discover and read permissions in Solr, and then ensuring Blacklight only returns search results that obey those permissions. For example, everyone will be able to search and retrieve public works, but works designated as "Penn State only" can only be searched by registered users.

Originally, we had intended to use Blacklight's access controls gem, but it proven to be incompatible and overly complicated.

In order to complete this, three other changes were made in separate commits:
* implementing a guest user to effect a "null object" type of pattern so we don't have to check for a user all the time
* adding default groups to users to control access rights to public content
* refactoring feature tests to use Warden instead of OAuth when applicable

Fixes #128

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -3,10 +3,64 @@
 class SearchBuilder < Blacklight::SearchBuilder
   include Blacklight::Solr::SearchBuilderBehavior
 
-  self.default_processor_chain += [:show_works_only]
+  self.default_processor_chain += %i(
+    restrict_search_to_works
+    apply_gated_discovery
+    log_solr_parameters
+  )
 
-  def show_works_only(solr_parameters)
+  def restrict_search_to_works(solr_parameters)
     solr_parameters[:fq] ||= []
     solr_parameters[:fq] << 'model_ssi:Work'
   end
+
+  def apply_gated_discovery(solr_parameters)
+    return if current_user.admin?
+
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] << gated_discovery_filters.compact.join(' OR ')
+  end
+
+  def log_solr_parameters(solr_parameters)
+    Rails.logger.debug("Solr parameters: #{solr_parameters.inspect}")
+  end
+
+  private
+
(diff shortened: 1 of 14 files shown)


spec/factories/psu_oauth_responses.rb
```diff
@@ -11,7 +11,9 @@
       given_name { Faker::Name.first_name }
       surname { Faker::Name.last_name }
       email { "#{access_id}@psu.edu" }
-      groups { Array.new(3) { Faker::Coffee.blend_name } }
```
rschenk: 😆

app/models/search_builder.rb
```diff
@@ -3,10 +3,64 @@
...
+
+  def apply_gated_discovery(solr_parameters)
+    return if current_user.admin?
+
+    solr_parameters[:fq] ||= []
+    solr_parameters[:fq] << gated_discovery_filters.compact.join(' OR ')
+  end
+
+  def log_solr_parameters(solr_parameters)
+    Rails.logger.debug("Solr parameters: #{solr_parameters.inspect}")
```
rschenk: May want to remove this trace?

awead: @rschenk I took this from the Blacklight Access Control gem code. It comes in very handy when you're trying to debug solr query issues, for example when a particular record isn't showing up in the display results. It wouldn't be needed in production, unless we enabled debug-level logging. Are you suggesting there's a better way to do this?

rschenk: Nah I just wasn't sure if it was a mistake left over from debugging 👍

PR #157, opened 2020-01-13
Introdude ActionView::Component in favor of presenters

### LINKED ISSUE #153: Refactor views to use ActiveComponent
Refactors existing view partials to use ActiveComponent architecture.

rschenk:
This PR installs [ActionView::Component](https://github.com/github/actionview-component) (to be a part of Rails core in v6.1) and converts most of our presenter objects to use it. 

The biggest commit is "Convert work history presenters to ActionView::Components" which is quite gnarly, due to those presenters also being quite gnarly in order to do their job efficiently. This was effectively a direct port from our presenter pattern to the component pattern with very little code changed.

If you want to learn how to make a component, the best commit to look at is "Make publication status badge into an ActionView::Component" which is a nice, small, self contained component used for our publication statuses. Formerly this was a partial with a bit too much logic inside it, now happily converted into a component.

The only weird gotcha is, despite following the installation instructions as best I could, I needed to `require 'action_view/component'` atop each component class in order to get them to run properly. I don't know why, this doesn't seem desireable, and there's no mention of this in the Component gem readme. I wasn't sure whether to dig in and try to fix this now, or perhaps wait until we upgrade to Rails 6.1.

Speaking of upgrading to Rails 6.1, I kind of think we should leave the first commit "Install ActionView::Component" as un-squashed in the git commit history. That way when we do upgrade to Rails 6.1 which ships with Component out of the box, presumably we'll need to undo whatever is done in this commit?

Fixes #153


awead:
@rschenk Agreed, I'm all for granular commits if we think, or are at least reasonably sure, they should be reverted at a later date.

app/components/work_histories/work_history_component.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      def lookup_user(user_id)
+        user_lookup_cache[user_id] ||= find_user(user_id)
+      end
+
+      def find_user(user_id)
+        User.find(user_id)
+      rescue ActiveRecord::RecordNotFound
+        null_user
```
awead: @rschenk code climate is complaining this line isn't tested. I'm cool if you want to ignore this, but I'll leave it up to you if you think it ought to be tested.

rschenk:
@awead Component gem's readme says the [convention should be plural](https://github.com/github/actionview-component#building-components)... that being said, they never actually _do_ that in any of their examples

app/components/work_histories/work_history_component.rb
rschenk: Hmm, yeah, it isn't tested. I figured this is a very, very unlikely scenario but maybe I should test this. Lemme spend 15 minutes on it and if I can come up with a good test, great, and if not, we'll skip it

PR #158, opened 2020-01-13
Clean group names from LDAP records

### LINKED ISSUE #155: Enforce valid group names
Group names need need some requirements, otherwise they'll break things.

* they cannot contain spaces: it'll break our Solr queries
* they should look like they do in current Scholarsphere

Since we get group names from LDAP via our OAuth app, it's unlikely that they'll ever contain spaces, but we should include some validation in the model to make this explicit, in case any groups are ever created manually or by some other process.

Currently, groups coming to us from OAuth look like: `cn=umg/up.its.sas.nagios,dc=psu,dc=edu` but in Scholarsphere, they look like: `umg/up.its.sas.nagios` with the other stuff around it removed.

# WDLL

* verify format of existing group names in Scholarsphere ✔️ 
* prevent any group with a space in its name from being saved
* clean up the names coming out of OAuth to match what's current in Scholarsphere

rschenk:
- Extract the group name from the LDAP record
- Any groups with spaces in the name are gracefully ignored

Fixes #155

### DIFF
--- a/app/services/ldap_group_cleaner.rb
+++ b/app/services/ldap_group_cleaner.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class LdapGroupCleaner
+  class Error < RuntimeError; end
+
+  class << self
+    def call(ldap_record)
+      call!(ldap_record)
+    rescue Error
+      nil
+    end
+
+    def call!(ldap_record)
+      _cn, value = ldap_record
+        .split(',')
+        .map { |str| str.split('=') }
+        .find { |key, _v| key == 'cn' }
+
+      raise Error.new('LDAP group cannot contain spaces') if value.to_s.match?(/\s/)
+      raise Error.new('LDAP group is empty') if value.to_s.empty?
+
+      value
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/ldap_group_cleaner.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class LdapGroupCleaner
+  class Error < RuntimeError; end
+
+  class << self
+    def call(ldap_record)
+      call!(ldap_record)
+    rescue Error
+      nil
```
awead: @rschenk we ought to log these errors. If ldap is returning oddly named groups, we'd want to know about it.

rschenk: Where should we log them? Do we have a tool like bugsnag or similar? Seems like just writing them to the Rails log, they'll probably never be seen again

whereismyjetpack: We use splunk for logging, we could do forensics there if we need to.

awead: @rschenk Rails log at the `error` is fine for now, but yes, we need a way of being alerted if these things are showing up. I'm imagining we can grep for LdapGroupCleaner::Error via Splunk.

PR #160, opened 2020-01-15
Visibility badges

### LINKED ISSUE #152: Create UI badges to denote visibility
Similar to Scholarsphere 3.0, we need to identify the visibility of works.

## WDLL
* differently labled "badges" next to works showing visibility
* appears in search results and any work display or dashboard pages

Blocked by #153

awead:
Adds "badges" for different visibility states that are displayed in the dashboard and the search results page. Includes images as well as text, but can be reconfigured later when UI requirements change.

Fixes #152

## Changes
* adds VisibilityBadgeComponent to indicate the visibility of a work
* overrides Blacklight's search results title header partial to display badge
* indexes visibility into Solr document so VisibilityBadgeComponent can work with both Work and SolrDocument resources
* includes image icons for each visibility

## Dashboard View
Screen Shot 2020-01-14 at 9 51 04 PM

## Search Results View
Screen Shot 2020-01-14 at 9 50 39 PM


app/components/visibility_badge_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= content_tag :span, class: html_class do %>
+  <%= image_tag details[visibility][:image], class: image_class %>
+  <%= details[visibility][:label] %>
+<% end %>
```
rschenk: I'll admit, the `details[][]` stuff in the view has a code smell to me. I know Sandi Metz would have a snappy and clever name for this, but I think it just exposes way too much of the internal workings of the class to the view. 

Could we instead have two methods in the class that do this, and call them in the view?

```ruby
class VisibilityBadgeComponent
  # ...
  def image_src
    details[viz][:image]
  end

  def label
    details[viz][:label]
  end
end
```

Second nit-pick is, there's something about the `content_tag` that threw me off. I get why you did it, and I also think that the `image_tag` is 💯, but I'm thinking about when we have a front-end person on this project, would they understand what content_tag is doing vs 

```html
<span class="<%= html_class %>">
  ...
</span>
```

awead: @rschenk first point, 👍 .... second point, I did the content tag because everything else was interpolated ruby, so why not all of it? `"<%= html_class %>"` just looks janky to me, too. Your point about the front-end person is valid, but I would expect them to understand that.

rschenk: Yep, I get it and I figured that's exactly why you did it.

PR #163, opened 2020-01-16
Determine access using a user's groups

awead:
Fixes a bug where a user who is a member of a group that is granted a level of access, must be allowed to access that resource. For example, given a group that is granted read access to a resource, any user that is a member of that group must also be allowed read access.

One of the issues created with this fix is that the PermissionsBuilder class must know that "groups" is a means of finding associated agents from another agent. This blurs the abstract nature of agents a bit, but would make good fodder for refactoring.

### DIFF
--- a/app/models/concerns/permissions_builder.rb
+++ b/app/models/concerns/permissions_builder.rb
@@ -31,8 +31,13 @@ class PermissionsBuilder < Module
       end
     end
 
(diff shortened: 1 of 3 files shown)


awead:
PR is failing due to an issue with the external DOI service, which doesn't indicate anything's wrong on our side. It's a just a 404 error after the DOI is minted, probably because the Fabrica service is a little slow to respond.

PR #168, opened 2020-01-21
Authorized and open access only applies to groups

awead:
If a user is granted read level access, it will cause NoMethodError because we were not checking the agent type. Open and authorized access only applies to groups with read level permissions.

This also refactors how we determine public and authorized agents by defining them in the access control resource.

### DIFF
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -23,4 +23,12 @@ class AccessControl < ApplicationRecord
   validates :access_level,
             uniqueness: { scope: [:agent, :resource] },
             inclusion: { in: AccessControl::Level.all }
+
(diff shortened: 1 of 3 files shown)


awead:
@rschenk it looks to be okay. I popped open the console, and there's one SELECT statement when I run:

``` ruby
agent.is_a?(Group) && agent.public? && access_level == AccessControl::Level::READ
```
But this is because we're not memoizing `agent.public?` bit in test. (see 4fde51b44a042c96a993b914a32d3190c9ba94ed) In production, this will be memoized, and there'd be no additional database call on lines 28 and 32 of `AccessControl`.

PR #170, opened 2020-01-22
Actual public resource URL provided to DOI service

### LINKED ISSUE #92: Canonical URLs
What is the canonical url for a work or a work version? These would be the urls that DOIs resolve to, or are otherwise the unique url for a given work or work version. All published works and their versions would have one.

## Options

1. Use Blacklight's catalog controller, with the UUID:
`/catalog/552b93f8-5997-4d5e-9954-a42b7e5bf7fa`
2. Use a different pathname, and/or a different kind of identifier:
`/resources/1234asdf`

###  Architectural Considerations

With option 1, we'd be invested more into Blacklight's Solr-based representation of the resource. This has the advantage of already being there, but means we'd need to maintain the Solr representation along with the Postgres one, if we're intending to still continue using it for display purposes.

Option 2 would allow us to put the work representation in Postgres, and relegate Blacklight's activity to searching only.

rschenk:
This completes the coding work necessary to provide the full public URL to DataCite. 

In situations like the DOI thing, when we will be outside the request-response cycle (either in a background job, importing over the API, etc) we need to explicitly tell Rails what its own hostname is because it cannot be inferred from the HTTP `Host` header.

I did this by adding an environment variable `DEFAULT_URL_HOST` and setting the Rails config from it in `config/initializers/default_url_host.rb`. This process will raise an exception if no environment variable is defined.

Fixes #92

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -92,6 +92,7 @@ steps:
         -e DATACITE_PUBLISHER=$DATACITE_PUBLISHER
         -e AWS_REGION=us-east-1
         -e REDIS_HOST=redis
+        -e DEFAULT_URL_HOST=scholarsphere
         -e APP_HOST=test-$DRONE_BUILD_NUMBER
         -e POSTGRES_USER=scholarsphere
         -e SELENIUM_URL=http://selenium:4444/wd/hub
@@ -209,6 +210,6 @@ volumes:
     path: /var/run/docker.sock
 ---
 kind: signature
-hmac: fec077364ec8d7b5af5d66761821be7226d877d9eda915f46125094a2660aa8d
+hmac: 1a1f3f4e1c2893e31fb1cdef8830cd316b4fcae1f3527d0b2cad386584a4adf4
 
 ...
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,4 +57,4 @@ CMD [ "/app/bin/ci-niftany" ]
 # Final Target
 FROM base as production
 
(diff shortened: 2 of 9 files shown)


awead:
@rschenk We did some similar shenanigans in SS3:

https://github.com/psu-stewardship/scholarsphere/blob/develop/config/application.rb#L31-L32

It used the ENV hash (set via Figaro) but defaulted to using `Socket`.

whereismyjetpack:
`Socket.gethostname` isn't going to give you anything meaningful in a dynamic environment such as kubernetes, you'll get something like `scholarsphere-develop-f64bb77d8-bc4ff`

In the past i've used host headers to answer the question of `who-am-i?`

something like `@hostname = request.host` however, this would allow us to have one less configurable, and it would be dynamic based on who's coming into the system. i'm not sure that information would be availble to rails at the right time though, may return an nil object or something 

<-- not a rails wizzard. 

in CI we generate an fqdn variable that we sprinkle around to various places, we could easily inject it into an environment variable and have the application read it in as adam suggested, i'd almost lean towards not defaulting it though, since whatever Socket gives you isn't going to resolve. 

I would lean towards throwing an exception if that variable is missing.

whereismyjetpack:
inject `DEFAULT_URL_HOST` environment variable into deployments, get this value from `fqdn` this variable will also need to be in CI.

I will push application.yaml.sample changes, and CI changes. and have ryan test in his setup this afternoon

rschenk:
This latest commit fetches the `DEFAULT_URL_HOST` env variable and inserts it into Rails. I've decided to make this variable required, and the app will throw an exception on boot if it's not present. For that reason, CI might fail.

whereismyjetpack:
@rschenk this should be good now

rschenk:
Thanks @whereismyjetpack, I'll remove the "WIP" tag

PR #171, opened 2020-01-22
Endpoint for uploading new works

awead:
This is the beginning of a REST API for creating works. For now, its main use case is for migration, and as such, it's doing things that we likely will _not_ want to do outside of that. For example, you can provide any user uid, and it will create a new user with that uid. If you provide additional permissions with user uids and group names, it will create those user and group resources as well.

Currently, there is _no_ authentication for the REST API. We'll do this later.

Beyond that, there's no validity checking other than what the UI currently does in terms of title, file names, and content requirements. It assumes that all the other metadata is valid and accurate.


config/routes.rb
```diff
@@ -63,4 +63,11 @@
       get 'history', to: 'work_histories#show'
     end
   end
+
+  namespace :api do
+    resources :ingest, only: [:create], controller: 'v1/ingest'
```
rschenk: I would propose removing this route, and only supporting the versioned route

awead:
@rschenk See my comments:

> I have to admit, the way that `UserRegistrationService` masquerades the API call as an OAuth call leaves a little bit of a bad taste in my mouth. I can't think of a better way to do it off the top of my head though.

Yeah, in working through your refactoring suggestions, it's really a complicated service. This is mostly because it's going to have to do end-runs around auth proxy information for the sake of migration.

> 
> What I would like to see is a few more specs in that service. Right now, we only test the `_create` part of the `User.find_or_create`. What I'm most interested in is,
> 
>     * given a user exists, what fields are updated by the API call

I've changed this so that _no_ fields are updated with an API call. The API call will _only_ create a new user if one isn't there.

> 
>     * given a user already exists with a full name, email, etc, if an API call comes in with only the access-id, are the name and email left alone, or nil'd out?

Good catch, the were getting nil'd out. This is updated now. There's a separate update process for call made via the UI's auth proxy hash.

> 
> 
> The other thing is, I think we're fine if the ingest API is only used for data migration over by SS3. But if/when this API is used going forward, I think we need more validation. For example, right now our User model doesn't do much in the way of validation because it assumes that the data coming out of LDAP is quality. But if outside folks are able to post to the ingest API with whatever kind of access-id they decide, i.e. `"superman"` then we could start getting some bogus data in our system.

For now, the service just needs to throw in a user if it doesn't exist. There's going to need to be some kind of override situation where we need to put in arbitrary UIDs for various users. Perhaps the method there should the service queries LDAP or auth proxy to get the additional metadata. For now, when the API only provides the UID, the service is only creating the _minimal_ amount of metadata needed for the user record.

config/routes.rb
awead: @rschenk Yeah, I was thinking having a default route for the user would be easier, but it's better to be explicit.

PR #173, opened 2020-01-22
Updating yarn packages

awead:
Updates everything to the latest, and resolves GHSA-h9rv-jmmf-4pgx related to XSS vulnerabilities with serialize-javascript.


awead:
@whereismyjetpack Note the build failure `error An unexpected error occurred: "EACCES: permission denied, open '/app/yarn.lock'".` This is first time (?) we've updated our JS packages, so could just be something in the setup that we didn't run into until now. Anyway, this is not urgent, but something to fix for this sprint.

whereismyjetpack:
something went to funk-town with this lock file. i modified Dockerfile to run `yarn --frozen-lockfile` however, then rails stared barking about the integrity check, so i ran yarn, updated the lock file locally and it's kosher now? 

< magic.gif >

whereismyjetpack:
This still isn't quite right. adding app content should always come _after_ a yarn install. this allows the best level of caching for docker builds. 

I don't think the generated lockfile in this case is quite right. I'm not sure how we got here, but if i clone the repository down and run in a folder with a fresh `node_modules` directory and run an integrity check, this is what i get. i've tried various versions of node and yarn and the results are always the same. 

```yarn check --integrity            yarn-updates +
yarn check v1.21.1
warning Integrity check: Lock files don't match```

the only way to clear it up is to regenerate the lock file and push it up. I'm curious if you could blow away `node_modules` and run yarn to generate a new lock file if it fixes it.

awead:
@whereismyjetpack I don't think we want the integrity check. After rebuilding the lockfile locally with a fresh install of node_modules, the build works. Leaving in the `--frozen-lockfile` option will ensure that yarn isn't having to pull in any dependencies.

Right now, the build is failing for other reasons: test database isn't available.

whereismyjetpack:
`--frozen-lockfile` will ensure the versions of packages in lockfile are installed, without it, it's akin to running `bundle update` 

The integrity check is just a safety net. rails runs this anyway, and i'd much rather have the build break, than the deployment. 

https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile
https://github.com/yarnpkg/yarn/issues/4147

PR #181, opened 2020-01-29
Avoid Capybara::Ambiguous match error

awead:
Avoids the error by ensuring we always have unique titles when calling `:with_complete_metadata`

### DIFF
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -38,7 +38,7 @@ FactoryBot.define do
     end
 
     trait :with_complete_metadata do
-      title { Faker::Book.title }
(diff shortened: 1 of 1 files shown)


rschenk:
I think there's a better way... hang on I'm going to push another commit

rschenk:
Previously, most of the time we were using the special `:work_title` generator, which is guaranteed to give a unique result every time. _Except_ when the work_version was called with the `:with_complete_metadata` trait, for some reason. This trait is being used in the features/catalog spec that failed. I just updated this trait to use the guaranteed-to-be-unique title generator and this should cure the ambiguous title problem

PR #182, opened 2020-01-29
Work Versions are linked to Creators and Aliases 

### LINKED ISSUE #176: Move alias association from work to work version
Moves the AR association between aliases and works, to aliases and work versions. 

## TODO
* [x] make the changes in Work, WorkVersion, and WorkCreation
* [x] rename WorkCreation to WorkVersionCreation
* [x] changeup the datamodel: remove the alias table and add the alias string field to WorkVersionCreation
* [x] update factories to create works with generated creators using the Faker gem
* [x] update BuildNewWorkVersion to copy over existing alias/creators from the old version to the new

rschenk:
Closes #176


app/models/work_version.rb
```diff
@@ -28,6 +28,12 @@ class WorkVersion < ApplicationRecord
            dependent: :destroy
   has_many :file_resources,
            through: :file_version_memberships
+  has_many :creations,
```
awead: @rschenk these are really aliases, or aliased creators, although future changes could add additional fields and the alias concept might not map anymore. `creations` though, sounds weird, like the work version itself has created something.

spec/factories/creators.rb
```diff
@@ -4,8 +4,8 @@
 
 FactoryBot.define do
   factory :creator do
-    surname { Faker::Name.first_name }
-    given_name { Faker::Name.last_name }
+    given_name { Faker::Name.first_name }
+    surname { Faker::Name.last_name }
```
awead: @rschenk nice catch.

app/models/work_version.rb
rschenk: Yeah good point, creations does sound weird. I'll change that to `aliased_creators` and we can go from there.

PR #189, opened 2020-02-05
Work Creator UI on WorkVersion creation form.

### LINKED ISSUE #179: Add new creator/alias combinations to the edit form
TBD

### LINKED ISSUE #186: Alias entered into DOIs
We need to send user alias in DOIs when we create the DOI

rschenk:
Adds creator editing capability to the dashboard. Take a look through and see what you think @awead 

Closes #179 
Closes #186

### DIFF
--- a/app/controllers/dashboard/work_versions_controller.rb
+++ b/app/controllers/dashboard/work_versions_controller.rb
@@ -37,6 +37,9 @@ module Dashboard
     def edit
       @work_version = policy_scope(WorkVersion).find(params[:id])
       authorize(@work_version)
+
+      my_creator = Creator.find_or_create_by_user(current_user)
+      @work_version.build_creator_alias(creator: my_creator)
     end
 
     def update
@@ -132,7 +135,20 @@ module Dashboard
             identifier: [],
             based_near: [],
             related_url: [],
-            source: []
+            source: [],
+            creator_aliases_attributes: [
+              :id,
+              :creator_id,
+              :_destroy,
+              :alias,
+              creator_attributes: [
+                :id,
+                :email,
+                :given_name,
+                :surname,
+                :psu_id
+              ]
+            ]
           )
       end
 
--- a/app/models/creator.rb
+++ b/app/models/creator.rb
@@ -1,10 +1,26 @@
 # frozen_string_literal: true
 
 class Creator < ApplicationRecord
-  # TODO validations? Indexes?
   has_many :work_version_creations,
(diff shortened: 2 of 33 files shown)


app/models/creator.rb
```diff
@@ -1,10 +1,28 @@
...
   has_many :work_version_creations,
            dependent: :restrict_with_exception
 
   has_many :work_versions,
-           through: :work_version_creations
+           through: :work_version_creations,
+           inverse_of: :creators
+
+  validates :email,
+            presence: true
```
awead: @rschenk I'm wondering if surname should required too. One of the things Scholarsphere does is parse a display name for given and surnames: https://github.com/psu-stewardship/scholarsphere/blob/develop/app/forms/creator_form.rb#L23 That way there's at least something in the field. This will affect indexing, because we want to facet on the creator name, and not the alias. Just having an email would make that weird.

app/controllers/dashboard/work_versions_controller.rb
```diff
@@ -37,6 +37,12 @@ def show
     def edit
       @work_version = policy_scope(WorkVersion).find(params[:id])
       authorize(@work_version)
+
+      my_creator = Creator.find_or_create_by_user(current_user)
+      my_creator.readonly!
+      if @work_version.creator_aliases.none? { |ca| ca.creator == my_creator }
+        @work_version.creator_aliases.build(alias: my_creator.name, creator: my_creator)
+      end
```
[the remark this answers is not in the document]
rschenk: 100% yes, and I was thinking the same thing as you, I just didn't want to go totally hardcore with the factoring if the underlying data model turned out to be bogus. But I'm glad you said this because I was thinking the same thing

app/models/creator.rb
awead: 🏈

app/controllers/dashboard/work_versions_controller.rb
```diff
@@ -37,6 +37,9 @@ def show
     def edit
       @work_version = policy_scope(WorkVersion).find(params[:id])
       authorize(@work_version)
+
+      my_creator = Creator.find_or_create_by_user(current_user)
+      @work_version.build_creator_alias(creator: my_creator)
```
rschenk: @awead I thought you'd appreciate the Walt Whitman vibes in this line

awead: "Oh captain, my captain!"

PR #190, opened 2020-02-05
adds ability to customize solr port

whereismyjetpack:


awead:
@whereismyjetpack so to startup on, say, port 8783, it'd do: `SOLR_PORT=8783 docker-compose up solr` ?

whereismyjetpack:
yup! that should do it!

awead:
itsgood

PR #194, opened 2020-02-07
Adds CircleCI 

whereismyjetpack:


Dockerfile
```diff
@@ -1,60 +1,43 @@
...
 
-# Build targets for testing
-FROM base as rspec
-CMD [ "/app/bin/ci-rspec" ]
+# Final Target
+FROM base as production
 
-FROM base as eslint
-CMD [ "/app/bin/ci-eslint" ]
+RUN RAILS_ENV=production \
```
awead: @whereismyjetpack this is only for the build that deploys, not the one that's used for running rspec, right? RAILS_ENV would need to be "test" for the build that runs rspec.

docker-compose.test.yml
```diff
@@ -0,0 +1,34 @@
+version: '3.5' 
+services: 
+  test:
+    environment:
+      REDIS_HOST: redis
+      REDIS_PASSWORD: redispassword
+      OAUTH_APP_ID: ${OAUTH_APP_ID}
+      DRONE_COMMIT_SHA: FOO
```
awead: @whereismyjetpack should this be there?

docker-compose.test.yml
```diff
@@ -0,0 +1,34 @@
...
+      OAUTH_APP_URL: ${OAUTH_APP_URL}
+      POSTGRES_DB: scholarsphere
+      POSTGRES_USER: scholarsphere
+      POSTGRES_PASSWORD: scholarsphere
+      AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID:-scholarsphere}
+      AWS_SECRET_ACCESS_KEY: ${AWS_SECRET_ACCESS_KEY:-scholarsphere}
+      AWS_BUCKET: scholarsphere
+      AWS_REGION: us-east-1
+      S3_ENDPOINT: http://minio:9000
+      RAILS_ENV: ${RAILS_ENV:-development}
```
awead: @whereismyjetpack I'm assuming this is the test build for running rspec. Wouldn't you want to default to "test" here?

docker-compose.yml
```diff
@@ -4,7 +4,11 @@ x-web_env: &web_env
       REDIS_HOST: redis
       REDIS_PASSWORD: redispassword
       OAUTH_APP_ID: ${OAUTH_APP_ID}
+      DRONE_COMMIT_SHA: FOO
```
awead: @whereismyjetpack see my previous comment about this variable.

docker-compose.yml
```diff
@@ -74,7 +82,7 @@ services:
     volumes:
     - solr-data:/var/solr
     ports:
-    - ${SOLR_PORT:-8983}:8983
+    - 8983:8983
```
awead: @whereismyjetpack don't we want to revert this change so we can change the port easily?

spec/rails_helper.rb
```diff
@@ -5,7 +5,7 @@
 ENV['RAILS_ENV'] ||= 'test'
 
 def ci_build?
-  ENV.key?('DRONE_COMMIT_SHA')
+  ENV.key?('DRONE_COMMIT_SHA') || ENV.key?('CIRCLECI') || ENV.key?('CI')
```
awead: @whereismyjetpack take out the drone env?

Dockerfile
whereismyjetpack: for tests we only run the `base` target. the last target is to bake the assets, and ship the image!

docker-compose.test.yml
whereismyjetpack: we can certainly clean these out.

docker-compose.test.yml
whereismyjetpack: yup. good eye.

docker-compose.yml
whereismyjetpack: doh!

PR #198, opened 2020-02-11
Use singular keyword method name

### LINKED ISSUE #192: Prefer singular method names even for multi-valued fields
While fields such as descriptions, subtitles, and subjects are multivalued, their method names and json keys are not, ex:

`WorkVersion.new(subject: ["one", "two"])` stores json as `{"subject" : ["one", "two"] }`

This is consistent with the existing Scholarpshere implementation, with one exception: keywords. It is plural in Scholarsphere 4 but singular in Scholarsphere 3. Unless we feel strongly otherwise, we should make keywords a singular method, ex:

`WorkVersion.new(keyword: ["blah"])`

If we feel that a field's cardinality should be reflected in its method names and json representation, then we need to change all the other singular methods to plural ones. This latter solution will complicate migration since we'll need to translate key/value pairs.

### Option 1
* rename `keywords` method to `keyword`

### Option 2
* rename `description`, `subject`, `identifier`, etc. all to their plural equivalents

awead:
Changes the method for "Keywords" to `keyword'. This keep it internally consistent, as other multi valued terms have singular method names; and mirrors Scholarsphere 3's method names as well allowing for easier migration.

Fixes #192

### DIFF
--- a/app/components/work_version_metadata_component.rb
+++ b/app/components/work_version_metadata_component.rb
@@ -16,7 +16,7 @@ class WorkVersionMetadataComponent < ActionView::Component::Base
     :creator_aliases,
     :version_number,
     :description,
-    :keywords,
+    :keyword,
     :rights,
     :resource_type,
     :contributor,
--- a/app/controllers/api/v1/ingest_controller.rb
+++ b/app/controllers/api/v1/ingest_controller.rb
@@ -57,7 +57,7 @@ module Api::V1
             :subtitle,
             :rights,
             :version_name,
-            keywords: [],
+            keyword: [],
             description: [],
             resource_type: [],
             contributor: [],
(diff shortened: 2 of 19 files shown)


awead:
@rschenk wanted to get your opinion on this. It was the least amount of change. The alternative would be 1) keep it like is with `keywords` and rework keyword -> keywords on the migration side, or 2) rename ALL TEH THINGZ and then rework ALL TEH THINGZ on the migration side too.

Again, this seems to be the smallest change to get everything consistent and to have the simplest migration pattern moving forward.

rschenk:
I don't really have a dog in this fight. I think your PR looks good (and like a lot of work!). I also think calling the field `keywords` makes more sense intuitively. But if you want it named `keyword` then by all means, this looks like a clean way to do it

awead:
@rschenk yeah, I agree it is more intuitive, except going down that route involves changing _a lot_ of code here, and in Scholarsphere 3's migration code. I'm choosing the path of least resistance.

PR #199, opened 2020-02-11
Add creators to work history timeline

### LINKED ISSUE #195: Creators/aliases are included in Work History
Child of #149

rschenk:
I implemented this the same way as we do files, rather than trying to insert them into the metadata changes (which would be very difficult). See a screenshot below.

If this looks good to you, approve it, but I'd like to consider possibly refactoring all the `-WorkHistoryChange` components to inherit from a base class or module. We're now at the "Rule of Three" and there's a decent amount of duplicated code among  them. (I'm curious to see if Code Climate will catch this) [Update: yes, Code Climate did catch this]

Closes #195 

Screen Shot 2020-02-11 at 4 06 39 PM

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,6 +13,10 @@
 Layout/LineLength:
   Max: 154
 
+Lint/ImplicitStringConcatenation:
+  Exclude:
+    - 'app/components/work_histories/paper_trail_change_base_component.rb'
+
 # Offense count: 1
 # Cop supports --auto-correct.
 Lint/NonDeterministicRequireOrder:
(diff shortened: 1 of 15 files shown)


rschenk:
@awead we are ready to rock post-refactor.

CodeClimate is [still complaining](https://codeclimate.com/github/psu-stewardship/scholarsphere-4/app/components/work_histories/work_history_component.rb?from_sha=99b0f586&to_sha=7b760e70) however, because these two methods are similar:

```
      def changes_to_work_versions_files(work_version)
      def changes_to_creator_aliases(work_version)
```

This should be marked as invalid or won't-fix. Yes these methods are similar, no refactoring them will not make the code cleaner or more understandable. What they do and where they're called is high on the WTF scale, so having the repetition IMHO is not a bad thing whatsoever.

PR #208, opened 2020-02-24
Install and configure Questioning Authority

### LINKED ISSUE #93: Controlled Vocabularies with Questioning Authority
The Questioning Authority gem (QA) is a standard gem used in the Samvera community for interacting with different controlled vocabularies. Presently, we use controlled vocabularies for subjects, locations, and languages, but we don't take advantage of doing this via QA.

Adopting QA as the unified interface for different vocabularies will help us manage them more easily, and adapt to changes in the future.

https://github.com/samvera/questioning_authority

awead:
Installs the qa gem and configures the geonames endpoint for use. The geonames search is provided by the gem and only requires a username in order to work. We're using the same setup that we use in Scholarsphere 3, but with a configurable username via the environment.

Note: There is no UI autocomplete integration yet since we're still working out changes to the overall form. Those bits will come with a later PR.

Fixes #93


config/oclcts-authorities.yml
```diff
@@ -0,0 +1,24 @@
...
+  fast:
+    name: Faceted Application of Subject Terminology (FAST subject headings)
+  gsafd:
+    name: Form and genre headings for fiction and drama
+  lcshac:
+    name: Library of Congress AC Subject Headings
+  lcsh:
+    name: Library of Congress Subject Headings
+  mesh:
+    name: Medical Subject Headings (MeSH®)
```
rschenk: [MeSH terms are my jam](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2815376/)

PR #209, opened 2020-02-24
Support for Scholarsphere-3 URLs

### LINKED ISSUE #166: Support for existing URLs
The current scholarsphere urls should resolve to the same object with new URL scheme in scholarsphere 4.

rschenk:
Here's a sketch for how to support old URLs with a minimum of coupling:

* `LegacyIdentifer` model contains a version number (which version of ScholarSphere this ID came from), an `old_id` string (e.g. the NOID), and a polymorphic ActiveRecord association to the corresponding thing in SS-4
* All old _routes_ are handled in `config/routes.rb` and a controller that uses the above model

### Open Questions
* Do we need support for files? That could get tricky and make this fall apart.

Closes #166

### DIFF
--- a/app/controllers/legacy_urls_controller.rb
+++ b/app/controllers/legacy_urls_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class LegacyUrlsController < ApplicationController
+  # Handles legacy urls from Scholarsphere v3
+  def v3
+    uuid = LegacyIdentifier.find_uuid(version: 3, old_id: params[:id])
+    redirect_to resource_path(uuid)
+  end
+end
--- a/app/models/legacy_identifier.rb
+++ b/app/models/legacy_identifier.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class LegacyIdentifier < ApplicationRecord
+  belongs_to :resource, polymorphic: true
+
+  def self.find_uuid(version:, old_id:)
+    legacy_id = find_by(version: version, old_id: old_id)
+
+    legacy_id&.resource&.uuid ||
+      raise(ActiveRecord::RecordNotFound)
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -71,4 +71,7 @@ Rails.application.routes.draw do
       resources :ingest, only: [:create]
     end
   end
+
+  # Legacy URL support
+  get '/concern/generic_works/:id', to: 'legacy_urls#v3'
 end
--- a/spec/request/legacy_urls_spec.rb
+++ b/spec/request/legacy_urls_spec.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe LegacyUrlsController, type: :request do
+  let(:resource) { create :work_version }
+
+  describe 'Scholarsphere v3 legacy URLs' do
+    before do
+      create :legacy_identifier,
+             version: 3,
+             old_id: 'old-v3-id-123',
+             resource: resource
+    end
+
+    it do
+      get '/concern/generic_works/old-v3-id-123'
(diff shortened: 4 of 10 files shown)


app/controllers/legacy_urls_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class LegacyUrlsController < ApplicationController
+  # Handles legacy urls from Scholarsphere v3
+  def v3
+    uuid = LegacyIdentifier.find_uuid(version: 3, old_id: params[:id])
+    redirect_to resource_path(uuid)
```
rschenk: Code Climate thinks this is untested, but that's not true: ` spec/request/legacy_urls_spec.rb`

app/models/legacy_identifier.rb
```diff
@@ -0,0 +1,12 @@
...
+
+class LegacyIdentifier < ApplicationRecord
+  belongs_to :resource, polymorphic: true
+
+  def self.find_uuid(version:, old_id:)
+    legacy_id = find_by(version: version, old_id: old_id)
+
+    legacy_id&.resource&.uuid ||
+      raise(ActiveRecord::RecordNotFound)
+  end
```
rschenk: Code Climate seems to think this is untested, even though it totally is `spec/models/legacy_identifier_spec.rb`

app/models/work_version.rb
```diff
@@ -40,6 +40,10 @@ class WorkVersion < ApplicationRecord
            through: :creator_aliases,
            inverse_of: :work_versions
 
+  has_many :legacy_identifiers,
+           dependent: :destroy,
+           inverse_of: :resource
+
```
awead: @rschenk Work versions won't have legacy ids, only works.

config/routes.rb
```diff
@@ -71,4 +71,8 @@
       resources :ingest, only: [:create]
     end
   end
+
+  # Legacy URL support
+  # @todo what are all the URLs supported by Sufia/Scholarsphere-3?
+  get '/concern/generic_works/:id', to: 'legacy_urls#v3'
```
awead: @rschenk I think collections are the only other resources that have/need resolvable urls.

spec/request/legacy_urls_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it do
+      get '/concern/generic_works/old-v3-id-123'
+      expect(response).to redirect_to resource_path(resource.uuid)
+    end
+
+    context 'when given a nonexistent id' do
+      it do
+        expect {
+          get '/concern/generic_works/doesnt-exist'
+        }.to raise_error(ActiveRecord::RecordNotFound)
```
awead: @rschenk shouldn't this just be a 404?

spec/request/legacy_urls_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        create :legacy_identifier,
+               version: 2,
+               old_id: 'old-v2-id-222',
+               resource: resource
+      end
+
+      it do
+        expect {
+          get '/concern/generic_works/doesnt-exist'
+        }.to raise_error(ActiveRecord::RecordNotFound)
```
awead: @rschenk same question as above... The application might be raising the error, but the response from the controller ought to be 404.

spec/request/legacy_urls_spec.rb
rschenk: It will be, that's handled by a Rails middleware thing. All `ActiveRecord::RecordNotFound` are caught and turned into 404s automagically in productionesque environments

rschenk:
@awead Yes, during the migration you'll have to do something like....

```ruby
my_new_work = #...
my_new_work.legacy_identifiers.find_or_initialize_by(version: 3, old_id: api_params[:noid] )
```

PR #218, opened 2020-02-28
Promote and destroy files asynchronously

### LINKED ISSUE #217: Make storage promotion async
The process of promoting Shrine storage out of the cache and into its proper location should be handled via ActiveJob as an async process. This should fix timeouts for large upload requests. See https://github.com/psu-stewardship/ss-shrine/blob/master/app/jobs/promote_job.rb for details, but it's basically like we did it there.

## WDLL
* define a promotion job
* add job to FileUploader class as per Shrine documentation
* re-run migration of the ant videos in QA and see if get 504 errors

awead:
Adds jobs for promoting files from cache to store, as well as deleting files from storage.

Fixes #217

### DIFF
--- a/app/jobs/shrine/destruction_job.rb
+++ b/app/jobs/shrine/destruction_job.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class Shrine::DestructionJob < ApplicationJob
+  queue_as :shrine
+
+  def perform(data:)
+    attacher = FileUploader::Attacher.from_data(data)
+    attacher.destroy
+  end
+end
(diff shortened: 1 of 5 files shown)


whereismyjetpack:
We will need changes to `start.sh` to conditionally run sidekiq, and I'm curious if we should give sidekiq a default configuration at config/sidekiq.yml did you want me to push those to this branch, or give a separate PR?

awead:
@whereismyjetpack I'd do it as a separate PR. Rails will actually run async without sidekiq, it just wouldn't be very good; a kinda of "faked" async. So we can tweak the Sidekiq config and setup completely separately from PR here.

awead:
@rschenk One thing I did here was use the `Shrine` namespace. This could be construed as "bad" since I'm defining something like `Shrine::DestructionJob` and Shrine could technically create the same class in its gem code and that might cause problems. But, I wanted to use the namespace because it was there and kept things organized.

PR #220, opened 2020-03-03
Adds default sidekiq config; starts sidekiq if APP_ROLE == 'sidekiq'

whereismyjetpack:
Sets a "sane" default config for sidekiq and runs `/app/bin/sidekiq` instead of `rails s` if the containers APP_ROLE is set to "sidekiq"

Related: #217

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -2,7 +2,6 @@ set -e
 
 if [ ${RAILS_ENV:-development} != "production" ]; then
   bundle check || bundle
-  yarn
 fi
 
 if [ ${APP_ROLE:-app} == "sidekiq" ]; then
--- a/start.sh
+++ b/start.sh
@@ -1,12 +1,18 @@
-set -e 
+set -e
 
 if [ ${RAILS_ENV:-development} != "production" ]; then
   bundle check || bundle
+  yarn
 fi
 
-echo "starting rails"
-rm -f tmp/pids/server.pid
-bundle exec rails db:create
-bundle exec rails db:migrate
-bundle exec rails solr:init
-bundle exec rails s -b '0.0.0.0'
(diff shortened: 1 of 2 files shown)


start.sh
```diff
@@ -1,12 +1,18 @@
...
 
-echo "starting rails"
-rm -f tmp/pids/server.pid
-bundle exec rails db:create
-bundle exec rails db:migrate
-bundle exec rails solr:init
-bundle exec rails s -b '0.0.0.0'
+if [ ${APP_ROLE:-app} == "sidekiq" ]; then
+    echo "starting sidekiq"
+    /app/bin/sidekiq
```
awead: @whereismyjetpack I'm assuming this means that if there's a docker container (?) running with the APP_ROLE of sidekiq, then it just starts sidekiq?

start.sh
```diff
@@ -1,12 +1,18 @@
-set -e 
+set -e
 
 if [ ${RAILS_ENV:-development} != "production" ]; then
   bundle check || bundle
+  yarn
```
awead: @whereismyjetpack this doesn't seem related to sidekiq. Is it?

start.sh
whereismyjetpack: This is correct. the deployment for sidekiq sets that variable, and then runs as that "role" this allows both "roles" to share the same entrypoint to bootstrap the environment

start.sh
whereismyjetpack: removed.

awead: @whereismyjetpack cool. Does it belong someplace else? I don't want to loose track of it if it is... we can ticket it or something if needed.

whereismyjetpack: I'll submit it as another PR!

PR #222, opened 2020-03-04
Update puma to a version that exists in rubygems

rschenk:


awead:
DependaBot beat you to this: #221

PR #226, opened 2020-03-05
Download files from S3

awead:
Uses the ResourcesController to download a given file from a work version. Specifying a work will result in a 404 since technically the files are attached to the work version and not the work.

We _could_ get fancy and return the files from the latest published version when specifying a work, or alternatively redirect to the work's display page.

## URLs

We're using the current resources controller which displays works and work versions:

`https://scholarpshere/resources/xxxxxxxx-yyyy-zzzz-vvvv-aaaaaaaa/content/1`

Other options include:

* `https://scholarpshere/resources/xxxxxxxx-yyyy-zzzz-vvvv-aaaaaaaa`
    * the resources controller would make the determination if the uuid was a Work, WorkVersion, or file resource and respond accordingly.

* `https://scholarpshere/resources/xxxxxxxx-yyyy-zzzz-vvvv-aaaaaaaa/files/1/download`
    * more RESTful
    * allows for a future `/files` and `/files/1` to list files and show individual ones

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class DownloadsController < ApplicationController
+  def content
+    work_version = WorkVersion.find_by!(uuid: params[:resource_id])
+    file = work_version.file_version_memberships.find(params[:id])
+    authorize(file)
+    redirect_to s3_presigned_url(file)
+  end
+
+  private
+
+    def s3_presigned_url(file)
+      file.file_resource.file_url(
+        expires_in: ENV.fetch('DOWNLOAD_URL_TTL', 6),
+        response_content_disposition: ContentDisposition.inline(file.title)
+      )
+    end
+end
--- a/app/policies/file_version_membership_policy.rb
+++ b/app/policies/file_version_membership_policy.rb
@@ -8,6 +8,10 @@ class FileVersionMembershipPolicy < ApplicationPolicy
     false
   end
 
+  def content?
+    download?
+  end
+
   private
 
     # @todo There's a bug in the permissions because a depositor should have edit access by default
--- a/app/views/shared/_work_version_files.html.erb
+++ b/app/views/shared/_work_version_files.html.erb
@@ -10,7 +10,7 @@
     <tbody>
       <% work_version.file_version_memberships.includes(:file_resource).each do |file| %>
       <tr>
-        <td><%= file.title %></td>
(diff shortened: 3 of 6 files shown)


app/controllers/resources_controller.rb
```diff
@@ -1,17 +1,40 @@
...
+  before_action :load_resource
+
   def show
-    @resource = find_resource(params[:id])
-    authorize @resource
+    authorize(@resource)
+  end
+
+  def download
+    if @resource.is_a?(WorkVersion)
```
rschenk: This kind of type checking would be unnecessary if we split downloads into its own `DownloadsController`

app/controllers/resources_controller.rb
```diff
@@ -1,17 +1,40 @@
...
   end
 
   private
 
+    def load_resource
+      @resource = if params[:resource_id]
+                    find_resource(params[:resource_id])
+                  else
+                    find_resource(params[:id])
+                  end
```
rschenk: This kind of type checking would be unnecessary if we split downloads into its own `DownloadsController`

awead:
@rschenk I've updated it to use the DownloadsController and made a slight change to the urls:
`/resources/d2163da3-9c21-479a-85df-2f5c4b10e5b9/downloads/2`
I kept `content` as the action name since `show` didn't seem to apply and `downloads#download` annoys me :)

app/controllers/downloads_controller.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DownloadsController < ApplicationController
+  def content
+    work_version = WorkVersion.find_by!(uuid: params[:resource_id])
+    file = work_version.file_version_memberships.find(params[:id])
+    authorize(file)
+    redirect_to file.file_resource.file_url(expires_in: ENV.fetch('DOWNLOAD_URL_TTL', 6))
```
whereismyjetpack: I'm curious if we should use `response_content_disposition: "attachment"` here; this _should_ inform the browser to initiate a download for files it knows about such as txt, pdf, jpeg?

awead: The pre-signed url is already doing that somehow. See below. The Amazon url (200 GET) has the disposition set to return the original filename.
Screen Shot 2020-03-06 at 3 11 13 PM

awead: After a slack discussion with @srerickson, we'll stick with the inline default, but can revisit later.

awead:
I've made the requested changes, and am approving since @rschenk is out-of-town.

PR #240, opened 2020-04-01
adds json logging

whereismyjetpack:
with no configurations: (i.e lograge_enabled: false, rails_log_json: false)
Screen Shot 2020-04-01 at 8 25 22 AM

with only JSON logging (i.e lograge_enabled: false, rails_log_json: true)
Screen Shot 2020-04-01 at 8 29 18 AM

with lograge and JSON logging (i.e lograge_enabled: true, rails_log_json: true)
Screen Shot 2020-04-01 at 8 42 30 AM

the second option with json, and without lograge isn't terribly useful? I thought I'd open this PR for discussion and see what others think.


rschenk:
I've never seen or tried this before, but seems neat.

app/controllers/application_controller.rb
```diff
@@ -11,4 +11,11 @@ class ApplicationController < ActionController::Base
   def current_user
     super || User.guest
   end
+
+  def append_info_to_payload(payload)
+    super
+    payload[:request_id] = request.env["action_dispatch.request_id"]
+    payload[:x_forwarded_for] = request.env["HTTP_X_FORWARDED_FOR"]
+    payload[:remote_addr] = request.env["REMOTE_ADDR"]
```
awead, line 19: @whereismyjetpack If I'm reading the docs right, isn't this is the same as what's happening in ` config/initializers/lograge.rb`? I prefer the initializer method as opposed to overriding a method, if we can do that.

lib/logging/json_formatter.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'json'
+
+class JSONFormatter
```
awead, line 5: This would need to be `Logging::JSONFormatter` given the directory structure. Maybe just move it up to `lib/json_formatter.rb`

lib/logging/json_formatter.rb
```diff
@@ -0,0 +1,23 @@
...
+  def parse_message(message)
+    begin
```
awead, line 7: `begin` is unnecessary here.

config/application.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'boot'
+require_relative '../lib/logging/json_formatter.rb'
```
awead, line 4: @whereismyjetpack if you make the file location `lib/json_formatter.rb` than this can be written as `require 'json_formatter'`

app/controllers/application_controller.rb
whereismyjetpack: I _think_ this is an ordering thing, I don't think the request ID is available within the intializer, but it is here. I put all the items here so they'd at least be in the same place.

lib/logging/json_formatter.rb
whereismyjetpack: I think I understand this, however not sure how it works now then. Is Logging::ClassName just mor explicit to avoid name collisions?

awead: How is this different from `Lograge::Formatters::Json`? Would it be worth extending from it? i.e.
``` ruby
class JSONFormatter < Lograge::Formatters::Json
end
```

awead: @whereismyjetpack can you also put in a spec test for the logger? It's lacking test coverage.

app/controllers/application_controller.rb
awead: The docs say "Alternatively, you can add a hook for accessing controller methods directly ..." Have you tried it without the change to the controller to see if it works?

whereismyjetpack: I thought I did. but i'll go back and try again!

whereismyjetpack: the controller method did the trick. I believe this to be resolved.

lib/logging/json_formatter.rb
whereismyjetpack: I believe this to be resolved. I don't think my spec file is terribly idiomatic, So i'm happy to go back and make some changes.

lib/logging/json_formatter.rb
whereismyjetpack: I did not know this. I was able to squash this down quite a bit!

awead: @whereismyjetpack yeah, it's one of the "ruby-isms"

config/application.rb
```diff
@@ -37,6 +38,16 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
+    # Logging.
+    config.lograge.enabled = true if ENV['LOGRAGE_ENABLED'].present?
```
awead, line 42: @whereismyjetpack one last point: you can move this line under 44 below and remove the `if`

awead:
@whereismyjetpack this is looking great! I made one more comment. Otherwise, this ready to go.

config/application.rb
whereismyjetpack: these are two different configuration options, I think we could probably ditch the `LOGRAGE_ENABLED` wasn't sure.

PR #245, opened 2020-04-06
Rename JSONFormatter to JSONLogFormatter

rschenk:
I think `JSONFormatter` is too generic of a name, especially at the top level namespace. This app can and will format many things as JSON, so this is the simplest fix I could think of to clamp things down a little bit.


rschenk:
https://www.youtube.com/watch?v=vLbhkM0LaYI

PR #247, opened 2020-04-07
Rename Creator model to Actor

rschenk:


app/models/work_version.rb
```diff
@@ -37,6 +37,7 @@ class WorkVersion < ApplicationRecord
            dependent: :destroy
 
   has_many :creators,
+           source: :actor,
```
awead, line 40: @rschenk It looks like this is the only place where were keep the `creators` method and return an Actor objects. If so, I think that makes sense. In looking over the rest of the PR, I don't think there's any other place we'd _want_ to do that. Every place else, it's going to be `actor` instead of `creator`

rschenk: Yup, that was my thinking as well. And you're right, this is the only place where we have a method called `#creators` that returns back actors, because in this context, those actors are the creators of the work version. It's hard to be both specific and generic at the same time

PR #249, opened 2020-04-07
Fix Blacklight's search API

awead:
Returning search results in json was broken because we've moved away from SolrDocument and we weren't testing the integration endpoint.

We're now testing json responses from the CatalogController and overriding the jbuilder file to make it work for now, but there's a better solution down the road.


whereismyjetpack:
tested on `scholarsphere-migration`; lgtm

PR #263, opened 2020-04-17
Swap order of SQL delete statements to (hopefully) make FK constraints happy

rschenk:


whereismyjetpack:
This code is now running in our "qa" environment and the migrations ran successfully

PR #268, opened 2020-04-23
combine linting / don't background linting

### LINKED ISSUE #267: Niftany is not failing the CI build
The Nitfany step fails, but the build is still passing:

https://circleci.com/gh/psu-stewardship/scholarsphere-4/374

whereismyjetpack:
in circle when `backgrounding: true` is set the status doesn't influence the build. As a fix I combined the two linting stages into one, and disabled backgrounding. 

Fixes #267

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,16 +28,10 @@ jobs:
         command: |
           docker build -t $REGISTRY_URL:$CIRCLE_SHA1 --target base .
     - run:
-        name: "Niftany"
-        background: true
+        name: "Linting"
         command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,16 +28,10 @@ jobs:
         command: |
           docker build -t $REGISTRY_URL:$CIRCLE_SHA1 --target base .
     - run:
-        name: "Niftany"
-        background: true
+        name: "Linting"
```
awead, line 31: I think the Niftany name is more appropriate since it should be running the `bundle exec niftany` command.

PR #286, opened 2020-05-05
Collections are indexed into solr and have ACL permissions

### LINKED ISSUE #271: Index Collections
Collections need to be indexed into solr

rschenk:
Closes #271

### DIFF
--- a/app/indexers/collection_indexer.rb
+++ b/app/indexers/collection_indexer.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class CollectionIndexer
+  def self.call(collection, commit: false)
+    IndexingService.add_document(collection.to_solr, commit: commit)
+  end
+end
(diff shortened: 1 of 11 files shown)


app/schemas/collection_schema.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CollectionSchema < DefaultSchema
+  def document
+    super.merge(creator_schema.document)
```
awead, line 5: Does this avoid having to do:
``` ruby
def document_builder                                                                                                
  SolrDocumentBuilder.new(                                                                                          
    WorkVersionSchema,                                                                                              
    CreatorSchema                                                                                                   
  )                                                                                                                 
end
```
Like we do in WorkVersion?

awead:
@rschenk FYI: you had an extra revert commit in here for some reason. I rebased off of master and you're all set.

PR #288, opened 2020-05-08
🐞 Collection policy respects ACL

rschenk:
It occurred to me that although we implemented ACLs for collections, I never made the public-facing authorization policy use them.


awead:
@rschenk looks good. I'd feel better if there was a feature test that ran this. We could add a AUTHORIZED, Penn State only collection to discover_search_spec.rb and make sure it doesn't show up.

rschenk:
@awead Added collections to the discovery feature spec (and did a git squash / force push)

PR #292, opened 2020-05-08
Niftany

awead:
Applies changes to update Niftany (and Rubocop) to their latest versions.

@rschenk there's some funny business going on with Rubocop's uniqueness validators.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -154,12 +154,12 @@ GEM
     docile (1.3.2)
     down (5.0.0)
       addressable (~> 2.5)
-    erb_lint (0.0.31)
+    erb_lint (0.0.32)
       activesupport
       better_html (~> 1.0.7)
       html_tokenizer
       rainbow
-      rubocop (~> 0.79.0)
+      rubocop (~> 0.79)
       smart_properties
     erubi (1.9.0)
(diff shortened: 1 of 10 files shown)


awead:
@rschenk Rubocop is failing the uniqueness check on AddIndexOnAccessLevelToAccessControls but as far as I can tell, it seems right to me.

rschenk:
I did some digging into Rubocop, and I think this is actually a bug/oversight in Rubocop.

Check out this output, which I inserted into Rubocop's validator thingy:
```
$ bundle exec rubocop app/models/access_control.rb
Inspecting 1 file
================================Rubocop-Index=================================
  Rubocop's expected column names: 
    #<Set: {"access_level", "agent_id", "resource_id"}>
  Indexed columns:
    #<Set: {"access_level", "agent_type", "resource_type", "agent_id", "resource_id"}>
------------------------------------------------------------------------------
```

Because these are polymorphic relationships, it's not enough to _just_ index the `*_id`. For example, let's just look at Agents, supposing you had....
agent_type | agent_id
----------- | ---------
Group | 123
User | 123

If you only index by agent_id, this would fail the uniqueness constraint, which is improper.

tldr; you are right and rubocop is wrong. So there.

PR #299, opened 2020-05-12
Allow ACLs to inherit other levels of access

### LINKED ISSUE #293: Inherited Permissions
How should we represent the notion of inherited permissions?

The ranking of access levels are such that each successive level includes the previous one, so discovery < read < edit. This means that if a user has read access to a work, they implicitly have discovery access as well. Presently, however, the methods don't reflect this:

```
>  work.read_access? 'agw13'
=> true
>  work.discovery_access? 'agw13'
=> false
>  work.read_users
=> ['agw13']
>  work.discovery_users
=> []
```

## Option 1

Write additional ACLs for every inherited permission. For example, defining edit access to a user would write three ACLs, one each for discover, read, and edit access. This would result in:

```
>  work.edit_access? 'agw13'
=> true
>  work.read_access? 'agw13'
=> true
>  work.discovery_access? 'agw13'
=> true
>  work.edit_users
=> ['agw13']
>  work.read_users
=> ['agw13']
>  work.discovery_users
=> ['agw13']
```

This has the advantage of "baking in" the inheritance into the ACL structure and making it explicit, but the adding and removing ACLs gets messier.

## Option 2

Use application logic. We're presently doing this with Solr, although we're missing the edit dimension: https://github.com/psu-stewardship/scholarsphere-4/blob/master/app/schemas/permissions_schema.rb#L8-L9

Another example would be in `WorkPolicy#show?` If we want discovery access to allow viewing a  record, then anyone with read or edit access would have the same ability:

``` ruby
def show?
  work.discover_access?(user) || work.read_access?(user) || work.edit_access?(user)
end
```

This would reduce the number of ACL records written, but might puts the onus on the application to enforce the inheritance.

awead:
Edit access should include read and discover access, and read access should include discover access. Instead of using application logic to effect this, additional ACLs are added that explicitly define these inherited access levels.

Additionally, a `white_list` function is provided that serves to prevent specific kinds of agents from being assigned using the `[]=` methods available for each combination of agent and access level. This applies to public and authorized group agents, so that they are not inadvertenly removed when setting permissions for other agents.

Depositors are no longer checked when assigning permissions. This means that the depositor will NOT appear as an edit user, and it's possible to create an ACL for the depositor. However, these issues will be handled within the application and the permissions module no longer has any opinions about the depositor.

Fixes #293


awead:
@rschenk I'm going to put in a ticket. I think this would be a good health check to add to OkComputer.

PR #300, opened 2020-05-14
Migrating Collections

awead:
Enables migrating collections from Scholarsphere 3. It replicates the same kind of pattern used in migrating works, and code is very similar in both places. This might need additional refactoring.

### DIFF
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+module Api::V1
+  class CollectionsController < RestController
+    rescue_from StandardError do |exception|
+      if exception.is_a?(ActionController::ParameterMissing)
+        render json: { message: 'Bad request', errors: [exception.message] }, status: :bad_request
+      else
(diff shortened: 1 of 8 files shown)


awead:
@rschenk I'm holding off on any refactoring since the API code will be changing a lot post-migration.

app/controllers/api/v1/collections_controller.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    private
+
+      def return_migrated_collection
+        ids = LegacyIdentifier.where(old_id: metadata_params['noid'], version: 3, resource_type: 'Collection')
+        return if ids.empty?
+
+        collection = Collection.find(ids.first.resource_id)
+        render json: migrated_response(collection), status: 303
+      end
```
rschenk, line 32: Could we do this like...

```ruby
def return_migrated_collection
  id = LegacyIdentifier.find_by(old_id: metadata_params['noid'], version: 3, resource_type: 'Collection')
  collection = id.resource
  render json: migrated_response(collection), status: 303
rescue ActiveRecord::RecordNotFound
  return # or just have nothing here
end
```

PR #303, opened 2020-05-16
Move all stylesheet and image assets to webpacker

awead:
This refactors and reorganizes our webpacker code, moving in all our other stylesheet and image assets.

The primary reason for this was because we were installing Bootstrap intwo places: the gem (for stylesheets), and the npm package (for javascript). A recent jQuery security vulnerability exposed the flaw with this setup: We could update jQuery (and consequently Boostrap) in the npm package to address the vulnerability, but not the gem.

The solution is remove the asset pipeline altogether and the Bootstrap gem, and only use webpacker. This necessitated some reorganizing of the app/javascripts directory.


awead:
Resulting layout looks like this:
```
app/javascript
├── blacklight
│   ├── index.js
│   └── styles
│       ├── _header.scss
│       └── index.scss
├── channels
│   ├── consumer.js
│   └── index.js
├── controllers
│   ├── filename_editor_controller.js
│   ├── index.js
│   ├── multiple_fields_controller.js
│   └── uppy_controller.js
├── packs
│   └── application.js
└── scholarsphere
    ├── images
    │   ├── visibility-authorized.png
    │   ├── visibility-open.png
    │   └── visibility-private.png
    ├── index.js
    └── styles
        ├── _diffy.css
        ├── _forms.scss
        ├── _progress_indicator.scss
        ├── _version_status_badge.scss
        ├── _visibility_badge.scss
        ├── _work_history.scss
        ├── _work_versions.scss
        ├── _works.scss
        └── index.scss
```
`packs/application.js` is called first. This is webpacker's default action. `application.js` then loads the other directories or "packs": blacklight, channels, controllers, and scholarsphere. The `index.js` of each directory loads the additional files within each pack.

awead:
@rschenk Yes, this is true. The solution to the gem-that-has-stuff-that-needs-to-go-in-the-asset-pipeline problem would be to add the gem to the `package.json` file. We're doing this already with twitter-typeahead-rails:

https://github.com/psu-stewardship/scholarsphere-4/blob/master/package.json#L18

The repo gets cloned to `node_modules/twitter-typeahead-rails` and we can reference its assets here:

https://github.com/psu-stewardship/scholarsphere-4/blob/master/app/javascript/blacklight/index.js#L6

It's a bit strange, and should the repo's location of the assets change, you'd need to update the path AND I'm not sure if yarn can calculate dependency issues with git repos. But at least it's a way to avoid copying over assets manually.

rschenk:
@awead Maybe we could write this move up as part of our Architecture Decision List or document it in the readme or wiki?

PR #316, opened 2020-05-19
Update RSolr to 2.3.0

awead:
The update broke our tests for SolrAdmin. Switching over to VCR instead of WebMock fixes this and makes the tests a bit easier to deal with in the future.


awead:
Note: getting erb to work with VCR yaml files is kind of a pain, and doesn't completely gel with the new `:vcr` option. If you look at the VCR source code it seems like doing `vcr: { erb: { foo: 'bar } }` would work, but it doesn't. It may be a bug since the `:vcr` tag is a relatively new thing.

PR #318, opened 2020-05-20
Move Lograge configuration to the production environment

awead:
This maintains standard Rails logging for development and test, keeping the production configuration separate from the other environments.


whereismyjetpack:
I kind of like being able to test the loggers in the development environment, I wonder if wrapping the initializer in the `LOGRAGE_ENABLED` would have the best of both worlds?

awead:
@whereismyjetpack the problem is we're not seeing the sql output to the logs. Also, lograge's output is designed for computer consumption, and not human.

I don't see you testing anything specific to lograge. We are testing JSONLogFormatter though.

PR #340, opened 2020-05-28
Manage work types using a string enum

awead:
Instead of using the standard integer to mark an enum's value, use a actual string. This way the value in the database is more informative, and we get all the bonus parts of the enum feature.


awead:
@rschenk I'm thinking this will be an easier way to manage the work types. The only question is how we want to render a display value in `WorkVersionMetadataComponent`.

PR #342, opened 2020-05-29
Specify UTC timezone for embargo date tests

### LINKED ISSUE #339: Embargo tests fail after 10 pm
`spec/components/embargo_detail_component_spec.rb` tests are failing after 10 PM because of differences in time zone. We need to correct the tests to take into account differences of time zone.

awead:
Because Rails converts to UTC time by default, the times were off by a few hours (Eastern time versus zulu time) and the test would only catch the difference late at night.

Fixes #339 

Question: Do we want to configure the application to store timestamps in local time?

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -114,7 +114,7 @@ class Work < ApplicationRecord
   def embargoed?
     return false if embargoed_until.blank?
 
-    embargoed_until > DateTime.now
+    embargoed_until > Time.zone.now
   end
 
   private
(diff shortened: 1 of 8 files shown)


rschenk:
@awead Using `DateTime.now` and `Time.now` in a Rails application is an anti-pattern. You need to always use `Time.zone.now` or `Time.zone.local`. This is due to Ruby not having any timezone support for years (decades) and Rails basically monkey patching it on.

PR #345, opened 2020-06-01
Edtf dates

### LINKED ISSUE #140: WorkVersion's published_date is a singular field and validated
`WorkVersion#published_date` is currently a multi-value field. We agreed in #101 that [it should be singular](https://github.com/psu-stewardship/scholarsphere-4/issues/101#issuecomment-566184244).

### Todo
- [x] Determine how many legacy ScholarSphere works have multiple values in this field.
- [x] Figure out how we're going to reduce those multiple values into a single one when we port the data over
- [x] Make `WorkVersion#published_date` a singular field

### Bonus Todo?
- [x] Enforce a standard format for storing dates as strings (date-picker on UI?). @srerickson suggests ISO8601, but will this handle every scenario we need to address?
- [x] Validate that `published_date` is both present and in the correct format when a WorkVersion is published
- [x] Ensure that the Solr document/indexer can properly "explode" an imprecise date out into a date range, e.g. "2019" becomes "2019-01-01 00:00:00" - "2019-12-31 23:59:59" (Ryan did this in CHO for EDTF dates. Yoink that code?)

### Open Questions
1. Does ISO8601 cover all the use cases that we have?

rschenk:
* published_date is a single-value field for Work Versions and Collections
* published_date, if provided, will be "exploded" in the Solr document to cover its entire possible range, for example "'202x'" will be exploded to "[2020 TO 2029]" in the Solr index
* EDTF-style dates will be humanized when displayed in the web ui, for example "202x" will be displayed as "2020s"

On Work Versions:
* published_date is not required to save-as-draft in the web-ui
* published_date is _required_ to publish in the web-ui
* published_date must be a valid EDTF format to publish in the web-ui
* the migrations/api does not validate this field, it can put anything they want in this field, including nothing

On Collections:
* published_date is not required to create a collection in the web-ui (there is no draft/published state for collections)
* in the web-ui _if_ you enter a value for published_date, then it must be a valid EDTF-format
* the migrations/api does not validate this field, it can put anything they want in this field, including nothing

Closes #140


app/components/collection_metadata_component.rb
```diff
@@ -31,11 +31,17 @@ class CollectionMetadataComponent < ActionView::Component::Base
...
-    @collection = collection
+    @collection = decorate(collection)
   end
 
   private
 
+    def decorate(col)
+      return col if col.is_a? ResourceDecorator
+
+      ResourceDecorator.new(col)
```
awead, line 42: Can we just make all the collection displays use the ResourceDecorator? We're already doing it in the resources show page. Is the only other place is the dashboard view?

app/components/work_version_metadata_component.rb
```diff
@@ -39,12 +39,18 @@ class WorkVersionMetadataComponent < ActionView::Component::Base
...
     @mini = mini
   end
 
   private
 
+    def decorate(work_ver)
+      return work_ver if work_ver.is_a? ResourceDecorator
+
+      ResourceDecorator.new(work_ver)
+    end
```
awead, line 52: Ditto for showing collections... Dashboard::WorkVersionDecorator inherits from ResourceDecorator, so we should be able to the same decorator throughout.

spec/decorators/dashboard/work_version_decorator_spec.rb
```diff
@@ -40,7 +40,7 @@
     context 'with a published version' do
       before { allow(work_version).to receive(:draft?).and_return false }
 
-      its(:display_date) { is_expected.to eq 'Published November 04, 2019' }
```
awead, line 43: Shouldn't we checking the Nov. 4th date too?

app/services/edtf_date.rb
```diff
@@ -12,4 +12,10 @@ def self.valid?(str)
 
     (regular_date || edtf_date).present?
   end
+
+  def self.humanize(str)
+    Date.edtf(str)&.humanize || str.to_s
+  rescue RuntimeError
+    str.to_s
+  end
```
awead, line 20: This should ensure that the migrated pubished dates, which may not be valid EDTF, will display, right?

.rubocop_todo.yml
```diff
@@ -27,7 +27,12 @@ Lint/NonDeterministicRequireOrder:
 # Offense count: 4
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 135
+  Max: 140
+
+# Offense count: 1
+RSpec/AnyInstance:
+  Exclude:
+    - 'spec/components/embargo_detail_component_spec.rb'
```
awead, line 35: This can go away... I disabled this in `.rubocop.yml` and probably got missed in a rebase.

rschenk:
@awead I moved the validation context into the migration api controller and service objects in 
39772fe. I also slightly changed how I was doing the module namespace in EdtfDate::SolrDateRangeFormatter in c0ed586

The one to check over is `PublishNewWork` in 39772fe. I had to change how we're interacting with AASM, due to the fact that AASM does not let you pass a validation context into its "bang" methods. But although the new code is more verbose, I think it's more obvious what it's doing.

Issue #327, opened 2020-05-22
Create new date field for migrated creation dates


awead:
The date/time stamp that holds the information about when a work or file was first _originally_ created needs to be migrated from Scholarsphere 3.

## WDLL
* create a new datetime stamp on Work, Collection, and FileVersionMembership (?)
* migrate original create dates to those fields
* ensure the field is auto-populated with the current datetime for new resources

awead:
`deposited_at` follows the Rails convention of `created_at` and `updated_at` database column names. The display name, of course, can be anything.

PR #346, opened 2020-06-02
I#327 date deposited

awead:
Related to #327 

This PR adds a new `deposited_at` column name to Work, FileResource, and Collection models, to store the date/time of a resource's original creation. For migrated resources, this would be a time in the past, before the `created_at` value automatically added by Rails. For new resources, it will be a date/time equal to (within a few milliseconds) to the auto-generated Rails value.


awead:
Note to our future selves: this has been tested in `preview/migration` and the dates are migrating for works, file resources, and collections.

spec/support/shared/a_resource_with_a_deposited_at_timestamp.rb
```diff
@@ -0,0 +1,46 @@
...
+RSpec.shared_examples 'a resource with a deposited at timestamp' do
+  it { is_expected.to have_db_column(:deposited_at).of_type(:datetime) }
+  it { is_expected.to validate_presence_of(:deposited_at) }
+
+  describe '#deposited_at' do
+    context 'with the default value' do
+      subject(:resource) { described_class.new }
+
+      it 'is set to the current time' do
+        expect(resource.deposited_at.day).to eq(Time.zone.now.day)
```
rschenk, line 12: This works, I typically do time-based tests like this, for what it's worth

```ruby
expect(resource.deposited_at).to be_within(2.seconds).of(Time.zone.now)
```

PR #353, opened 2020-06-04
Display deposited dates instead of creation dates

### LINKED ISSUE #352: Search and display for dates
With the changes to published date and deposit dates, we need the UI to:

* show the `deposited_at` field for the date of creation

awead:
Sets the default time zone to Eastern time and ensures that the deposited date is displayed is displayed for search results and resource pages.

Fixes #352

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,4 +6,9 @@ module ApplicationHelper
   def new_session_path(*_)
     new_user_session_path
   end
+
+  def date_display(args)
+    Time.zone.parse(args[:document][args[:field]])
+      .to_formatted_s(:long)
+  end
 end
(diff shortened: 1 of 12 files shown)


app/helpers/application_helper.rb
```diff
@@ -6,4 +6,10 @@ module ApplicationHelper
   def new_session_path(*_)
     new_user_session_path
   end
+
+  def date_display(args)
+    Time.parse(args[:document][args[:field]])
+      .in_time_zone(Rails.configuration.time_zone)
```
rschenk, line 12: This works but is verbose and thankfully Rails has a method for this: `Time.zone.parse`

```ruby 
Time.parse(args[:document][args[:field]])
  .in_time_zone(Rails.configuration.time_zone)

### vs 

Time.zone.parse(args[:document][args[:field]])
```

rschenk: Another good one to know is `Time.zone.local` which you can use to construct new date-times in the local tz

PR #354, opened 2020-06-04
Allow 'unspecified' for migrated works

### LINKED ISSUE #349: Works without a resource type
There are little over 1,500 works without a resource type. This is likely because resource type was not required until later releases of Scholarsphere.

Since resource type is used to determine work type in SS4, and that field is required, we're going to need to assign some kind of work type to a work that doesn't have a designated resource type. The easiest way to do this would be to add an "unknown" option to the Work model's list of available work types, and make that the work type for migrated works that have no resource type.

IMPORTANT: we must make sure that the unknown option is NOT included in the dropdown?

awead:
For migrated works that don't have a work type, we can use 'unspecified' to allow the work to be published.

Fixes #349


app/models/work.rb
```diff
@@ -70,7 +71,7 @@ def self.display(type)
     end
 
     def self.options_for_select_box
-      all
+      (all - ['unspecified'])
```
rschenk, line 74: I know we already talked through this ad nauseam, and came to an agreement to do this, and no I'm not having second thoughts. _But_ using strings like this is the reason I do like the approach of using constants for these, even though _yes_ it's double the work. At least this usage of the string is inside the class. Just making a note of it, I don't think we need to change anything

awead: Yeah, I agree. I went back-and-forth on this a lot. I wonder if work type could extracted out into its own class, `WorkType` and managed that way. I _do_ like the enum approach with strings. Anyway, I'm open to refactoring this.

rschenk: I think it's fine as long as we're only using the string version inside the Types class/module. If they start leaking outside (which the enum should prevent), I think that's when we've got cause for concern

awead: So, I did notice that it's "leaking" into the PublishNewWork service, so I'm pushing a small change that identifies 'unspecified' as `Work::Types.unspecified` just like we do with `Work::Types.default`.

PR #355, opened 2020-06-04
match admin group to what is stored

whereismyjetpack:


config/application.rb
```diff
@@ -23,6 +23,7 @@
 
 module Scholarsphere
   class Application < Rails::Application
+    require 'healthchecks/healthchecks'
```
awead, line 26: What is 'healthchecks' ?

whereismyjetpack: eek. good catch. this is part of sidekiq monitoring that i've been working on. not sure how it snunk in here.

awead: No worries... this is what PR reviews are for! This kinda stuff happens to me all the time, especially when we're juggling lots of different branches.

PR #356, opened 2020-06-04
Preview/sidekiq healthcheck

whereismyjetpack:
Adds checks for sidekiq queue latency and DeadSet queue monitoring. Also tweaks the default max_retries:

Default sidekiq max retries is fairly high and won't drop something in the DeadSet queue until 21 days (https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry) I set this to 2 as a number, but we can tweak with an environment variable in production. 

Latency is defaulted to 30 seconds, and can also be tweaked. These checks are also "OPTIONAL" which means they won't return an unfavorable status code, but will show a "FAILED" message. this will allow us to take human action on failures, instead of automatic restarts which will be unlikely to fix the problem. 

There were some things I was unsure about stylistically, so I'm putting this as WIP for conversation on those things.

Output for these checks can be viewed here 
https://scholarsphere-sidekiq-healthcheck.dsrd.libraries.psu.edu/health/all
https://scholarsphere-sidekiq-healthcheck.dsrd.libraries.psu.edu/health/sidekiq
https://scholarsphere-sidekiq-healthcheck.dsrd.libraries.psu.edu/health/sidekiq_deadset


lib/healthchecks/healthchecks.rb
```diff
@@ -0,0 +1,38 @@
...
+      mark_failure if size.positive?
+      mark_message "There are #{size} messages in the DeadSet Queue"
+    end
+  end
+
+  class QueueLatencyCheck < OkComputer::Check
+    attr_accessor :threshold
+
+    def initialize(threshold = 30)
+      self.threshold = Integer(threshold)
```
rschenk, line 17: Ruby style tip, there are two ways to convert a string to an integer `Integer()` and `.to_i` that behave slightly differently. `nil.to_i #=> 0` while `Integer(nil) #=> raises an exception`

You very, very rarely see `Integer()` used in the wild unless you specifically want that exception raising feature. Which in this case, maybe you do? Or maybe it's an OK Computer idiom? Barring either of those, `.to_i` is much more common to see in Ruby code. Anyways, that just caught my eye

whereismyjetpack:
I reorganized the module, and I think it's much cleaner now. Thanks! 

I left it `require 'healthchecks'` for no real reason other than naming things is hard.

lib/healthchecks/queue_latency_check.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module HealthChecks
+  class QueueLatencyCheck < OkComputer::Check
+    attr_accessor :threshold
+
+    def initialize(threshold = 30)
+      self.threshold = threshold.to_i
```
awead, line 8: It's preferable to use `@threshold` instead of `self.threshold` for attr_accessor variables.

lib/healthchecks/queue_latency_check.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module HealthChecks
+  class QueueLatencyCheck < OkComputer::Check
+    attr_accessor :threshold
```
awead, line 5: You only need an attr_reader here.

lib/healthchecks/queue_latency_check.rb
```diff
@@ -0,0 +1,29 @@
...
+      @failures = nil
+      @message = Hash.new
+      queues = Sidekiq::Queue.all
+      queues.each do |q|
+        latency(q.name, threshold)
+      end
+
+      mark_failure if @failures
+      mark_message @message
+    end
```
awead, line 21: I think this can be simplified:
``` ruby 
def check
  @message = Hash.new
  Sidekiq::Queue.all.map do |queue|
    queue_latency = Sidekiq::Queue.new(queue).latency
    @message[queue] = "has a latency of #{queue_latency} seconds"
    mark_failure if queue_latency > threshold
  end
end
```
I don't believe `mark_message` is necessary because `message` has been declared an attr_accessor in the super class.

spec/lib/healthchecks/queue_latency_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      config.redis = redis_config.to_hash
+    end
+
+    Sidekiq.configure_client do |config|
+      config.redis = redis_config.to_hash
+    end
+  end
+
+  describe '#check' do
+    context 'when latency is fine' do
```
awead, line 23: What about when the latency is not "fine" ?

spec/lib/healthchecks/queue_latency_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
```
awead, line 1: The file ought to be named `queue_latency_check_spec.rb` so it maps to directly to the Ruby source file. A lot of automated test runners use that as the logic for finding the spec test for a matching file.

spec/lib/healthchecks/queue_dead_set_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
```
awead, line 1: ditto here on the spec file name: queue_dead_set_check_spec.rb

lib/healthchecks/queue_latency_check.rb
whereismyjetpack: Thanks for the tip. I've seen it both ways, The OkComputer gem uses `self` _shrug_

spec/lib/healthchecks/queue_latency_spec.rb
whereismyjetpack: I'm not sure how to mock out an item on the queue, and fake it's date. I'd be happy to add that path though.

spec/lib/healthchecks/queue_latency_spec.rb
whereismyjetpack: Thanks for the tip!

lib/healthchecks/queue_latency_check.rb
whereismyjetpack: Neat! I'm not sure why i didn't keep this as one method prior.

spec/lib/healthchecks/queue_latency_spec.rb
awead: We could look at this today during refactoring, if you'd like.

spec/rails_helper.rb
```diff
@@ -8,6 +8,10 @@ def ci_build?
   ENV.key?('CIRCLECI') || ENV.key?('CI')
 end
 
+def redis?
```
whereismyjetpack, line 11: we probably don't need this anymore.

whereismyjetpack:
I tested this locally and i think we are good to rock and roll

spec/rails_helper.rb
awead: good catch!

spec/lib/healthchecks/queue_dead_set_check_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      it 'returns no failure' do
+        hc = described_class.new
+        hc.check
+        expect(hc.failure_occurred).to be nil
+      end
+    end
+
+    context 'when there are messages in deadset' do
+      before do
+        serialized_job = Sidekiq.dump_json(jid: '123123', class: 'SomeWorker', args: [])
```
awead, line 36: @whereismyjetpack I wonder if a "nicer" way of doing this would be:
``` ruby 
class DeadSetTestJob < ApplicationJob
  queue_as :test_queue

  def perform
    raise StandardError, "I'm failing on purpose!"
  end
end
```
Would that go in the failure queue?

spec/lib/healthchecks/queue_dead_set_check_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    Sidekiq.configure_server do |config|
+      config.redis = redis_config.to_hash
+    end
+
+    Sidekiq.configure_client do |config|
+      config.redis = redis_config.to_hash
+    end
+
+    ds = Sidekiq::DeadSet.new
+    ds.clear
```
awead, line 22: I think this could be replaced with:

``` ruby
Sidekiq::Testing.disable!
Sidekiq::DeadSet.new.clear
```

spec/lib/healthchecks/queue_dead_set_check_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'sidekiq/testing'
+require 'okcomputer'
+require 'healthchecks'
+require 'scholarsphere/redis_config'
+
+RSpec.describe HealthChecks::QueueDeadSetCheck do
```
awead, line 9: do we need to add a `unless: !Scholarsphere::RedisConfig.new.valid?` ?

spec/lib/healthchecks/queue_dead_set_check_spec.rb
whereismyjetpack: That would go in the failure queue, but in sidekiq "failure" is just for statistics, we'd have to wait for retries before it got into the deadset -- which is why i decided to short circuit that and plop something on deadset.

spec/lib/healthchecks/queue_dead_set_check_spec.rb
whereismyjetpack: Yeah, i do this a lot. i think in blocks, and it ends up in my text editor in that way

spec/lib/healthchecks/queue_dead_set_check_spec.rb
whereismyjetpack: good catch, this should match the other one.

whereismyjetpack:
I'm going to fix up the remaining bits, and we can merge on Monday. this was fun. thanks!

PR #368, opened 2020-06-09
Solr indexing is async

### LINKED ISSUE #248: Make indexing async
Currently, indexing works and their versions into Solr happens synchronously when the resource is created or updated in the database. We could make this process async, so that the indexing would occur immediately after the record is saved in the database.

## Advantages
* greater independence: Solr wouldn't prevent records from being created
* if indexing is Solr it slow, the user won't notice

## Disadvantages
* slightly more complex
* the record for searching might lag behind the database record

rschenk:
Closes #248


whereismyjetpack:
We presently don't have a `default` queue in our sidekiq. in `config/sidekiq.yml` we should specify `default` as a queue. 

Which makes me wonder, should we call it something like `solr` instead? 🤷

rschenk:
Good thinking @whereismyjetpack. Why don't you give this a "request changes" review so Adam doesn't accidentally merge it

rschenk:
@whereismyjetpack I added that as a separate commit but we can squarsh when we merge

awead:
@rschenk I approved the request. Although, I wonder if it's possible to add the tests for `after_save` and `update_index_async` to the shared spec. They're really similar, but if it's too much trouble, don't bother.

PR #372, opened 2020-06-18
Run MetadataListener Jobs

awead:
This ties into https://github.com/psu-stewardship/metadata-listener/pull/28

Scholarsphere submits a job via ActiveJob to the `metadata` queue on Redis when a file is promoted to S3 storage in Shrine. This MetdataListener application is listening to that queue and performs additional work on the file, downloading it from S3 and checking for viruses. The result of the virus check is submitted by the MetadataListener application back to Scholarsphere via the API.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,3 +56,7 @@ RSpec/AnyInstance:
 Style/ClassVars:
   Exclude:
     - 'spec/system/active_job_system_spec.rb'
+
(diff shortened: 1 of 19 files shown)


whereismyjetpack:
@awead mind if i deploy this as scholarsphere-jobs?

rschenk:
Looks good to me, I second your idea to rename the endpoint `/files`

PR #373, opened 2020-06-18
Set defaults for url host and protocol

awead:
Instead of requiring the host, we can have defaults for local development and set the protocol as well to construct the full url when needed.


awead:
Closed in #372

PR #377, opened 2020-06-19
Do not persist failed oauth to session

### LINKED ISSUE #279: Gracefully handle integration failures with Psu-authproxy
When logging in, a user's profile is obtained from the psu-authproxy application. If that application cannot obtain all the needed information, Scholarpshere should be able to continue to work. There should be some messaging to the user indicating that there's an issue.

See https://github.com/psu-stewardship/psu-authproxy/issues/28

rschenk:
This was causing a cookie overflow during retro.

Closes #279


awead:
@rschenk what do we loose by not persisting the oauth failure to the session? I'm guessing nothing?

rschenk:
@awead I think nothing. That bit was some code I cargo-culted from the OmniAuth docs, and I can't tell what it does... I figure if the auth is invalid, there's no sense in persisting it...

PR #383, opened 2020-06-22
Preview/matomo

whereismyjetpack:
Using `MATOMO_` ENV vars, adds the tracking code to the header. After numerous attempts to get this working with webpack, we went back to the original standard approach. It's probable that webpack integration isn't supported.

### DIFF
--- a/app/views/layouts/blacklight/base.html.erb
+++ b/app/views/layouts/blacklight/base.html.erb
@@ -1,22 +1,6 @@
 <!DOCTYPE html>
 <html lang="en" class="no-js">
-  <head>
-    <meta charset="utf-8">
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-
-    <!-- Internet Explorer use the highest version available -->
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-
-    <title><%= render_page_title %></title>
-    <%= opensearch_description_tag application_name, opensearch_catalog_url(format: 'xml') %>
-    <%= favicon_link_tag %>
-    <%# Overrides blacklight's base layout to use the webpacker pack tag instead of asset pipeline %>
-    <%= javascript_pack_tag 'application' %>
-    <%= stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
-    <%= csrf_meta_tags %>
-    <%= content_for(:head) %>
-  </head>
+  <%= render partial: 'layouts/head' %>
   <body class="<%= render_body_class %>">
     <div id="skip-link">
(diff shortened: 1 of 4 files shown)


app/views/layouts/blacklight/base.html.erb
```diff
@@ -16,6 +16,27 @@
     <%= stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= csrf_meta_tags %>
     <%= content_for(:head) %>
+
+<% if ENV['MATOMO_SITE_ID'] and ENV['MATOMO_URL'] %>
```
awead, line 20: No need for the if block. We can use defaults.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -16,6 +16,27 @@
...
+    <script type="text/javascript">
+    var _paq = window._paq || [];
+    /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+    _paq.push(['trackPageView']);
+    _paq.push(['enableLinkTracking']);
+    (function () {
+        var u = "<%= ENV['MATOMO_URL'] %>";
```
awead, line 27: Default to `https://analytics.libraries.psu.edu/matomo/`

app/views/layouts/blacklight/base.html.erb
```diff
@@ -16,6 +16,27 @@
...
+        _paq.push(['setTrackerUrl', u + 'matomo.php']);
+        _paq.push(['setSiteId', <%= ENV['MATOMO_SITE_ID'] %>]);
```
awead, line 29: Default to 16. See https://analytics.libraries.psu.edu/matomo/index.php?module=CoreHome&action=index&date=yesterday&period=day&idSite=16

I created a "Scholarpshere Dev" site in matomo.

awead:
@whereismyjetpack Ok, ended up completely changing this up. It'd be worth seeing that when make changes to ENV['MATOMO_SITE_ID'] in the Docker container, webpack picks it up.

awead:
@whereismyjetpack take 3? This is better, I think. I'd like to get @rschenk to weigh in. Common header code is extracted out in the first commit, and then we install Matomo in the second. Configs are managed via data attributes.

whereismyjetpack:
it appears that this will alleviate any concerns. ✅

rschenk:
This looks good to me. I like having the head section extracted to a partial, that alone may make it easier to integrate Christie's designs

awead:
@whereismyjetpack aaaaaaand we're back where we started! It appears that the tracking code won't play well with webpack. I started a thread here about it https://forum.matomo.org/t/tracking-code-with-wepack/37732. CC: @cdmo

cdmo:
Well I tried unsuccessfully at this a year or so ago, but I was hopeful for your attempts here! Googling around it does seem that adding the scripts with regular old html (_cringe_) still seems to be the way folks are adding stuff like this (I searched "google analytics webpacker" because GA works the same way). Oh well.

Issue #384, opened 2020-06-22
SS3 migration: concatenate multiple description fields into a single value


rschenk:
In SS4 we've made WorkVersion#description into a single-valued field. In order to migrate, we'll need to concatenate any multiple descriptions that are in SS3 into a single value, much like we did for [publication dates](https://github.com/psu-stewardship/scholarsphere/blob/migration/lib/scholarsphere/migration/work.rb#L24)

awead:
Fixed in https://github.com/psu-stewardship/scholarsphere/commit/20c7fd0a6872e2bf50917e49aec77d57f8991e5b

PR #385, opened 2020-06-22
Description is single-valued and required to publish

### LINKED ISSUE #351: Make work description required
There is no min/max word count on description but we should make that field mandatory.

rschenk:
Closes #351

Will depend on #384 to be implemented before migration succeeds


rschenk:
@awead I totally forgot about collections! Good catch

awead:
@rschenk I do it all the time...

PR #390, opened 2020-06-24
Track views and downloads with ViewStatistic model

awead:
Adds a view statistic model with relationships to work versions, collections, and file resources, in order to track views and downloads of resources.

Requests from known bots are not counted using the Browser gem, and only views for work versions are counted. Viewing a work increments the view count for the latest published version only.

### DIFF
--- a/app/models/concerns/view_statistics.rb
+++ b/app/models/concerns/view_statistics.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ViewStatistics
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :view_statistics,
+             as: :resource,
+             dependent: :destroy
+  end
+
+  def count_view!
+    view_statistics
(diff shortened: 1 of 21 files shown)


app/models/concerns/view_statistics.rb
```diff
@@ -0,0 +1,16 @@
...
+  included do
+    has_many :view_statistics,
+             as: :resource,
+             dependent: :destroy
+  end
+
+  def count_view!
+    today = view_statistics.find_or_initialize_by(date: Time.zone.today)
+    today.update(count: today.count + 1)
+  end
```
rschenk, line 15: Fun fact, ActiveRecord actually has a method for this (because, of course it does). So if you wanted to, you could write this method like this. 

```ruby
  def count_view!
    view_statistics
      .find_or_initialize_by(date: Time.zone.today)
      .increment(:count)
      .save
  end
```

PR #393, opened 2020-06-26
Show view statistics on resource page

rschenk:
Show a graph of views through time on resource pages. 

WorkVersions and Collections show the exact number of views for that particular resource.

Works show an aggregate of all _published_ versions.

File downloads are not shown at this time.

### DIFF
--- a/app/components/collection_metadata_component.rb
+++ b/app/components/collection_metadata_component.rb
@@ -16,7 +16,6 @@ class CollectionMetadataComponent < ActionView::Component::Base
     :title,
     :subtitle,
     :creator_aliases,
-    :description,
     :keyword,
     :contributor,
     :publisher,
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+class AnalyticsController < ApplicationController
+  def show
+    resource = FindResource.call(params[:resource_id])
+    view_statistics = load_view_statistics(resource)
+
+    respond_to do |format|
+      format.html { head :unsupported_media_type }
+      format.json { render json: view_statistics }
+    end
+  end
+
+  private
+
+    def load_view_statistics(resource)
+      resource
+        .stats
+        .map { |date, count, total| [date.iso8601, count, total] }
+    end
+end
--- a/app/services/load_view_statistics.rb
+++ b/app/services/load_view_statistics.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# @abstract Loads the view statistics for a given model. It returns these
+# statistics as a 2d array of raw values, rather than ActiveRecords, for speed.
+# The returned rows are ordered by date ascending, and have the structure:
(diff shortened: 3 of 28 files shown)


awead:
@rschenk Looks good to me. Bypassing the ActiveRecord stuff really improves performance here, right? I'm assuming the main use case for this is powering the analytics graphs? I was wondering if there'd ever be the need to query a date range in the array of results and if another data structure might help with that. I can't think of one...

rschenk:
@awead Yep that's exactly its intention is to power the graphs. Instantiating an ActiveRecord for each row would be expensive and pointless, since the graphs want to be fed a 2d-array _anyways,_ so why not cut out the middle man. 

This class can totally have date range filtering, and batched/page loading added to it later if need be

app/components/collection_metadata_component.rb
```diff
@@ -16,7 +16,6 @@ class CollectionMetadataComponent < ActionView::Component::Base
     :title,
     :subtitle,
     :creator_aliases,
-    :description,
```
awead, line 19: @rschenk why did description get removed?

app/controllers/analytics_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    respond_to do |format|
+      format.html { head :unsupported_media_type }
+      format.json { render json: view_statistics }
+    end
+  end
+
+  private
+
+    def load_view_statistics(resource)
+      stats = if resource.is_a?(Work)
```
awead, line 17: @rschenk what if this was a method on the model? Work#stats and WorkVersion#stats

app/services/load_view_statistics.rb
```diff
@@ -0,0 +1,95 @@
...
+#   [Date, number of views that day, running total of all views through time]
+# @example
+#   > LoadViewStatistics::Aggregate.call(models: [model_1, model_2])
+#   => [
+#        [Date(2020-06-01), 1,     1],
+#        [Date(2020-06-02), 100, 101],
+#        [Date(2020-06-03), 3,   104]
+#      ]
+class LoadViewStatistics
+  class Aggregate
```
awead, line 63: @rschenk This makes more sense to me as a separate file and named service, like `AggregateViewStatistics`. Alternatively, we can keep the class in the same file, but the Aggregate class ought to be inside the parent:

``` ruby
class LoadViewStatistics
  class Aggregate
  end
end
```

That way it looks more deliberate and not two different classes in the same file.

app/components/collection_metadata_component.rb
rschenk: @awead Description is no longer shown in the metadata listing in the new UI design, it's shown in its own paragraph in the main content area of the view

app/controllers/analytics_controller.rb
rschenk: I thought about that. Maybe I could put it inside the ViewStatistics concern module, then override it on Work

awead: meh. I'd say leave it for now. up to you...

PR #400, opened 2020-07-01
Use Redis to track unique visits to pages

awead:
Using Rails' feature that lets Redis serve as a cache storage layer, we can store key value pairs that correspond to unique visits to a page. By creating a key using a user's session id and the UUID of the resource they are visiting, we store that key value for a period of time. As long as the key is present, no additional page views will be counted.

There are a couple of gotchas to this method:
1. We need to expire the key on a periodic basis. Fortunately, this is very easy to do in Redis; however...
2. Using Redis as a cache layer has a couple of caveats which are outlined in https://guides.rubyonrails.org/caching_with_rails.html#activesupport-cache-rediscachestore

Notably, the recommendation is to use a _separate_ Redis service solely for caching:

> Deployment note: Redis doesn't expire keys by default, so take care to use a dedicated Redis cache server. Don't fill up your persistent-Redis server with volatile cache data! 

But, we not really using Redis as a persistent store, since job data isn't persisted. So we may have some wiggle room. I'll leave it to @whereismyjetpack to weigh in.

It is also worth noting that using Redis as a cache store alone, without any of the page view verification, is useful as well since it performs better than Rails' standard memcache method.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -4,12 +4,18 @@ class ResourcesController < ApplicationController
   def show
     @resource = ResourceDecorator.new(find_resource(params[:id]))
     authorize @resource
-    @resource.count_view! unless browser.bot?
+    @resource.count_view! if count_view?
   end
 
   private
 
     def find_resource(uuid)
       FindResource.call(uuid)
+    end
+
+    def count_view?
+      return false if browser.bot?
+
+      SessionViewStatsCache.call(session: session, resource: @resource)
     end
 end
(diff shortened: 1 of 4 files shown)


cdmo:
You can take a look at what we've done for Redis at https://github.com/psu-libraries/myaccount/pull/194 if you like. The overview is we use Redis for:

*Caching*

* Configured a connection in application.rb, this is a one-line Rails thing

*Temp key value store*

* Set an initializer to create a connection to a redis instance

*Sidekiq*

* Another initializer with config to establish a connection to a redis instance.

That's 3 separate Redis instances. We've got a heroku instance setup with 3 free tier redis instances that was turnkey. We use the [config gem](https://github.com/rubyconfig/config) to store connection details to these instances, overrideable via env vars.

Seems to be working for us and I'd certainly welcome feedback!

PR #402, opened 2020-07-06
Redis configuration is now a core service

awead:
Refactors the Redis configuration to be a core service. We no longer check to see if the service is valid or active, it is assumed to be, just like our database services. The configuration is persisted as a hash within Rails' application configuration structure and can be referenced from anywhere within application.

ActiveJob configuration is now kept separate from Redis and Sidekiq tests are tied to this configuration setting instead.

Note: Includes an additional commit that changes how the dates of view statistics are set. This bug was discovered when running the test suite after 10 PM local time.

### DIFF
--- a/app/models/view_statistic.rb
+++ b/app/models/view_statistic.rb
@@ -2,4 +2,12 @@
 
 class ViewStatistic < ApplicationRecord
   belongs_to :resource, polymorphic: true
+
+  after_initialize :set_defaults
+
+  private
+
+    def set_defaults
+      self.date ||= Time.zone.now
+    end
 end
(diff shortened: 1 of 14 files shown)


rschenk:
I just wanted to throw it out there that I've found it really handy to have "background" jobs run synchronously in testing situations. I made [this test helper](https://github.com/psu-stewardship/scholarsphere-4/blob/master/spec/support/active_job_helpers.rb) to force this to happen. I am _preeeetttty_ sure it will play nice with this new configuration (presumably, if the tests are passing that use it) but wanted to toss it out there as a consideration.

The mire of forcing activejob/sidekiq to perform jobs synchronously is a real brain bender. I dug through so much (outdated) documentation to make that dumb helper.

awead:
@rschenk ah, that's probably why if I specify `:sidekiq` in test, it just works. Yes, I agree. I actually prefer `:inline` or `:async` for dev and test because you see the failures immediately. I made the above change because I thought it was related to the ci failures. It wasn't, so I will revert everything back.

PR #407, opened 2020-07-08
Update to ViewComponent 2.x and Blacklight 7.10

### LINKED ISSUE #403: Upgrade to Blacklight 7.10 and ViewComponent 2.x
The latest version of Blacklight, which is 7.10 at this time, uses the ViewComponent gem. We use the gem as well but in a previous iteration that was built when it was thought the gem would be incorporated in Rails. As it turns out, that's not going to happen  and the gem has since continued on a new course of development.

Because we're using the older form of the gem, we can't upgrade to Blacklight 7.10 without first changing our view components to match with the new 2.0-line of design.

## WDLL
* upgrade to ViewComponent 2-latest
* change our existing view component classes to match the new implementation
* upgrade to Blacklight 7-latest

awead:
The newest version of Blacklight uses the new ViewComponent gem, however, we were using an older, different version of it that is incompatible with the new Blacklight version.

Previous versions of ViewComponent were under ActionView because at the time the plan was that the gem would be added to Rails. That is now not the case, and it will remain a separate gem outside of the Rails and ActionView namespace.

This upgrades us to the new gem and changes our component classes accordingly. The only major difference is that ViewComponent does not support validation. But, since we didn't seem to be making much use of
that feature, removing it from our components doesn't affect the outcomes of the tests.

Fixes #403

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,11 +33,8 @@ gem 'scholarsphere-client', github: 'psu-stewardship/scholarsphere-client', bran
 gem 'shrine', '~> 3.0'
 gem 'sidekiq', '~> 6.0'
 gem 'uppy-s3_multipart', '~> 0.3'
+gem 'view_component'
 gem 'webpacker', '~> 4.0'
-
-# Experimental
-# @todo upgrade to latest version that alters the implementation pattern
-gem 'actionview-component', '= 1.17.0'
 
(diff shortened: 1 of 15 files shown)


awead:
@rschenk I ticketed the questions ticketed in #409 and have merged this to move on.

Issue #411, opened 2020-07-10
Azure Groups Discrepancy


whereismyjetpack:
The `.owner` and `.admin` groups within ldap are _not_ represented in azure token, or graph api groups. as well as some other groups. 

Currently we have edit acls on the following groups that are _not_ coming out of the graph api
```
umg-personal.cam156.access_to_my_docs
umg-psu.nursing.simulations.admin
umg-up.emsl.staff.admin
umg-up.emsl.staff.owner
```

Currently we have read acls on the following groups that are _not_ coming out of the graph api
```
umg-psu.nursing.simulations
umg-psu.nursing.simulations.admin
umg-up.lps.augmented_learning_research_group
umg-up.lps.augmentedlearnin
```

If we launch with Azure AD, and leave the acls as such, members of those groups will stop maching, and stop having access (presumably owners will retain access) @awead ?

It seems that groups that aren't synced into o365, .admin, .owner, and -personal groups are not part of the _new_ group system. How should we migrate these ACLs? I don't see an automated way of dealing with these discrepancies. I see 3 options:

1.) migrate them as is, and let the users discover they no longer have  access. 
2.) contact the users ahead of time, and have them create new standard, synced (non .admin .owner) groups for this use case and we can make that mapping after migration
3.) we don't launch with Azure AD, and presumably run into this issue when we are forced to :)

whereismyjetpack:
The other difference that I will document here, just so we don't forget is that the shape of groups is slightly different. For migration we'll have to transform the groups in the existing database

i.e
```
"umg/foo".gsub('/','-')
```

awead:
The issues with read acls will be rendered moot by the fact that everything is viewable and downloadable by Penn State users. The edit acls can be dealt with on a case-by-case basis, but I think the basic plan is migrate everything as-is and then edit the group names post-migration to reflect what's coming out of Azure. We really only have two groups to deal with: umg-psu.nursing.simulations and umg-up.emsl.staff. The first one looks like something Carolyn Cole (cam156) had done probably for testing. I don't think we need the "admin" vs. "owner" distinction. We just need a group to identify as the editors for these resources. I think that's something @srerickson could ask them about.

awead:
Here's an example from the "emsl" group https://scholarsphere.psu.edu/concern/generic_works/n02870z134
and ones from nursing appear to be private https://scholarsphere.psu.edu/concern/generic_works/5t34sj73g so those will migrated in a "draft" status.

[...]

PR #413, opened 2020-07-15
AzureAD

whereismyjetpack:
Adds azure_oauth authentication strategy that cribs heavily off of our old oauth strategy. Before, we weren't explicitly setting `OAUTH_TOKEN_URL` which used the default `/oauth/token` -- in Azure this is a custom address for each tennant. In Azure groups come across in the token, or through the Graph API will return more than 150 items, where the token will be limited.

There are some outstanding questions regarding groups #411  and the groups come across a bit different. basically `"umg/foo".gsub('/','-')` 

For Local development I've shared with @awead I'll get @rschenk his when he returns


awead:
Overall, this looks very good. `AzureOauth` just needs time TLC and testing, then I think this should be good to go!

whereismyjetpack:
I moved a lot of the logic into the user model, as the strategy should probably just emit what it sees. 

I'm at a bit of a loss for how to test the `graph_groups` method. I'm not 100% sure we even need it but I anticipate a problem with folks that have many groups, and that's the way to do it. 

I tried to make this as idiomatic as i could. I'm unsure about minor things like `delete_if?` vs `reject`

I left the LdapCleaner code in, but i presume we don't need it with the way info comes out of the azure API 

I'm looking for any guidance on what to focus on next. Thanks!

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -8,20 +8,21 @@
...
       .with(oauth_response)
       .and_return(user)
   end
 
-  describe '#psu' do
+  describe '#azure_oauth' do
     context 'when the user persists correctly' do
       before do
-        get :psu
+        allow(user).to receive(:persisted?).and_return(true)
```
awead, line 24: This seems a little suspect. If we're trying to test if a user persists correctly, this could be bypassing that.

whereismyjetpack:
I fixed up the one concern. I might need to pair to figure out how to test the graph_groups. Or, i might just yank it out to get unstuck.

PR #414, opened 2020-07-17
Create a new homepage with our custom UI

awead:
Incorporates our customized UI layout styles with a new frontpage using a controller to render the home page as well as other future customized pages.

The original scholarsphere webpack is kept intact so existing pages can continue to render. Eventually, other pages/controllers will be migrated over to the new UI layout.

## Remaining Todos

- [x] fix the blue secondary navbar
- [x] add a place for flash messages
- [ ] decide how to mark works as featured (see #417)
- [x] add feature tests
- [x] decide what to do about the sass linting errors

## Differences From Wireframes

The navbar was simplified a bit, choosing one method of rendering dropdowns and only having them in the secondary blue nav. This allows us to put links to other fixed pages in the upper right-hand corner.


awead:
Here's a preview
Screen Shot 2020-07-17 at 3 31 31 PM

PR #433, opened 2020-07-27
it appears rails comes with sprockets. this fixes  CVE-2014-7819

whereismyjetpack:


awead:
@whereismyjetpack I took out sprokets in 4e5c43388ccbd54917d68575ca57b30b6125525c. Is it still lurking around somewhere?

whereismyjetpack:
Looks like rails depends on `sprockets-rails` but at least now a patched version is resolvable

PR #447, opened 2020-08-04
Display icons for work types

### LINKED ISSUE #386: Work Type Icons
We need icons for work types, to be used when listing works/collections (browse, dashboard, etc.). We don't need an icon for each work type, just a few general one. I think the following would be sufficient:

- Journal Article / Conference Proceeding
- Book / Book Chapter
- Audio / Video
- Image
- Map
- Poster / Presentation
- Data Set
- Software / Source Code

awead:
Selects a icon from the materials.io icon set based on the work type. The icons are displayed for search results and dashboard results.

Works and work versions are indexed differently, and include the work type identifying key as well as the display variation of the type. Additionally, our icons are compiled into our webpack instead of being loaded separately.

Fixes #386 

## Search and Dashboard Results

Screen Shot 2020-08-04 at 2 06 40 PM

## Featured Works

Screen Shot 2020-08-04 at 2 07 37 PM

### DIFF
--- a/app/components/thumbnail_component.rb
+++ b/app/components/thumbnail_component.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+class ThumbnailComponent < ApplicationComponent
+  attr_reader :resource, :featured
+
+  # @param [SolrDocument, WorkVersion] resource
+  def initialize(resource:, featured: false)
+    @resource = resource
+    @featured = featured
+  end
+
+  def icon
+    icon_map.fetch(resource.work_type, 'bar_chart')
+  end
+
+  def html_classes
+    if featured?
+      'thumbnail col-xxl-6 ft-work__img'
+    else
+      'thumbnail'
+    end
+  end
+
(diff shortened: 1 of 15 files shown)


app/components/thumbnail_component.rb
```diff
@@ -0,0 +1,57 @@
...
+      'thumbnail col-xxl-6 ft-work__img'
+    else
+      'thumbnail'
+    end
+  end
+
+  private
+
+    def featured?
+      current_page?('/')
```
rschenk, line 26: I'm not crazy about linking a dependency here between this component and the globally available helpers in Rails. For example, we might want featured displays somewhere else besides the homepage eventually (like maybe on a person's dashboard we might want to show "your most popular works" with a featured style).

how about we pass `featured` into the constructor as a keyword option that defaults to false?

Also as part of my navbar PR, I added a method, `ApplicationHelper::home?` that can do the `current_page?('/')` bit.

PR #451, opened 2020-08-05
Status badges with and without a version number

rschenk:
This also includes a cool feature of View Components, that in dev mode gives you the ability to create a simple controller/view to demo the components.

Check out `test/components/previews/status_badge_component_preview.rb` and `test/components/previews/status_badge_component_preview/default.html.erb`

Which in development mode can be accessed at the route (both puma-dev and bin/rails s):
http://scholarsphere-4.test/rails/view_components/status_badge_component/default
http://localhost:3000/rails/view_components/status_badge_component/default


rschenk:
@awead 

We will need the inverted version on the resource page:
Screen Shot 2020-08-06 at 12 16 47 PM

Good call on moving the layout, I can do that I reckon

PR #455, opened 2020-08-06
in development we need to check that yarn is up to to date.

whereismyjetpack:


bin/sidekiq
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 if [ ${RAILS_ENV:-development} != "production" ]; then
   bundle check || bundle
+  yarn
```
awead, line 4: I'm not sure what we're trying to check, but you may want `yarn install --check-files` instead.

PR #461, opened 2020-08-07
Refactor decorators, plus a ResourceDecorator factory

rschenk:
I refactored the WorkVerisonDecorator and WorkDecorator out of the Dashboard namespace.

While I was at it, I also:
* I also changed the names of a couple methods in WorkVersionDecorator to match what the new design will need
* Made a factory method on ResourceDecorator that will return the given resource with the decorator of the correct subclass. I named this method `.decorate` instead of `.build` to avoid colliding with `ActiveRecord::Base.build`

With this change, this _should_ be the last arduous refactor that we need to implement the resource pages. Yay.


app/components/work_histories/work_history_component.rb
```diff
@@ -37,7 +37,7 @@ def load_changes
         work
           .versions
           .map do |work_version|
-            decorated_work_version = Dashboard::WorkVersionDecorator.new(work_version)
+            decorated_work_version = WorkVersionDecorator.new(work_version)
```
awead, line 40: @rschenk Do you want to switch to the new `ResourceDecorator.decorate` pattern here?

app/controllers/dashboard/work_histories_controller.rb
```diff
@@ -4,7 +4,7 @@ module Dashboard
   class WorkHistoriesController < BaseController
     def show
       @work = policy_scope(Work).find(params[:work_id])
-      @latest_work_version = Dashboard::WorkVersionDecorator.new(@work.latest_version)
+      @latest_work_version = WorkVersionDecorator.new(@work.latest_version)
```
awead, line 7: @rschenk Same question here, too.

PR #462, opened 2020-08-07
Monthly stats email

### LINKED ISSUE #380: Monthly Report
Make sure our monthly report is being run to let users know their download counts.

awead:
Sends an email to each actor who has deposited works into Scholarpshere, informing them of how many files they have and the collective number of downloads for those files.

Screen Shot 2020-08-07 at 4 35 02 PM

A rake task sends the email to users unless they have opted out of the process or are no longer active. Additional boolean attributes on the User model define these for us. It is assumed that a later process will mark inactive users, and than additional features added to the users' profile page will allow them to elect to NOT receive future emails.

For migration, we can migrate any opt outs to the new user records.

Fixes #380

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -20,6 +20,7 @@ linters:
       - '*app/views/dashboard/catalog/_index_work_version.html.erb'
       - '*test/*'
       - '*app/views/layouts/component_preview.html.erb'
+      - '*app/views/actor_mailer/*'
   RightTrim:
     enabled: true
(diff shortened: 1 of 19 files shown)


rschenk:
@awead optimizing that query was not actually as hard as it looked on the surface. 

The actor's depositor_id lives in the works table, and a foreign key to the file_resource_id lives in the file version membership. So all we had to do was inner join from file_version_memberships, through work_versions, to works.

PR #470, opened 2020-08-14
[WIP] Adds Scheduled task to run the Monthly Email job

whereismyjetpack:


whereismyjetpack:
I need to wire in tests, but wanted to put this out for review / help

app/jobs/mailers/monthly_stats.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Mailers
+  class MonthlyStats < ApplicationJob
+    queue_as :mailers
+
+    def perform
+      User.where(provider: 'azure_oauth', opt_out_stats_email: false, active: true).each do |user|
```
rschenk, line 8: Generally speaking, always use `.find_each` instead of `.each` when dealing with ActiveRecord. What that does is secretly  under the hood load records out in batches, rather than attempting to load out the entire result set at once (which can lead to memory swapping).

app/jobs/mailers/monthly_stats.rb
```diff
@@ -0,0 +1,13 @@
...
+        ActorMailer.with(actor: user.actor).monthly_stats.deliver_later
```
rschenk, line 9: When you call `user.actor` here, inside the loop, it is hitting the database for the associated actor once per iteration of the loop. This is what's commonly called a "n+1 query bug" because it hits the database once to load out all the Users, then _n_ times for every iteration to load the Actor. The way to avoid this is in Rails-speak is to change the line above from `User.where(...)` to `User.includes(:actor).where(...)`. This tells rails to automatically inner-join the actors table and load everything out at once, so you only have a single database hit.

whereismyjetpack:
I think the main benefit of sidekiq-scheduler is that it uses redis as a semaphore and doesn't rely on a node / thread to be healthy in order to meet schedules. When I looked at rufus-scheduler directly, the only locking mechanism is with a simple file lock.

awead:
I'm not saying that it doesn't have benefits. It would make sense if there was an existing job that were trying to schedule. If there was a way to schedule ActorMailer directly, that would be ideal. One of the things that is problematic with SS3 is we have a lot of job-calling-job type architecture and it's very confusing to follow. I want to avoid the situation where every cron job becomes a job plus another job situation.

At the very least, it shouldn't be called a mailers job because it's not the queue that's doing the mailing. If the mail failed to send, it would be on the queue that ActorMailer is using.

Maybe we should step back and again and review why we're not going to use cron. In the rufus-scheduler (RS) README is states: 'rufus-scheduler is not a cron replacement'. Sidekiq-scheduler does have benefits, yes, but RS is a dependency. I would also note that the RS build is failing and hasn't had any updates since Dec. 2019.

whereismyjetpack:
you typically don't run cron in a container, and you have the same locking problem with system cron, where you have to pin it to a host, and that host has to be available in order to run it. 

There is also the kubernetes cronjob method, which i'd be ok with. i just need to know the rake task and the cron schedule to use. closing in favor of that.

Issue #416, opened 2020-07-20
Accessibility Fixes


binkylush:
## WDLL

- [x] Uniform Headings for all pages #474
- [x] Address label and contrast issues in work edit form #475
- [x] Add label to search form #476
- [x] Label for file downloads #477
- [x] Form labels for the access permissions modal #478
- [x] Use definition lists for metadata (fixed in #473)
- [x] Readable ordering for sidebar layouts #480

### Details Tab

![Addnewwork-workdetails](https://user-images.githubusercontent.com/30327057/87953046-8b4a2580-ca78-11ea-8df7-843403694f93.png)

### Publish Tab
![addnewwork-publish](https://user-images.githubusercontent.com/30327057/87953161-b2085c00-ca78-11ea-9e3d-84c12293113f.png)

## Dashboard List/Grid:
![dashboardgrid](https://user-images.githubusercontent.com/30327057/87953027-86857180-ca78-11ea-82d6-ca3a742999f7.png)
![Dashboardlist](https://user-images.githubusercontent.com/30327057/87953112-a2891300-ca78-11ea-84bf-8e1a2b246295.png)

### View work page 
![Viewwork](https://user-images.githubusercontent.com/30327057/87953156-b03e9880-ca78-11ea-9ac3-7465829be156.png)
### View work modal:
![Viewwork-modal](https://user-images.githubusercontent.com/30327057/87953138-aae14e00-ca78-11ea-92a1-43a13e03b2b1.png)

awead:
## Changes to the Landing Page
* consistent use of header tags H1 through H3
* use definition list for metadata
* use ordered list for featured works
![Screen Shot 2020-08-12 at 1 52 53 PM](https://user-images.githubusercontent.com/312085/90051808-40ea4c00-dca6-11ea-9f42-91780461e01c.png)

awead:
## Changes to View Work Page
* consistent heading tags H1-H3

![resource-display](https://user-images.githubusercontent.com/312085/90056762-6b8bd300-dcad-11ea-94b1-b9e326281fe6.png)

awead:
## Changes to Search Results
* add H3 tags to titles (H1 and H2 tags can be added later using existing styles or made to be "screen reader only")

![search-results](https://user-images.githubusercontent.com/312085/90056968-b9084000-dcad-11ea-8dfa-d167d28475dd.png)

PR #473, opened 2020-08-17
Public resource view styling

### LINKED ISSUE #431: Update resource display page with new UI elements
This should apply https://github.com/brycekirk/scholarsphere/blob/master/html/view-work.html to our public resource display pages, i.e. `/resources/[uuid]` as well as dashboard show pages, i.e. `dashboard/work_versions/[id]`.

## Notes

* navbar components may look slightly to match the existing display
* while this page will also render collections, it's ticketed separately in #284

rschenk:
Big whopper commit that also notably includes:
- Decorator refactors
- Global header ordering for accessibility
- Some metadata lists being marked up as DLs 

Closes #431

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -18,6 +18,7 @@ linters:
       - '*app/components/flash_message_component.html.erb'
       - '*app/views/catalog/_search_form.html.erb'
       - '*app/views/dashboard/catalog/_index_work_version.html.erb'
+      - '*app/views/shared/_work_version_files.html.erb'
       - '*test/*'
       - '*app/views/layouts/component_preview.html.erb'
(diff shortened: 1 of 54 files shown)


rschenk:
@awead Code Climate is having a squirrel gripper because I duplicated the analytics chart javascript from `app/javascript/scholarsphere/view_statistics_chart.js` to `app/javascript/frontend/scripts/view_statistics_chart.js`. We talked about this, how we were just going to bring everything over into the frontend folder, then delete scholarsphere when appropriate, so we'll just need to update the codeclimate config (I don't know how to do that).

awead:
Also fixes #474 and address a number of other issues in #416.

Issue #467, opened 2020-08-14
ScholarSphere capitalization in lower left footer


DanCoughlin:
Scholarsphere should be ScholarSphere

![Screen Shot 2020-08-14 at 9 52 13 AM](https://user-images.githubusercontent.com/398837/90256533-e242e080-de13-11ea-88d8-5ae8722afb3f.png)

PR #481, opened 2020-08-18
Accessbility Fixes

awead:
Accessibility fixes identified in #416.


config/locales/en.yml
```diff
@@ -205,6 +205,7 @@ en:
     search:
       heading: 'Browse and search for works'
       placeholder: 'Search for titles, keywords, resource types, creators, etc.'
+      label: 'Search Scholarsphere'
```
rschenk, line 208: Shouldn't this be inter-capped a la #467

awead: @rschenk No, because it's a screen-reader only thing. BUT, if it ever wasn't I supposed we might want to...

awead: @rschenk meh. I changed it anyway. I don't expect it'll affect the screen reader at all.

PR #483, opened 2020-08-19
creates script to run monthly stats email task

whereismyjetpack:
Kubernetes CronJob runs `/app/bin/monthly_stats_mail`, on a configurable schedule.

### DIFF
--- a/bin/envconsul-wrapper
+++ b/bin/envconsul-wrapper
@@ -1,0 +1,36 @@
+#!/bin/bash
+# Wraps a command in envconsul, if vault token is present
+
+# Check for vault token from an init container.
+if [ -f /vault/.vault-token ]; then
+    vault_token=$(cat /vault/.vault-token)
+    export VAULT_TOKEN=$vault_token
+fi
+
+function start() {
+    if [ "$1" != "" ]; then
+        cmd="$*"
+        if [ -n "${VAULT_TOKEN}" ]; then
+            echo "have token. wrapping ${cmd} with envconsul"
+            start_envconsul
+        else
+            echo "Running $cmd"
+            ${cmd}
+        fi
+    else
+        echo "Missing Command Argument"
+        exit 1
+    fi
+}
+
+
+function start_envconsul() {
+    set -u
+    envconsul \
+        -vault-addr="${VAULT_ADDR}" \
+        -secret="${VAULT_PATH}" \
+        -no-prefix=true \
+        -vault-renew-token=true \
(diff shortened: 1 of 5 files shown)


bin/envconsul-wrapper
```diff
@@ -0,0 +1,45 @@
...
+    export VAULT_TOKEN=$vault_token
+fi
+
+function start() {
+    if [ "$1" != "" ]; then
+        cmd="$*"
+        if [ -n "${VAULT_TOKEN}" ]; then
+            echo "have token. starting envconsul"
+            start_envconsul
+        else
```
awead, line 22: This implies that the command can still run if there is no vault token present. Is this a valid scenario?

config/sidekiq.yml
```diff
@@ -2,4 +2,5 @@
 :max_retries: <%= ENV.fetch("SIDEKIQ_MAX_RETRIES", 2) %>
 :queues:
   - shrine
+  - mailers
```
awead, line 5: Do we need this?

whereismyjetpack:
Totally agree. I think we can have the container setup the env and then maybe use positional arguments for the commands. That would require a bit more reworking of things but I'll certainly think on it

config/sidekiq.yml
whereismyjetpack: Acionmailer dropped the messages on the mailers queue. I needed to setup sidekiq to read it in

bin/envconsul-wrapper
```diff
@@ -0,0 +1,45 @@
...
+if [ -f /vault/token ]; then
+    vault_token=$(cat /vault/token)
+    export VAULT_TOKEN=$vault_token
+fi
+
+# New versions of vault-init calls the file something slightly different. export vault token if that file is found
+if [ -f /vault/.vault-token ]; then
+    vault_token=$(cat /vault/.vault-token)
+    export VAULT_TOKEN=$vault_token
+fi
```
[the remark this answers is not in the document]
whereismyjetpack: Good catch.

bin/envconsul-wrapper
whereismyjetpack: The command can still run if the env is setup outside of vault. For testing etc.

bin/envconsul-wrapper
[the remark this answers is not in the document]
whereismyjetpack: `/vault/token` is no longer used, so i just removed that branch.

PR #485, opened 2020-08-19
Preview/refactor entrypoint

whereismyjetpack:
Refactors the method in which we startup the container. `envconsul-wrapper` sets up the environment, and then starts any argument we pass the container. 

This allows us to, for example, run scheduled tasks via kubernetes cron jobs by passing `args:` to it. i.e 
```
args:
  - bundle
  - exec
  - rake
  - some_awesome_task
```

without creating stub scripts like we attempted with the `monthly_stats_email` script. 

I also moved `start.sh` and `entrypoint.sh` out of $pwd to tidy things up a bit.


Dockerfile
```diff
@@ -42,5 +44,6 @@ RUN RAILS_ENV=production \
   rm -rf /app/tmp/
 
 
+ENTRYPOINT [ "/app/bin/envconsul-wrapper" ]
 
-CMD ["./entrypoint.sh"]
+CMD ["/app/bin/startup"]
```
awead, line 49: I'm assuming `ENTRYPOINT` must be specified prior to every `CMD` or can you just define it once and all commands are run through it afterwards?

Also, why does it look like we're starting the app twice?

whereismyjetpack:
@awead  `args` is to be a list, it's defined along with the deployment manifests (https://github.com/psu-stewardship/scholarsphere-config/blob/master/chart/templates/email-cron.yaml#L36)

Dockerfile
whereismyjetpack: `CMD` is passed to `ENTRYPOINT` as a positional argument, `ENTRYPOINT` is optional, but every Dockerfile needs a `CMD` and each stage needs a `CMD`  this is a multi-stage docker file, `base` and `production`

PR #488, opened 2020-08-21
Implementing featured resources

### LINKED ISSUE #417: Implement featured works
We need to be able to select which works will be featured on the front page. For now, this could be done without a UI, such as a form, and could be done solely as a database update via console or API.

awead:
Mark resources such as works and work versions, as part of a list of "featured" resources that will appear on the home page.

Fixes #417


app/models/featured_resource.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class FeaturedResource < ApplicationRecord
```
rschenk, line 3: I think you should use an activerecord polymorphic association here, with a foreign key constraint in the database. You're essentially reinventing that wheel via the UUID, but the UUID columns don't have any sort of key constraints on them. That will also allow you to load all the featured resources out in one database call in PagesController as well.

awead: @rschenk I thought about that, except there's only going to be a few of these in here. The pages controller only needs to load three. That seemed like over-optimization for such a small set of things. But, I can go that way if you think it's the better solution.

There's the other issue of sorting. I provided a sort field, but I think I need to provide some kind of constraint that there can only be one record with a sort value of "1". It's also getting quite overly-complex for a solution that only needs to support three featured works at a time. But I don't know of any other way to do it. 🤷‍♂️

app/controllers/pages_controller.rb
```diff
@@ -11,8 +11,10 @@ def determine_layout
       'frontend'
     end
 
-    # @note for now, just grab three randomly published work versions
-    def work_versions
-      WorkVersion.where(aasm_state: 'published').order(Arel.sql('RANDOM()')).limit(3)
+    def resources
+      FeaturedResource
```
rschenk, line 15: I would do `.includes(:resource)` here to eager load the resources and eliminate the n+1 query bug. Granted, you know that your _n_ is going to be 3-5, but I feel like it's a good habit to look out for and correct these situations.

awead: @rschenk I've added it, but I may need more explanation as to what it's actually doing. Is it going to preload the resource (work, work version, or collection) for _every_ record in the table?

PR #517, opened 2020-09-04
Initial stab at new work form

rschenk:


app/helpers/application_helper.rb
```diff
@@ -19,4 +19,11 @@ def link_to_dropdown_item(link, path)
       link_to link, path, class: 'dropdown-item'
     end
   end
+
+  def form_field_id(form, attribute)
```
awead, line 23: Used to override/modify a non-public Rails method. We should put a note here.

PR #522, opened 2020-09-15
adds after_commit_everywhere gem to fix deprication / race condition

whereismyjetpack:
"[DEPRECATION] :after_commit AASM callback is not safe in terms of race conditions and redundant calls."


awead (approved):

awead:
Installed as per recommendations in AASM Readme.

PR #523, opened 2020-09-18
Searching and adding new creators

### LINKED ISSUE #244: Interactive UI for searching and adding creators
Populate the work version edit form with existing creators taken from our local database and Penn State's identity API.

## WDLL
* search bar within creators that searches both sources and inserts new creators from the selection in the search bar
* UI feature parity with Scholarsphere 3 

Blocked by #242

awead:
This adds the bulk of the features for adding new creators to works. There are still outstanding accessibility issues and the requirements for metadata need to be changed. These can be addressed in forthcoming tickets.

Fixes #244


awead:
@rschenk I added some bit for a11y and styled the search input like the other inputs, but it may need some work.

* the input doesn't look quite right
* card heading text needs to be adjusted due to the icon size

Screen Shot 2020-09-24 at 12 06 15 PM

app/javascript/controllers/actors_controller.js
```diff
@@ -0,0 +1,40 @@
+import { Controller } from 'stimulus'
+import axios from 'axios'
+import { csrfToken, jumpToCreator } from './stimulus_modules'
+
+export default class extends Controller {
+  // Intercept Rails UJS's ajax:success callback and update the modal with the resulting content.
+  // This avoids having to have a separate .js.erb partial in a typical Rails AJAX setup.
+  success (event) {
+    const [data, status, xhr] = event.detail // eslint-disable-line no-unused-vars
+    document.getElementsByClassName('modal-content').item(0).innerHTML = data.body.innerHTML
```
rschenk, line 10: Even though there's only one modal in the layout (currently!), it would be more robust and more idiomatic javascript if the modal-content element had an ID, and this line here used `getElementById`. It's a good habit to try to use IDs and not classes whenever searching for elements in js (sometimes lookup by class is totally warranted, but much of the time it's not a good idea)

app/javascript/controllers/actors_controller.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  append (event) {
+    event.preventDefault()
+    const data = { actor_id: event.target.dataset.id }
+    data.index = document.getElementsByClassName('nested-fields').length
```
rschenk, line 16: There might be more than one set of nested-fields on the page in the future. In Christie's mockups, Creators, Collaborators, and Acknowledgements are all on the same tab (and might have nested fields). Probably a good idea to scope this to within `#creator_aliases`

app/javascript/controllers/autocomplete_controller.js
```diff
@@ -0,0 +1,61 @@
...
+          }
+        }
+      }
+    ]).on('autocomplete:selected', (event, suggestion, dataset, context) => {
+      this.appendSuggestion(suggestion)
+    })
+  }
+
+  appendSuggestion (suggestion) {
+    suggestion.index = document.getElementsByClassName('nested-fields').length
```
rschenk, line 32: Ditto with `actors_controller.js` Scope this `.nested-fields` to a particular element, in case we use .nested-fields class for multiple components on the page.

app/javascript/controllers/actors_controller.js
```diff
@@ -0,0 +1,40 @@
...
+      method: 'POST',
+      url: event.target.dataset.url,
+      data: data,
+      headers: {
+        'X-CSRF-Token': csrfToken()
+      }
+    })
+      .then(function (response) {
+        document
+          .getElementById('creator_aliases')
```
rschenk, line 28: Might be a cool idea to pass the `'creator_aliases'` id in as a data-attribute and really decouple this logic from the html

app/javascript/controllers/autocomplete_controller.js
```diff
@@ -0,0 +1,61 @@
...
+      method: 'POST',
+      url: this.data.get('post'),
+      data: suggestion,
+      headers: {
+        'X-CSRF-Token': csrfToken()
+      }
+    })
+      .then(function (response) {
+        document
+          .getElementById('creator_aliases')
```
rschenk, line 43: Might be a cool idea to pass the `'creator_aliases'` id in as a data-attribute and really decouple this logic from the html

app/javascript/controllers/stimulus_modules.js
```diff
@@ -0,0 +1,23 @@
...
+  if (node === null) {
+    return ''
+  } else {
+    return node.content
+  }
+}
+
+// Move to the last creator added to the form
+export function jumpToCreator () {
+  const nodes = document.getElementsByClassName('card-body')
```
rschenk, line 15: This seems really brittle, because this is a bootstrap class used for styling. If we decided to restyle these, it would break this javascript. Best bet here would be to create a new class for creator fields, and if you want to go full BEM style use a "js" prefix so someone coming in to tweak the html would know to think twice before changing it, like  `js-creator-fields`

rschenk (changes requested):
I had a couple really specific technical comments with the javascript, but I also have a couple UX feedbacks too... 

* I wonder if we need another method to pop open the "new actor" modal, besides completely empty search results. It's possible that I could search for someone who is not in the list, and yet still have search results—in that scenario I can't add them. Perhaps there should always be a "add someone not in this list" at the very bottom of the auto complete?

* If the creator search field loses focus, I can't get the list to pop open again. If I do a search for "Wead", click away from the field so that it loses focus, then focus it again, the results list does not reappear. I have to change the text, like delete the "d" and add it again

* In the New Creator modal, I'm not sure if the two-step process of making them confirm the Actor, then click Add Creator to Work on the second page is useful. I can't see any reason why a user _wouldn't_ want to add them to the work, since this is why they opened the dialog to begin with. I think making them confirm will be confusing and lead to errors. I reckon if the New Creator form submits correctly without validations, it should just be inserted into the main form. That is the desired action after all.

* We really need to beef up our validations and constraints on the Actor model/table. For example, I can add multiple people with the same PSU access ID. Yikes.

* Despite not totally following Christie's design for this page, I do think it'd be nice to implement most of the styling. I will attach the drawing as a separate comment.

rschenk:
Even if we aren't going completely with Christie's UX design with the modals, I do think it'd be nice to update our forms to use some/most of her visual styling

contributors2

addCreators

awead:
> I had a couple really specific technical comments with the javascript, but I also have a couple UX feedbacks too...
> 
>     * I wonder if we need another method to pop open the "new actor" modal, besides completely empty search results. It's possible that I could search for someone who is not in the list, and yet still have search results—in that scenario I can't add them. Perhaps there should always be a "add someone not in this list" at the very bottom of the auto complete?
> 

fixed!

>     * If the creator search field loses focus, I can't get the list to pop open again. If I do a search for "Wead", click away from the field so that it loses focus, then focus it again, the results list does not reappear. I have to change the text, like delete the "d" and add it again
> 

fixed!

>     * In the New Creator modal, I'm not sure if the two-step process of making them confirm the Actor, then click Add Creator to Work on the second page is useful. I can't see any reason why a user _wouldn't_ want to add them to the work, since this is why they opened the dialog to begin with. I think making them confirm will be confusing and lead to errors. I reckon if the New Creator form submits correctly without validations, it should just be inserted into the main form. That is the desired action after all.
> 

fixed. New actors are added directly to the form.

>     * We really need to beef up our validations and constraints on the Actor model/table. For example, I can add multiple people with the same PSU access ID. Yikes.
> 

ticket separately?

>     * Despite not totally following Christie's design for this page, I do think it'd be nice to implement most of the styling. I will attach the drawing as a separate comment.

amended!

rschenk (approved):

PR #538, opened 2020-09-23
Style validation error messages on work form

### LINKED ISSUE #528: Style form error messages and fields that have errors
In the Review and Publish tab, errors preventing publication need to be styled correctly. Also, individual fields with errors need to be styled as well.

rschenk:
Styles validation errors to the work form (and elsewhere).

* Adds a big error message block to the top of the form
* Adds red border and background to form fields that have errors
* Adds individual error messages underneath the form field in question

Closes #528 

Screen Shot 2020-09-23 at 2 06 57 PM


awead (changes requested):
What about a component? You can use the `render?` method like so:

``` ruby
class ErrorMessageComponent
  attr_reader :form

  def initialize(form:)
    @form = form
  end

  def render?
    form.object.errors.any?
  end

  def heading
    [...]
  end

  def messages
    [...]
  end
end
```
Then you don't have to indent the whole erb template, and it's a bit easier to invoke:
` render ErrorMessageComponent.new(form: form)`

rschenk:
Good idea @awead

awead (approved):
💥

PR #561, opened 2020-10-05
Link dashboard work versions to their resources pages

### LINKED ISSUE #560: Update work version links in the dashboard
Links in the dashboard need to point to the resource page for work versions.

awead:
Updates the dashboard search results to link work versions to their resource pages. Includes a new NullWorkVersion object as a stand-in for works that do have a published versions. This follows the "null object pattern" and will help us in the future when dealing with works that have no published version.

Fixes #560

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -73,3 +73,7 @@ Style/ClassVars:
 Style/EmptyMethod:
   Exclude:
     - 'app/jobs/metadata_listener/job.rb'
+
+Style/MethodMissingSuper:
+  Exclude:
+   - 'app/models/null_work_version.rb'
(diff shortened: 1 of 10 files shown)


rschenk (approved):
I think it's good for now, but like we talked about in Zoom, it might eventually be better to have two different `VersionNavigationDropdownComponent`s, one for resources that the current_user has edit access to, and one that they don't. The "edit access" one would show drafts and would also show the "Draft" and "Published" badges next to the versions. The "read access" one would only show published versions with no need for the badges. That might let us be a little more targeted with how this works, and whether it needs the Null Object, perhaps. I do like the Null Object pattern though, I think that's nice.

PR #562, opened 2020-10-06
Update ORCiD validation

### LINKED ISSUE #552: ORCiD storage
Store orcid's as a string of numbers only, with no dashes. Ensure that the uniqueness validation enforces no identical orcids regardless of formatting.

awead:
Removes the OrcidValidator class in favor of a simpler regex-based verification that ensures the id will be stored in the database as a 16-digit string. The Orcid model is updated to handle verification and translations between the different formats. While the model is not used directly for validation, it can be used to validate data from other sources, or present numeric ids in a more human-readable format.

Fixes #552


app/models/orcid.rb
```diff
@@ -1,34 +1,34 @@
...
 
   def to_s
-    "https://orcid.org/#{parsed_id}"
+    id
   end
 
-  def uri
-    URI(to_s)
+  def to_human
+    id.gsub(/(\d{4})(?!$)/, '\1-')
```
rschenk, line 28: Slick!

rschenk (dismissed):

app/models/orcid.rb
awead: @rschenk I stole that from somewhere, I know I did, I just can't remember where.

rschenk: @awead good artists borrow, great artists steal

PR #563, opened 2020-10-06
Strict enforcement of EDTF on date fields

### LINKED ISSUE #548: Revist EDTF validations
Certain "dates" are validating as EDTF, such as _January_ and _17th Century_. Maybe we don't want that and want something more stringent. This is because "January" is a valid Ruby date but not valid EDTF.

```
> DateTime.parse("January")
=> Wed, 01 Jan 2020 00:00:00 +0000]

DateTime.parse("17th Century")
=> Sat, 17 Oct 2020 00:00:00 +0000
```

## WDLL

* allow _only_ EDTF dates and not Ruby dates

rschenk:
The EDTF validation code came from CHO. In CHO it was a feature that the fields should accept either valid Ruby dates, or valid EDTF dates, therefore things like "17th Century" were considered valid. This behavior is a bug in Scholarsphere, so I've updated the validator to require strict compliance to EDTF.

Closes #548

### DIFF
--- a/app/services/edtf_date.rb
+++ b/app/services/edtf_date.rb
@@ -2,15 +2,7 @@
 
 module EdtfDate
   def self.valid?(str)
-    regular_date = begin
-                     Date.parse(str)
-                   rescue ArgumentError, TypeError
-                     nil
-                   end
-
-    edtf_date = Date.edtf(str)
-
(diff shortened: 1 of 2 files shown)


awead (approved):

awead:
🎸

PR #565, opened 2020-10-06
Switch sidebar with main content

### LINKED ISSUE #480: Readable ordering for sidebar layouts
Pages that utilize a sidebar/content-on-the-right layouts, need to "read" in an order that reflects the layout of the page. The main content that is present on the right-hand side of the page should be read by a screen reader before the content in the side bar. Currently, the reverse happens, and a screen reader reads the sidebar content before reading the main content.

## WDLL
* swap sidebar and main areas in the html

## Example

The circled `div.container-fluid` should come before the `aside.sidebar` in the HTML while maintaining the same layout.

Screen Shot 2020-08-18 at 9 31 08 AM

awead:
For accessibility reasons, it's preferred to have a screen reader present the main content first, before it presents any sidebar content. Because of layout reasons, this content was appearing *after* the sidebar. Using Bootstrap's reordering classes, we can present the main content fist in the page structure, but have it display second, after the sidebar.

Fixes #480

### DIFF
--- a/app/views/layouts/_with_sidebar.html.erb
+++ b/app/views/layouts/_with_sidebar.html.erb
@@ -1,12 +1,12 @@
 <div class="container-fluid">
   <div class="row">
-    <div class="col-lg-3">
+    <div class="col-lg-9 order-last">
+      <%= yield %>
+    </div>
+    <div class="col-lg-3 order-first">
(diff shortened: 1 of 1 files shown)


rschenk (approved):
All hail flexbox!

PR #567, opened 2020-10-06
Mint a DOI via background job

### LINKED ISSUE #105: DOI Job
Write an async job that runs the DOI service that was build in #103. The job would be called any time the service is invoked, such as to create a new DOI or update an existing one. It's only requirement is that it executes the service with the provided parameters. The only thing the job needs to do is run the service async via Sidekiq/Redis.

Blocked by #104 & #103 

## WDLL

* enqueues job
* updates Redis status as needed based on results from the minting service

## Todos

Let's worry about the update process later and focus just on minting new DOIs for new or existing works and work versions.

rschenk:
The entrypoint for this (aka, what the controller(s) will use) is `MintDoiAsync.call`. That method sets the status of the doi to `waiting` and fires off a background job. From there, the DoiMintingJob will eventually take over, and mint the DOI using our existing service while keeping the status updated accordingly

Closes #105

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -66,6 +66,11 @@ RSpec/AnyInstance:
 RSpec/NestedGroups:
   Enabled: false
 
+RSpec/MessageSpies:
+  Exclude:
+    - 'spec/jobs/doi_minting_job_spec.rb'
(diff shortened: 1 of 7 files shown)


awead (approved):
Nice!

PR #569, opened 2020-10-07
Increase wait times for async creator test

awead:
CI builds were failing intermittently. This should allow more time for the DOM elements to load.


awead:
We're still hitting the live endpoint for Penn State's identity service, so that could be why there are longer load times.

whereismyjetpack:
could we do an A/B test where we skip those for a CI run? I'd be curious if it would be worth the effort of VCRing them

rschenk (approved):

awead:
Follow up work ticketed in #570

PR #580, opened 2020-10-14
Subclass WorkVersion to index immidately in the dashboard

### LINKED ISSUE #501: Index new works synchronously
When creating a new work, we need to index into Solr immediately so that it will appear in the user's dashboard results if they save and exit before adding more metadata.

## WDLL

* `update_index_sync` to work version model on `after_create`

awead:
Dashboard::WorkVersion serves as an identical stand-in to WorkVersion within the user's dashboard. This allows for synchronus updating of the work version in Solr, so any changes will appear immediately to the user. Outside of the dashboard, such as API calls, in the console, or elsewhere, the updates are performed async via ActiveJob.

Fixes #501


awead:
We found a better way to do this.

PR #581, opened 2020-10-14
Sorting order

### LINKED ISSUE #555: Correct sorting order in Dashboard
Dashboard search results are being sorted in inverse relevance.

### LINKED ISSUE #556: Sorting order in search results
Public search should sort according deposited date, dashboard should sort according to updated_at for the work version.

awead:
Sorts public searches according to date deposited and user dashboard searches according to date updated.

Fixes #555 
Fixes #556

### DIFF
--- a/app/models/dashboard/postgres_response.rb
+++ b/app/models/dashboard/postgres_response.rb
@@ -13,6 +13,7 @@ module Dashboard
 
       def work_versions
         WorkVersion.where(uuid: work_version_uuids)
+          .order(updated_at: :desc)
           .compact
           .map { |work_version| WorkVersionDecorator.new(work_version) }
       end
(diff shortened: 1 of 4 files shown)


app/models/dashboard/postgres_response.rb
```diff
@@ -13,6 +13,7 @@ def documents
 
       def work_versions
         WorkVersion.where(uuid: work_version_uuids)
+          .order(updated_at: :desc)
```
rschenk, line 16: I think this works for now, but it will break when we start interlacing Collections into the mix. In that case we'll need to do the ordering in the Solr query, and then maintain the order of the results when we map them to ActiveRecords... but we will _already_ need to do that to add collections into the list, so maybe that will inherently sort itself out (pun intended)

rschenk (approved):

PR #582, opened 2020-10-14
Index work versions synchronously in the dashboard

### LINKED ISSUE #501: Index new works synchronously

awead:
When updating or creating work versions via the user's dashboard, changes should be made to Solr immediately to avoid any lags in displaying search results. This way, when a user creates or updates a work version and then navigates back to the dashboard search page, they'll see the changes instead of having to wait for the indexing job to finish.

Fixes #501

### DIFF
--- a/app/controllers/dashboard/work_form/base_controller.rb
+++ b/app/controllers/dashboard/work_form/base_controller.rb
@@ -29,6 +29,16 @@ module Dashboard
         def next_page_path
           raise NotImplementedError, 'You must implement this method in your controller subclass'
         end
+
+        def update_or_save_work_version(attributes: nil)
+          @work_version.indexing_source = SolrIndexingJob.public_method(:perform_now)
+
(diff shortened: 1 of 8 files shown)


rschenk (approved):
This looks great, just one question: does this work for deleting a version?

awead:
@rschenk I took care of deletions in a previous commit 80d07917b7607ebdc2327f17676d91399d23843a and they're done synchronously, via a service and without a callback, regardless of whether you're in the dashboard or not. At this time, I dodn't see a need for async deletes, but we might want to revisit that later. For now, if we need to delete a bunch of records, we can script that ourselves and do a batch delete in Postgres, followed by a batch delete in Solr.

PR #583, opened 2020-10-15
DOI minting ui

### LINKED ISSUE #8: DOI Integration
Leaning towards having a DOI for a work not for each version of the work. If you would like to have a version to be attached, you can, but it is not the default (to have a version for each work).

## Related Tickets

- [x] Mapping Datacite DOI states to Scholarsphere work states #99
- [x] Add DOI fields to Work and WorkVersion #100
- [x] Mapping metadata to Datacite #101
- [x] Datacite client service #102
- [x] DOI Service #103
- [x] Implement aliases #186 
- [x] Sidekiq and Redis #104
- [x] DOI Job to run the DOI service #105 
- [ ] Mint DOIs for works and work versions (with UI) #150
- [ ] Map our work_type metadata field to DataCite's resourceTypeGeneral #576

### LINKED ISSUE #150: DOI Minting UI Component
Provide UI elements for minting DOIs on works and work versions. @srerickson 

## WDLL

* controller to respond to link
* controller kicks off service
* service sets a pending status in Redis and enqueues job
* DOI field in UI will display status based on Redis key
* link will be greyed out
* by default, the DOI should be on the *work*

If the work doesn't have a DOI, then the DOI is on the work, regardless of what version you're looking at. You can only mint a DOI for a work version if the work already has a DOI, and you're viewing the resource page for that specific version.

### LINKED ISSUE #576: Map our work_type metadata field to DataCite's resourceTypeGeneral
DataCite requires us to provide a value for the field `resourceTypeGeneral`. I need a mapping that translates our `work_type` to DataCite's.

This needs to be completed before we release the DOI minting UI

Our list is below:
```
article
audio
book
capstone_project
collection
conference_proceeding
dataset
dissertation
image
journal
map_or_cartographic_material
masters_culminating_experience
masters_thesis
other
part_of_book
poster
presentation
project
report
research_paper
software_or_program_code
thesis
unspecified
video
```

rschenk:
Closes #150 
Closes #576
Closes #8

### DIFF
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -19,26 +19,13 @@ class DoisController < ApplicationController
 
   private
 
-    # @note We have some ridiculous hoops to jump through here.
-    # 1. We can only send an actual, undecorated ActiveRecord through ActiveJob
-    # 2. The ResourceDecorator and associated subclasses define a method called
-    #    `resource_with_doi` because the WorkVersion cannot currently receive a
-    #    doi, only the parent work can.
-    # Therefore, we need to decorate the resource, ask it what the
-    # resource-with-doi is, then un-decorate the returned object.
+    # At the moment, WorkVersions cannot get their own DOIs, their parent work
+    # does. This rule is encapsulated by the #resource_with_doi interface on
+    # Collections, Works, and WorkVersions
     def find_resource
       # Note the param is called :resource_id, but it's actually a uuid
-      undecorated_resource = FindResource.call(params[:resource_id])
-      decorated_resource = ResourceDecorator.decorate(undecorated_resource)
-      resource_with_doi = decorated_resource.resource_with_doi
-
-      # If resource_with_doi is decorated (`__getobj__`), undecorate before
-      # returning
-      if resource_with_doi.respond_to?(:__getobj__)
-        resource_with_doi.__getobj__
-      else
-        resource_with_doi
-      end
+      requested_resource = FindResource.call(params[:resource_id])
+      requested_resource.resource_with_doi
     end
 
     def doi_present_or_minting?(resource)
--- a/app/controllers/dois_controller.rb
+++ b/app/controllers/dois_controller.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+class DoisController < ApplicationController
+  class DoiIsMintingError < RuntimeError; end
+
+  def create
+    resource = find_resource
+
+    authorize(resource, :edit?)
+    raise DoiIsMintingError if doi_present_or_minting?(resource)
+
+    MintDoiAsync.call(resource)
+  rescue DoiIsMintingError
+    # I don't think we need to do anything special here, just skip over the call
+    # to MintDoiAsync and redirect back where we came from?
+  ensure
+    redirect_to resource_path(params[:resource_id])
+  end
+
+  def show
+    @resource = find_resource
+    render :show, layout: false
+  end
+
+  private
+
+    # @note We have some ridiculous hoops to jump through here.
+    # 1. We can only send an actual, undecorated ActiveRecord through ActiveJob
+    # 2. The ResourceDecorator and associated subclasses define a method called
+    #    `resource_with_doi` because the WorkVersion cannot currently receive a
+    #    doi, only the parent work can.
+    # Therefore, we need to decorate the resource, ask it what the
+    # resource-with-doi is, then un-decorate the returned object.
+    def find_resource
+      # Note the param is called :resource_id, but it's actually a uuid
+      undecorated_resource = FindResource.call(params[:resource_id])
+      decorated_resource = ResourceDecorator.decorate(undecorated_resource)
+      resource_with_doi = decorated_resource.resource_with_doi
+
+      # If resource_with_doi is decorated (`__getobj__`), undecorate before
+      # returning
+      if resource_with_doi.respond_to?(:__getobj__)
+        resource_with_doi.__getobj__
+      else
+        resource_with_doi
+      end
+    end
+
+    def doi_present_or_minting?(resource)
+      resource.doi.present? || DoiMintingStatus.new(resource).present?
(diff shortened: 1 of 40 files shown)


app/controllers/dois_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+      decorated_resource = ResourceDecorator.decorate(undecorated_resource)
+      resource_with_doi = decorated_resource.resource_with_doi
+
+      # If resource_with_doi is decorated (`__getobj__`), undecorate before
+      # returning
+      if resource_with_doi.respond_to?(:__getobj__)
+        resource_with_doi.__getobj__
+      else
+        resource_with_doi
+      end
```
awead, line 46: I may be missing something, but couldn't this be simplified to:

``` ruby
resource = FindResource.call(params[:resource_id])

return resource.try(:work) || resource
```

It just returns whatever the work is. Are we dealing with collections too?

app/policies/work_policy.rb
```diff
@@ -13,4 +13,15 @@ def limit
...
+
+  def edit?
+    editable?
+  end
+
+  private
+
+    # @todo There's a bug in the permissions because a depositor should have edit access by default
+    def editable?
+      record.edit_access?(user) || record.depositor == user.actor || user.admin?
```
awead, line 25: Can we move this to ApplicationPolicy? I think it would clear up the confusion. Also, I don't think it's a bug anymore because this was put it back before we decided that depositors would not have their own ACL.

app/jobs/doi_minting_job.rb
```diff
@@ -4,14 +4,22 @@ class DoiMintingJob < ApplicationJob
...
-    status = DoiStatus.new(resource)
+    status = DoiMintingStatus.new(resource)
 
     status.minting!
     DoiService.call(resource)
     status.delete!
   rescue DoiService::Error => e
+    handle_error(e, status)
+  rescue DataCite::Client::Error => e
+    handle_error(e, status)
```
awead, line 15: This line isn't getting covered.

app/jobs/doi_minting_job.rb
```diff
@@ -4,14 +4,22 @@ class DoiMintingJob < ApplicationJob
...
+  rescue DataCite::Metadata::Error => e
+    handle_error(e, status)
```
awead, line 17: This line isn't getting covered either. Not sure if it's important or not.

awead (changes requested):

app/controllers/dois_controller.rb
rschenk: We _totally could be_ dealing with collections in the future, yes. Also there's no guarantee that the decorator has not overloaded `#work` to return a WorkDecorator. `__getobj__` is a feature of `SimpleDelegator` (the thing we use to make decorators) that gives you access to the raw, undecorated object being wrapped

app/controllers/dois_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  rescue DoiIsMintingError
+    # I don't think we need to do anything special here, just skip over the call
+    # to MintDoiAsync and redirect back where we came from?
+  ensure
+    redirect_to resource_path(params[:resource_id])
+  end
+
+  def show
+    @resource = find_resource
+    render :show, layout: false
```
awead, line 22: @rschenk I'm also seeing this not getting covered either, but you're redirecting to it in the tests.

awead (changes requested):

app/controllers/dois_controller.rb
awead: But, FindResource is returning the undecordated object, which is ultimately what you want to return, correct? You've defined `resource_with_doi` to be used to display a MintableDoiComponent but here it seems like you're jumping through a lot of hoops just to reuse that method. I understand that reasoning, but for clarity, wouldn't 

``` ruby
return resource.work if resource.is_a?(WorkVersion)
resource
```

Or something to that effect, be easier to understand? I realize that probably violates some principle because the concept of what does nor does not have a doi is now in two different places.

rschenk: Yes, then you've duplicated the logic of "a WorkVersion shouldn't have a DOI, only its parent work should"

rschenk: Another option might be to put `#resource_with_doi` onto the models themselves, not on the decorator, since that is a fairly important piece of business logic?

awead: Yes, I was just going to suggest that. Or have `WorkVersion#doi` return nil. On the other hand, work versions will _eventually_ have dois so all this business can go away.

app/controllers/dois_controller.rb
rschenk: Come to think of it, this controller action is not currently used at all. I intended for it to feed an ajax auto-refresher when the DOI is minting, but I never built that auto-refresher. I think I'll remove it and its associated view for now. All it did was render a MintingStatusDoiComponent with no layout

app/controllers/dois_controller.rb
rschenk: OK, I'll relocate `#resource_with_doi` into the models, I think that's a better place for it anyways than hiding out in a decorator

app/controllers/dois_controller.rb
awead: Yay for less code!

app/policies/work_policy.rb
rschenk: We are going to 🏈 on this one, and move it into a new ticket, because it has some bigger implications and refactoring possibilities.

awead: @rschenk Let's pass on this. It's going to take more work that originally though, but I'll ticket it for future reference. See #586

rschenk:
Screen Shot 2020-10-15 at 3 58 04 PM
Screen Shot 2020-10-15 at 3 58 28 PM

awead (approved):

PR #584, opened 2020-10-15
Render zero results page in the dashboard

### LINKED ISSUE #545: Page for no results in a dashboard search
When on the the user's dashboard, if the search result returns no results we should see a "no results" page instead of the current "welcome to the dashboard" page.

awead:
If the user has works, but their search returns no results, it should look similar to the public "no search results" page. This refactors the view code to make them identical in layout and features, except with changes in text to indicate the differences.

Fixes #545

### DIFF
--- a/app/views/catalog/_search_results.html.erb
+++ b/app/views/catalog/_search_results.html.erb
@@ -1,0 +1,25 @@
+<h2 class="sr-only"><%= t('blacklight.search.search_results') %></h2>
+
+<div class="action-bar mb-3">
+  <%= render_search_bar %>
+</div>
+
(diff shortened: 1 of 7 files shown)


rschenk (commented):
@awead I assume the fact that the search results partial is in `app/views/catalog/_search_results` while the zero results partial is nested under dashboard `app/views/dashboard/catalog/_zero_results` is intentional and some kind of Blacklight thing?

awead:
@rschenk and, yes, zero results is a view in the Blacklight gem.

rschenk (approved):

PR #585, opened 2020-10-15
Make License validations more user-friendly on Work Form

rschenk:
Our field is called "rights" but I felt like "License" was a more self-explanatory label.

I also tweaked the inclusion validation, because if the rights field was blank, you would get two error messages, one from the presence validator telling you it couldn't be blank, and another one from the inclusion validator telling you that nil wasn't in the list. I changed it so the inclusion validation allows a nil value, thereby bypassing the inclusion validation when we know the presence validator will pick up those errors.

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -117,7 +117,10 @@ class WorkVersion < ApplicationRecord
 
   validates :rights,
             presence: true,
-            inclusion: { in: WorkVersion::Licenses.ids },
+            inclusion: {
+              in: WorkVersion::Licenses.ids,
+              allow_nil: true # Avoid duplicating the above presence validation
+            },
             if: :published?
 
   validates :published_date,
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -37,7 +37,7 @@ en:
         version_name: Semantic Version (optional)
         subtitle: Subtitle
         keyword: Keyword
-        rights: Rights
+        rights: License
         description: Description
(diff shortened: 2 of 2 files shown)


app/models/work_version.rb
```diff
@@ -117,7 +117,10 @@ def label(id)
 
   validates :rights,
             presence: true,
-            inclusion: { in: WorkVersion::Licenses.ids },
+            inclusion: {
+              in: WorkVersion::Licenses.ids,
+              allow_nil: true # Avoid duplicating the above presence validation
```
awead, line 122: Rights can't be nil. Are you saying we can take out `presence: true`?

awead (changes requested):

app/models/work_version.rb
rschenk: No, the allow_nil is on the inclusion validation only. 

Previously if you had a nil value in the rights you would get two different validation errors:
- Rights can't be blank
- Rights is not included in the list [Because nil is not a member of the list]

That's super confusing and unintuitive. If you pop `allow_nil: true` onto the inclusion validation, it _still_ will not allow rights to be nil, because you have the presence validation on there, but it will not duplicate the two error messages

rschenk: Before: 
Screen Shot 2020-10-15 at 1 23 56 PM

After:
Screen Shot 2020-10-15 at 1 23 42 PM

awead (approved):

PR #588, opened 2020-10-16
watch doi queue, too

whereismyjetpack:


awead:
@whereismyjetpack good catch!

awead (approved):

PR #614, opened 2020-10-20
Fix overflowing x-axis labels in analytics chart

### LINKED ISSUE #466: Analytics padding bug
When looking at analytics the left most date can be cut off - below should say Mon 03 and display Ion 03 

Screen Shot 2020-08-14 at 9 43 27 AM

rschenk:
In #466 we saw that the x-axis labels on the view analytics chart could be clipped off in confusing ways, if they were positioned too close to the edge and overflowed beyond the parent SVG element.

I fixed this in two ways: 
- Increased the horizontal padding on the chart quite a bit, which will prevent the issue from happening the vast majority of the time.
- Added a backup function, executed after the x-axis labels are drawn, that hides any labels whose left or right edge extends beyond the bounds of the parent SVG element and thus are clipped.

Closes #466

### DIFF
--- a/app/javascript/frontend/scripts/view_statistics_chart.js
+++ b/app/javascript/frontend/scripts/view_statistics_chart.js
@@ -22,7 +22,7 @@ function parseData (data) {
 }
 
 function drawChart (selection, data) {
-  const margin = { top: 2, right: 10, bottom: 20, left: 10 }
+  const margin = { top: 2, right: 25, bottom: 20, left: 25 }
   const width = 350
   const height = 80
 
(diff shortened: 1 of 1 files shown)


rschenk:
Oh yeah I forgot that CodeClimate has a spasm about this file, because it's duplicated in both `app/javascripts/frontend` and `app/javascripts/scholarsphere`

awead:
@rschenk can we just delete the duplicate file? I think we're safely away from the old `scholarsphere` UI pack.

rschenk:
@awead it makes no difference to me when we delete the old pack. I know we were keeping it around for insurance until we'd finished implementing the new design, but like you said, we're pretty close to being there

awead:
@rschenk meh. I'd rather delete all of them at once, so I approved the duplication for now.

awead (approved):

PR #626, opened 2020-10-22
Move user menu up to the white top navbar

### LINKED ISSUE #490: Move the user dropdown to the white bar above
The user dropdown/login button should be moved to the upper-righthand corner, directly above (and out of) the blue bar where it is now.

awead:
Takes the existing user drop-down menu and moves it up to the white navbar above. Because the blue navbar will now have no menu items in it on certain pages, there needs to be a "spacer" to ensure headings will display.

Fixes #490 

Screen Shot 2020-10-22 at 11 59 04 AM

## Caveats

We're trying to use the navbar in a way that it wasn't intended for: as a heading. To do this, we have to "fake" some space, which has some consequences because the page still renders it like a navbar in some cases:

Screen Shot 2020-10-22 at 11 59 17 AM

### DIFF
--- a/app/javascript/frontend/styles/_common.scss
+++ b/app/javascript/frontend/styles/_common.scss
@@ -329,6 +329,11 @@ h4,
       align-items: center;
     }
   }
+
+  .navbar-filler {
+    display: inline-block;
+    padding: ($spacer * 1.5) 0;
+  }
 }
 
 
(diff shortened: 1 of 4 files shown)


rschenk (approved):
I tried out a few other techniques, but I honestly think this is as good as any.

PR #629, opened 2020-10-23
Update policies to restrict viewing of draft works

### LINKED ISSUE #566: Draft works should not be publicly accessible
Draft works are not index, but can currently be accessed if the url is known. This needs to change so that even if the url is known, it will return 404 unless the current user has edit access.

## WDLL

* draft works are accessible via the `/resources` path only to users who have edit access
* depositors have access
* anyone with an edit ACL

awead:
This has two commits, with two different purposes, but they each overlap.

## Merging Dashboard Policies

This eliminates the Dashboard module of our polices and combines all of the policy control to the main application policy classes. It also includes a considerable amount of refactoring, and refining existing policies to cover all the edge cases.

The policies now (officially) combine both ACL and non-ACL logic into one place. For example, non-ACL based permissions such as access via depositor or proxy ownership, as well as administrative access, are combined with ACL-based permissions such as edit, read, and discover. As a result, asking a policy for a type of access will interrogate all those dimensions, both ACL and non-ACL.

Because of the varied nature of the permissions and because Solr drives the search features, policy scopes are largely ignored. This is reflected in the changes to controllers were scoping find operations are either superfluous, or not accurate.

## Restrict Draft Works

Prevents viewing and downloading draft works and work versions to anyone who does not have edit access.

Fixes #566


rschenk (approved):
Epic

PR #633, opened 2020-10-26
Edit, Create New Version buttons on resource page

### LINKED ISSUE #491: Add additional actions to resource page
If the user is logged in and has the permissions, they should have the same buttons/actions available to them that they would have on the dashboard page. This makes the resource page and dashboard view page appear basically the same.

rschenk:
If the user is logged in and has the permissions, they can now edit a draft, or create a new draft from the resource page. 

Eventually we would like to remove those buttons from the Dashboard "search results" view and have them only on the resource page. To further this goal, I added a second commit that changes the paths of the work form's Cancel and Save Draft & Exit buttons to go to the resource page rather than back to the dashboard root—this was Seth's suggestion (which I agree with) when we were going through the UX last week.

Closes #491

### DIFF
--- a/app/components/link_disabled_by_tooltip_component.html.erb
+++ b/app/components/link_disabled_by_tooltip_component.html.erb
@@ -1,3 +1,3 @@
-<%= link_to_if enabled?, text, path, html_options do %>
-  <%= content_tag(:span, text, html_options) %>
+<%= link_to_if enabled?, text, path, html_options_enabled do %>
+  <%= content_tag(:span, text, html_options_disabled) %>
 <% end %>
--- a/app/components/link_disabled_by_tooltip_component.html.erb
+++ b/app/components/link_disabled_by_tooltip_component.html.erb
@@ -1,0 +1,3 @@
+<%= link_to_if enabled?, text, path, html_options do %>
+  <%= content_tag(:span, text, html_options) %>
+<% end %>
--- a/app/components/link_disabled_by_tooltip_component.rb
+++ b/app/components/link_disabled_by_tooltip_component.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+class LinkDisabledByTooltipComponent < ApplicationComponent
+  def initialize(enabled:, text:, path:, tooltip:, method: :get, class_list: nil)
+    @enabled = !!enabled
+    @text = text
+    @path = path
+    @tooltip = tooltip
+    @method = method
+    @class_list = class_list || default_classes
+  end
+
+  private
+
+    attr_reader :enabled,
+                :text,
+                :path,
+                :tooltip,
+                :method,
+                :class_list
+
+    def enabled?
+      enabled
+    end
+
+    def disabled?
+      !enabled?
+    end
+
+    def html_options
+      { class: classes }
+        .deep_merge(tooltip_options)
+        .deep_merge(link_options)
+    end
+
+    def classes
+      klasses = Array.wrap(class_list).dup
+      klasses << 'disabled' if disabled?
+      klasses
+    end
+
+    def tooltip_options
+      return {} if enabled?
+
+      {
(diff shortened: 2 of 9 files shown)


app/components/link_disabled_by_tooltip_component.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+class LinkDisabledByTooltipComponent < ApplicationComponent
+  def initialize(enabled:, text:, path:, tooltip:, method: :get, class_list: nil)
+    @enabled = !!enabled
```
awead, line 5: Does this equate to `true` by default?

app/components/link_disabled_by_tooltip_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= link_to_if enabled?, text, path, html_options do %>
+  <%= content_tag(:span, text, html_options) %>
```
awead, line 2: are the link tag and content tag's html options always the same?

app/controllers/dashboard/work_form/base_controller.rb
```diff
@@ -30,6 +30,15 @@ def next_page_path
           raise NotImplementedError, 'You must implement this method in your controller subclass'
         end
 
+        helper_method :cancel_path
+        def cancel_path
+          if @work_version.present? && @work_version.persisted?
+            resource_path(@work_version&.uuid)
```
awead, line 36: Does this avoid the reloading of the work version because the database created the UUID? Otherwise, I'm not sure why you'd need the `&`.

app/controllers/dashboard/work_form/base_controller.rb
```diff
@@ -19,7 +19,7 @@ def save_and_exit?
 
         def redirect_upon_success
           if save_and_exit?
-            redirect_to dashboard_root_path,
+            redirect_to resource_path(@work_version&.uuid),
```
awead, line 22: Ditto here. A work version should always have a uuid, unless it's just been created and need to be refreshed.

awead (changes requested):
@rschenk looks good. I just had some questions.

app/components/link_disabled_by_tooltip_component.rb
rschenk: This is a Ruby idiom to ensure you've got a boolean value. It's a required keyword, so it doesn't default to anything

app/components/link_disabled_by_tooltip_component.html.erb
rschenk: Kinda yes but no. It decides whether to make a link or a span depending on the value of `enabled?`. The keys and values inside `html_options` will also be different depending on the value of `enabled?`. 

If enabled is false (when we render a span) then there will be items inside `html_options` for the tooltip behavior, and the class list will contain a `disabled` class. If enabled is true, there will be no tooltip behavior items, but there will be link-specific stuff in there (such as `method`)

app/controllers/dashboard/work_form/base_controller.rb
rschenk: Yeah good point. I guess I didn't want to get into a race condition with the db, but it's probably not necessary

app/components/link_disabled_by_tooltip_component.html.erb
rschenk: I suppose another way to do this would be two different methods... `link_html_options` and `span_html_options`?

rschenk: @awead I refactored `html_options` into two different methods (one for enabled, and one for disabled) and the code is MUCH cleaner!

awead (approved):
Hooray for code reviews! I got to learn something _and_ give valuable feedback.

PR #634, opened 2020-10-26
🐞 WorkVersionPolicy#new? checks to see if work is editable

rschenk:


awead (changes requested):
@rschenk I'm wondering what this means now. I'm confused about why we need the `latest_version` parameter. We always create a new work version from the most recent published version, if there's no existing draft. So, if I'm understand this correctly, there are three criteria:

1. You can edit
2. The current record is published
3. There are no other existing drafts, i.e. the current (published) record is the same as the most recent version

``` ruby
def new?
  return false unless editable?

  record.published? && record == record.work.latest_version
end
```

Maybe something's been lost in the translations from the original dashboard use cases.

rschenk:
@awead Yeah there was some reason why I had injected the latest version into this method, I think to avoid an n+1 query bug when we were backing the dashboard via activerecord? But you're right, that's not necessary anymore. I like your solution above better

rschenk:
@awead Our thinking about this issue had become very uptight, man.

We already have a method `WorkPolicy#create_version?` which answers the question "can a new version theoretically be created for this Work?" (i.e. is it editable, and does it already have a draft?). 

The method we're discussing, `WorkVersionPolicy#new?` answers the question "can a new version be created _from this specific version?"_

I was able to re-write it to require no parameters, as you described above, but also to use the WorkPolicy for the editable and draft stuff.

awead:
@rschenk Well, you know, there are a lotta ins and out to this, a lotta strings to keep in our heads.

awead (approved):

PR #640, opened 2020-10-27
Render markdown partials with custom markdown handler

### LINKED ISSUE #167: Documentation and Policy Pages
We need a set of pages within the ScholarSphere site that tell users what our policies are for backups, storage, etc. 

## What We Need

* names of pages, i.e. "About"
* what the link should be, i.e. `/about` 

## WDLL

* markdown pages in `/app/views/pages`
* corresponding actions in the pages controller, ex. `/about`
* updates will happen via PR and releasing

awead:
This registers an ActionView template handler for rendering partials written in mardown. Redcarpet is used to convert the markdown into HTML, and the rest relies on Rails' internal templating architecture.

The MarkdownController defines a single action to render the page. Specific pages are defined in routes, linking the name of the route to the corrsponding markdown file.

Fixes #167


awead:
@srerickson The content you want for these pages would go into `app/views/pages/_about.md`. We'll also need to know now many other pages you want and what you want their main titles and urls to be.

rschenk (changes requested):
I think this is solid, but I think we can make it more easier since we know are going to have multiple pages with markdown.

What do you think about this idea:
* A MarkdownPages controller
* Convert what you've got in `about.html` into a [nested layout](https://guides.rubyonrails.org/layouts_and_rendering.html#using-nested-layouts) and set that as the default layout for the MarkdownPagesController
* Automatically pull the title in ☝️ the nested layout from en.yml using `action_name`
* `_about.md` becomes `about.md` which uses the new nested layout for rendering
* Subsequent Markdown pages are as simple as adding `def new_page; end` and `new_page.md` to that controller (plus a route) 

Alternatively, we could scrap the MarkdownPagesController from above, but keep everything else including the nested layout. Inside our PagesController we could flip between the frontend layout and the markdown nested layout depending on the method. 

Whatcha think?

awead:
@rschenk Excellent. I had had a similar idea, but I didn't know about the nested layouts option and was hung up on duplicating the entire layout just for this one thing. I was also wondering if there was a way to do this with routing and just _one_ controller action since they'd all be the same:

``` ruby
get '/about', 'markdown#render', page: 'about'
```

And `page` is effectively a parameter passed to the controller. Maybe that's too weird.

rschenk:
@awead Whoa that's cool. Let's try it

rschenk (dismissed):

rschenk (approved):

PR #643, opened 2020-10-28
Migrate statistics via ImportStatisticsJob

### LINKED ISSUE #641: Make statistics migrations async
Migrating view and download statistics is a time-intensive task. We should make updates to each statistic its own job and run them async.

awead:
Migrating statistics from Scholarsphere 3 is a time-intensive task and will need to be run repeatedly. This extracts our current rake task into multiple jobs which can be run asynchronously via Sidekiq.

Fixes #641


rschenk (approved):
Looks good to me!

Issue #277, opened 2020-04-29
UI component for modifying ACLs for Works & Collections


srerickson:
We need a way to modify ACLs for individual works and collections. It probably makes sense to include this component in the edit page for works and collections (#207, #254).

A screenshot from SS3's UI component for this is attached. 

<img width="400" alt="Screen Shot 2020-04-29 at 10 42 16 AM" src="https://user-images.githubusercontent.com/876323/80610186-0a463980-8a07-11ea-84ad-10de89102498.png">

PR #650, opened 2020-10-29
Define /authorities/search/users endpoint

awead:
Uses Penn State's identity service to search for users that can be granted edit access.

Related to #277

### DIFF
--- a/lib/qa/authorities/users.rb
+++ b/lib/qa/authorities/users.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+# @abstract This duplicates Qa::Authorities::Persons with the only difference being only Penn State users are returned
+# in a search. An alternative implementation would be to define this as a subauthority of Persons, but this just seems
+# easier.
+module Qa
+  module Authorities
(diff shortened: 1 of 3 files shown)


lib/qa/authorities/users.rb
```diff
@@ -0,0 +1,17 @@
...
+# in a search. An alternative implementation would be to define this as a subauthority of Persons, but this just seems
+# easier.
+module Qa
+  module Authorities
+    class Users < Persons
+      private
+
+        # @return Array<PennState::SearchService::Person>
+        def persons
+          identities
```
rschenk, line 13: This is some kinda magical mystery code, but everything looks good to me.

rschenk (approved):

lib/qa/authorities/users.rb
awead: @rschenk it's in the super class. Persons is a combination of local actors and Penn State people (which is the `identities` method). But here, we just want the Penn Staters... so, identities. Not very intuitive.

Issue #338, opened 2020-05-27
Work Edit Actions UI Components


srerickson:
We need a UI component providing users with the ability to initiate a variety of actions related to the management of their work. This component will likely appear on the work page for works the user can edit. (replaces #207).

## Edit actions for a work include:
- [x] Change Access Controls #277 
- [x] Create a DOI for the work if none exists (#150)
- [x] Changing visibility (toggle b/w Public & Penn State) (#601)
- [x] Setting Embargo (#621)
~~List versions, w/ actions for requesting withdrawal and deleting (if the draft)~~ #696 

*updated by seth 10/23*

srerickson:
Instead of packing all this functionality into the drop down and modal UIs, it may be easier and more user friendly for the gear icon to be a link ("gear Work Settings") to a separate page where you can do all this stuff. 

I'm imagining something like this:
![Screenshot from 2020-10-23 14-22-03](https://user-images.githubusercontent.com/876323/97040147-8f845400-153b-11eb-9564-363e03021912.png)

PR #654, opened 2020-10-29
Work settings page

### LINKED ISSUE #601: Changing visibility (toggle b/w Public & Penn State)
Users should be able to change the visibility of a work without creating a new version. This can happen in the resources page via a IU element.

Related to #277

rschenk:
Create a separate page for modifying work-specific settings. Right now it only has things for DOI and Visibility, but it will soon gain much more.

It has a sticky sidebar for navigating within this page, along with scrollspy for highlighting the correct sidebar navigation item as the page scrolls. This is overkill now, but it's quickly going to become useful.

I also cleaned out a bunch of dead code left over from the old dashboard

Closes #601
Related to #338


app/controllers/dois_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
     # I don't think we need to do anything special here, just skip over the call
     # to MintDoiAsync and redirect back where we came from?
   ensure
-    redirect_to resource_path(params[:resource_id])
+    redirect_back fallback_location: resource_path(params[:resource_id])
```
rschenk, line 17: When a DOI is minted, return to wherever HTTP_REFERRER header is set to. Barring that, just go back to the resource page

app/decorators/resource_decorator.rb
```diff
@@ -27,6 +27,8 @@ def display_published_date
   end
 
   def display_doi
-    MintableDoiComponent.new(resource: resource_with_doi)
+    return if resource_with_doi.doi.blank?
+
+    MintingStatusDoiComponent.new(resource: resource_with_doi)
```
rschenk, line 32: No longer mint DOIs from the metadata table. This is why I made all these `-DoiComponent`s compose each other. So changing this is really simple

app/views/dashboard/work_form/publish/_fields.html.erb
```diff
@@ -9,21 +9,7 @@
...
-          <%= work_form.label :visibility, lbl, value: val, class: 'form-check-label' do %>
-            <%= lbl %>
-            <% if I18n.exists? "helpers.hint.permissions.visibility.#{val}" %>
-              <span class="label__help"><%= t "helpers.hint.permissions.visibility.#{val}" %></span>
-            <% end %>
-          <% end %>
-        </div>
-      <% end %>
-    </div>
+    <%= render 'visibility_field', form: work_form %>
```
rschenk, line 12: Extract the visibility options into a partial so we can show them on the publish form and on the work settings page

app/views/dashboard/works/edit.html.erb
```diff
@@ -0,0 +1,58 @@
+<%= content_for :navbar_items do %>
+  <li class="nav-item">
+    <%= link_to(t('.back'),
+                resource_path(@work.uuid),
+                class: 'btn btn-outline-light btn--squish mr-lg-2 my-1') %>
+  </li>
+<% end %>
+
+<%= content_for :body_attributes, 'data-spy="scroll" data-target="#navbar-settings" data-offset="10" style="position: relative"'.html_safe %>
```
rschenk, line 9: This is needed for scrollspy. We have to add these attributes to the `<body>` tag

app/views/dashboard/works/edit.html.erb
```diff
@@ -0,0 +1,58 @@
...
+            <%= visibility_form.submit t('.visibility.submit_button'), class: 'btn btn-primary' %>
+          </div>
+        <% end %>
+
+        <div class="keyline keyline--left">
+          <h2 id="<%= t('.doi.heading').parameterize %>" class="h4"><%= t('.doi.heading') %></h2>
+        </div>
+
+        <p><%= t('.doi.explanation') %></p>
+        <%= render MintableDoiComponent.new(resource: @work) %>
```
rschenk, line 54: Booyah, can still use this bad boy!

app/views/dashboard/works/edit.html.erb
```diff
@@ -0,0 +1,58 @@
...
+        </div>
+
+        <p><%= t('.visibility.explanation') %></p>
+
+        <%- visibility_form_id = "edit_visibility_#{dom_id @work}" %>
+        <%= form_for [:dashboard, @work], remote: false, html: { id: visibility_form_id, class: 'edit-work-visibility' } do |visibility_form| %>
+          <%= hidden_field_tag 'form_id', visibility_form_id %>
+          <% if params['form_id'] == visibility_form_id %>
+            <%= render FormErrorMessageComponent.new(form: visibility_form) %>
+          <% end %>
```
rschenk, line 40: ☝️ This is a thing I'm playing around with. We are going to have _multiple forms_ on this page. Many of them will all submit to the same endpoint, `Dash::WorksController#update`. So if there is an error in the form submission and this page is re-rendered, how do you know which form to show the errors on? They all use `@work`, they would _all_ simultaneously show the same error message, even if it doesn't affect them.

One idea I had was each form could have a hidden field called `form_id`, with a unique value for each form. If the form was submitted, had validation errors, and this page got-rendered, each form would "know" whether or not they had made the original submission, and therefore if they should show the errors or not. It's a work in progress, but it does seem to be the simplest and most foolproof solution I've thought of.

awead (approved):
Solid. I've pushed this up to the preview/migration branch for retro today, but didn't merge it in case there's anything else you're still working on.

Issue #625, opened 2020-10-22
Remove "scholarsphere" layout


awead:
The original Blacklight-based "scholarpshere" layout and UI elements should be removed. These have been replace by our "frontend" code.

## WDLL

* under  app/views/layouts, remove _user_menu.html.erb, blacklight, application, _head?
* remove app/javascripts/scholarpshere and any other references in js
* remove associated tests

Related to #492 and #457

PR #665, opened 2020-10-31
Preview/dast

whereismyjetpack:
OIS used a DAST (dynamic application security testing) tool against scholarsphere, and we discovered that, by default, rails does not use secure, httponly cookies, and the matomo cookie isn't secure by default. this PR addresses those issues.

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -19,6 +19,7 @@
     var _paq = window._paq || [];
     /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
     _paq.push(['trackPageView']);
+    _paq.push(['setSecureCookie', true]);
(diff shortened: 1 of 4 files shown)


whereismyjetpack:
I wasn't sure if the session_store info would be better suited for production.rb or in an intializer, every example i found showed it as an initializer

awead (changes requested):
This looks good. We only need one Matomo file. The head partial I think can be removed. The session store config can probably go in production.rb?

whereismyjetpack:
i moved the cookie_store config into environments, that feels right.. i am going to address the base partial as a different PR, if we don't need it we can remove it entirely?

awead:
@whereismyjetpack the other partial needs to be removed, but that can be part of #625

awead (approved):

PR #672, opened 2020-11-03
Preview/docs

srerickson:
Work on markdown pages (about, policies, agreement, etc.). I'm sure there are better ways to arrange things between routes, `en.yml`, and the filenames, but this seems to work well enough. Note, I needed to add an additional option to the redcarpet config to get TOCs to work like I want.


app/views/markdown/agreement_1_0.md
```diff
@@ -0,0 +1,31 @@
+# ScholarSphere Deposit Agreement
+
+<p class="alert alert-warning">
+ This agreement (v1.0) has been superseded by <a href="agreement-2.0">version 2.0</a>.
```
awead, line 4: Did you verify the links work?

app/views/markdown/about.md
```diff
@@ -1,3 +1,13 @@
...
 
+ScholarSphere supports Penn State’s commitment to open access and open science. Researchers at Penn State can use ScholarSphere to satisfy open access and data availability requirements (e.g., FAIR) from funding agencies and publishers. 
+
+Penn State's [open access policy](https://openaccess.psu.edu/)  (AC02: Open Access to Scholarly Articles) requires that university researchers provide the [post-print version](http://psu.libanswers.com/faq/273465) of their published scholarly articles to the University Libraries to be made available to the public (unless the authors have obtained a waiver). Depositing in ScholarSphere satisfies this requirement. 
+
+## Additional Resources:
+
+- [Policies for Deposit, Access, Curation, and Preservation](policies)
+- [Deposit Agreement](agreement)
+- [GitHub Repository](https://github.com/psu-stewardship/scholarsphere-4)
```
awead, line 13: At some point, very soon after release, this will change to just https://github.com/psu-stewardship/scholarsphere

app/views/markdown/help.md
```diff
@@ -0,0 +1 @@
+help page
```
awead, line 1: Did you want anything here?

awead (changes requested):
Looks good. I just had a couple of questions, and I assume you checked that all the links work.

The current configured deposit agreement is 1.0, so migrated works will be under the 1.0 version. I'm assuming the move to s 2.0 agreement will involve another process later?

app/views/markdown/help.md
srerickson: Yes, but I'd like to do it in a future PR.

app/views/markdown/agreement_1_0.md
srerickson: yep: https://scholarsphere-docs.dsrd.libraries.psu.edu/policies-1.0

app/views/markdown/about.md
srerickson: thanks, I'll update this in a future revision

srerickson:
My thought is that *migrated* works should stay with the 1.0 agreement and any *new* works created after launch should use the 2.0 agreement.

app/views/markdown/help.md
awead: @srerickson Will that PR be before release? If so, no problem. However, if it's going to be after 4.0, we should remove the page and the link from top bar.

awead (approved):

PR #678, opened 2020-11-06
Manage required group agents with seeds

### LINKED ISSUE #677: Seed group data
There's a bug with multi-worker setups that  could result in multiple public and authorized groups. This can cause incorrect access controls because a resource would be linked to the wrong "public" group.

To fix this, we should seed the database with the required record beforehand to ensure the records are there even before the application is deployed.

awead:
Instead of relying on the application to create required resources like public and authorized group agents, we'll seed them into the database prior to testing and deployment. To further avoid any unintended consequences of multiple agents with the same name, we'll ensure all group names are unique.

Fixes #677


rschenk (approved):
Looks good to me!

PR #679, opened 2020-11-09
Return 404 for Pundit authorization errors

### LINKED ISSUE #661: Pundit::NotAuthorizedError should return 404
If a user is unable to do something due to a policy restriction, like viewing a draft work that they don't own, Pundit raises a `Pundit::NotAuthorizedError` which the application returns as a 500 Server Error. These should really be just a 404 Not Found. While _technically_ it should be 401 or 403 (Unauthorized or Forbidden), the "not found" is a more secure response because it does not expose the fact that the resource actually does exist, even though you can't see it.

awead:
Taps into ActionDispatch::ExceptionWrapper#rescue_responses and adds an additional directive to return :not_found responses for Pundit authorization errors. Previously, these were returning the default 500 error which was misleading.

Fixes #661

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,6 +8,8 @@ class ApplicationController < ActionController::Base
   # Authorization
   include Pundit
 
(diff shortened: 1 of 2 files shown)


rschenk (approved):
Nice, that is clean and simple

PR #681, opened 2020-11-09
Update search results display

### LINKED ISSUE #638: 🐞Sorting order is still not correct in Dashboard
We had an issue to ensure that recently updated works were shown at the top of the dashboard #555. This was solved by PR #581, except that we still have the problem if the dashboard contains more than 1 page of works.

What happened was, #581 solved the issue by taking the Solr result set as given, then applying an order clause to the method that pulls the corresponding records out of the Postgres database. The problem with this approach is that pagination is done by Solr, well before this is ever handed off to that ActiveRecord code. So #581 correctly sorts the records _within each page of results_ but does not sort them properly globally. To fix this, we'll need to do the sorting in the Solr query, i.e. before the pagination is computed.

### LINKED ISSUE #647: Remove "History" link in Search/Dashboard List Items
Remove the "History" link that appears for each item:

Screen Shot 2020-10-29 at 12 34 09 PM

(can also remove the "new version" link)

### LINKED ISSUE #667: Fix up search results
Search results should look the same in the dashboard as well as the public results.

## WDLL

* use deposited date
* remove any links to create, edit, or delete
* remove history link (#647)
* remove work type
* add first three creators

awead:
This changes the both the public and dashboard search so that all results are displayed the same way. The deposited date for works and collections is used, there are no edit or delete links used, history and work type is no longed displayed, and instead the first three creators are shown.

Fixes #667
Fixes #647 
Fixes #638 

This is how search results would look in either the dashboard or the main search:

Screen Shot 2020-11-09 at 3 28 41 PM


awead:
@rschenk If this looks good, there's a second commit that I'd like to push to this that removes the database-backed display of dashboard results. Because we don't need to display the edit, create, or delete buttons in the dashboard results, we don't need the postgres records to determine the permissions. Therefore, the Solr document alone are enough. This would take care of our sorting problems.

rschenk (approved):

PR #684, opened 2020-11-10
Re-query Actors using Penn State's identity results

### LINKED ISSUE #659: Inconsistent search results for creators
Depending on what the user enters when looking for creators to add to a work, they may or may not get the appropriate actor record.

## Scenario

A user wants to add Jennifer Arnold as a creator on their work. This person is at Penn State and has also deposited content in Scholarsphere. Therefore, they have both an existing record in Penn State's identity service, and a record in Scholarsphere.

### The Wrong Search

If the user enters "Jennifer Arnold" into the search bar, the user's Penn State record is returned. This is incorrect because we need the search to return the local Actor record which was created previously.

    /authorities/search/persons?q='Jennifer Arnold'

### The Right Search

If the user enters "arnold" into the search bar, the second hit is the right one.

    /authorities/search/persons?q=arnold

## The Problem

Because we're search two different sources (local actors and Penn State identities), no one search will return the same set of users. If there are duplicates, we filter out the Penn State records based on access account and only return the actor results. In the above case, the "Jennifer Arnold" search only returned the Penn State identity, but _not_ the actor record. The "arnold" search returns the actor record, and if the Penn State identity is there as well, it is removed.

## The Solution

It's unclear at this point. We'll probably need to do some kind of "double query" where we re-query the actor records using a list of access ids that are returned from the initial search. This would solve the above problem because we could find the local actor record based off of the results from the "Jennifer Arnold" search. The only problem is that it requires an additional database query for each search.

awead:
This avoids a problem whereby a user's query will return the correct results from Penn State's service, but not our local actors. The fix is to just re-query the actor's table using the results from the identity service. This way we will always have the preferred local records.

Fixes #659

### DIFF
--- a/lib/qa/authorities/persons.rb
+++ b/lib/qa/authorities/persons.rb
@@ -30,7 +30,9 @@ module Qa
         end
 
         def creators
-          @creators ||= Actor.where('surname ILIKE :q OR given_name ILIKE :q OR psu_id ILIKE :q', q: "%#{term}%")
+          @creators ||= Actor
+            .where('surname ILIKE :q OR given_name ILIKE :q OR psu_id ILIKE :q', q: "%#{term}%")
+            .or(Actor.where(psu_id: identities.map(&:user_id)))
         end
 
         # @note using Set enables faster searching for a given id instead of iterating over the entire array.
--- a/spec/lib/qa/authorities/persons_spec.rb
+++ b/spec/lib/qa/authorities/persons_spec.rb
@@ -95,8 +95,34 @@ RSpec.describe Qa::Authorities::Persons, type: :authority do
       it { is_expected.to include(formatted_result) }
     end
 
+    context "when Penn State's identity service contains existing Scholarsphere Actors" do
+      let!(:creator) { create(:actor) }
+
+      let(:formatted_result) do
+        {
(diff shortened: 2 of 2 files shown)


lib/qa/authorities/persons.rb
```diff
@@ -30,7 +30,10 @@ def persons
         end
 
         def creators
-          @creators ||= Actor.where('surname ILIKE :q OR given_name ILIKE :q OR psu_id ILIKE :q', q: "%#{term}%")
+          @creators ||= begin
+                          Actor.where('surname ILIKE :q OR given_name ILIKE :q OR psu_id ILIKE :q', q: "%#{term}%") |
+                            Actor.where(psu_id: identities.map(&:user_id))
```
rschenk: This seems expensive to do with two queries + ruby. Can we not do this all in a single SQL query? 

```ruby
@creators ||= Actor
  .where('surname ILIKE :q OR given_name ILIKE :q OR psu_id ILIKE :q', q: "%#{term}%")
  .or(Actor.where(psu_id: identities.map(&:user_id)))
```

rschenk (changes requested):

rschenk (approved):

PR #690, opened 2020-11-10
Embargo a work from Work Settings page

### LINKED ISSUE #621: A way for users to set an embargo on a work
We need to have a way for users to set an embargo on a work. Probably the best place to put this is with other "visibility" settings. For example, there would be three visibility options on the deposit form. If you select the embargo option, you need to add a date when the embargo lifts.

At one point in time, I suggested that embargoes might only be created by admins, at user request. This won't work, unfortunately. 

## WDLL

* validate date with YYYY-MM-DD
* unchecking the checkbox removes the embargo
* embargoes are lifted at 12:01 AM UTC on the given date

rschenk:
Closes #621 and sets up an arguably dubious pattern of using form objects on the work settings page

### DIFF
--- a/app/forms/embargo_form.rb
+++ b/app/forms/embargo_form.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+class EmbargoForm
+  include ActiveModel::Model
+
+  attr_reader :work
+
+  attr_writer :embargoed_until
+  attr_accessor :remove
+
+  validate :embargoed_until_is_valid_date,
+           unless: :remove?
+
+  def initialize(work:, params:)
+    @work = work
+    super(params)
+  end
+
+  def embargoed_until
+    work_embargoed_until = if work.embargoed_until.respond_to?(:strftime)
+                             work.embargoed_until.strftime('%Y-%m-%d')
+                           end
+    @embargoed_until || work_embargoed_until
+  end
+
+  def remove?
+    !!ActiveModel::Type::Boolean.new.cast(remove)
+  end
+
+  def save
+    return false unless valid?
+
+    work.embargoed_until = if remove?
+                             nil
+                           elsif embargoed_until.present?
+                             Time.zone.parse(embargoed_until).beginning_of_day
+                           end
+    work.save
+  end
+
+  private
+
+    def embargoed_until_is_valid_date
+      return if embargoed_until.blank?
+
+      valid_format = embargoed_until.to_s.match(/^\d{4}-\d{2}-\d{2}$/)
+
+      valid_date = begin
+                     Date.parse(embargoed_until.to_s)
+                     true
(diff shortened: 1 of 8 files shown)


app/forms/embargo_form.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      if !valid_format
+        errors.add(:embargoed_until, :format)
+        return
+      end
+
+      if !valid_date
+        errors.add(:embargoed_until, :not_a_date)
+        nil
+      end
```
awead, line 55: What about:
``` ruby
def embargoed_until_is_valid_date
  return if embargoed_until.blank?
  
  unless embargoed_until.to_s.match(/^\d{4}-\d{2}-\d{2}$/)
    errors.add(:embargoed_until, :format)
    return
  end

  begin
    Date.parse(embargoed_until.to_s)
  rescue ArgumentError, TypeError
    errors.add(:embargoed_until, :not_a_date)
  end
end
```
I'm not sure if it's any clearer. I was thinking about suggesting _two_ new methods, one for format and one for date, and then `embargoed_until_is_valid_date` calls the two and returns early if the format's wrong.

It's not clear to me what the method is supposed to return, or if that's even important, but I think the only this it really needs to do is add errors or not.

app/forms/embargo_form.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def initialize(work:, params:)
+    @work = work
+    super(params)
+  end
+
+  def embargoed_until
+    work_embargoed_until = if work.embargoed_until.respond_to?(:strftime)
+                             work.embargoed_until.strftime('%Y-%m-%d')
+                           end
```
awead, line 22: does this work? `work_embargoed_until = work.embargoed_until.try(:strftime, '%Y-%m-%d')`

awead (dismissed):
I think this is good. I left some suggestions if you want to entertain them. Yes, it's dubious, but I don't think we really have enough examples _yet_ to formulate alternative patterns, so this seems as good as any. With more examples, we could refactor into something better.

app/forms/embargo_form.rb
rschenk: It works if work.embargoed_until is nil or a date, but it does not work if for some reason it was a string.

awead: @rschenk ok, it's fine as-is, I was just curious.

app/forms/embargo_form.rb
rschenk: Yeah, this method doesn't need to return anything, just call `errors.add` if necessary. On the one hand, I sort of like to "set up all the stuff" in one block at the top of a method, then perform actions on it in all in the second half of a method, but your version is quite a bit cleaner. Maybe I'll go with that

awead (approved):

PR #693, opened 2020-11-11
Preview/help

srerickson:
just a help page


app/views/markdown/help.md
```diff
@@ -1 +1,81 @@
...
+*   No, the University provides this service at no cost to current Penn State faculty, students, and staff and to emeriti faculty.
+
+#### What can I deposit to ScholarSphere? 
+
+*   All types of scholarly materials, including publications, instructional materials, creative works, and research data can be deposited to ScholarSphere To be eligible for deposit in ScholarSphere, works must:
+1. Be produced or sponsored by at least one member of the Penn State community;
+2. Be of scholarly import, including publications, instructional materials, creative works, and research data produced in support of Penn State's teaching, learning, and research mission; and
+3. Meet the requirements of[ Penn State Policy AD69 Accessibility of Electronic and Information Technology](https://policy.psu.edu/policies/ad69).
+*   Information with a Sensitive Information Classification of Moderate (Level 2) or higher according to[ Penn State Policy AD95 Information Assurance and IT Security](https://policy.psu.edu/policies/ad95) is not permitted in ScholarSphere.
+*   For more, see[ ScholarSphere’s Content and Deposit Policy](https://scholarsphere.psu.edu/about).
```
rschenk: One thing I might do, if this is possible in markdown, is any internal links to Scholarsphere routes, I might make those as relative paths rather than full URLs.

i.e. `/about` instead of `https://scholarsphere.psu.edu/about`

That way these links will still work when deployed to other hosts, such as beta/staging/preview/etc. 

As I mentioned, I do not know if Markdown actually supports this, I think it would be worth trying. If that's not possible, feel free to ignore this

rschenk (approved):
Looks good to me

PR #698, opened 2020-11-12
WIP: Resolve CVE 2015 9284 with POST to azure

awead:
Resolves the security issue with CVE 2015 9284 by allowing only POST methods to the Azure omni auth service.

Note: This works, but the JS-based tests fail.


awead:
It looks like there's a forthcoming 2.0 release that will address this. https://github.com/omniauth/omniauth/pull/1010

PR #701, opened 2020-11-12
Twitter and Google Scholar metadata meta tags

### LINKED ISSUE #410: Meta Tags on Resource Pages
We should add structured metadata to pages, particularly work pages to improve SEO and support citation importing (e.g. the Zotero import tool).

Mandee provided this resource: https://developers.google.com/search/docs/guides/search-gallery

rschenk:
Closes #410 

The only WTF here is, I had to pass down the full URL from the ResourcesController#show view, down into the partials. This was necessary because, when looking at the "latest published version" we want to ensure we are rendering the URL for the parent Work and not the WorkVersion. 

I also added page titles to our sad, neglected Collection

### DIFF
--- a/app/views/layouts/frontend.html.erb
+++ b/app/views/layouts/frontend.html.erb
@@ -7,6 +7,7 @@
 
     <!-- Internet Explorer use the highest version available -->
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <%= yield(:meta_tags) if content_for? :meta_tags %>
 
     <title><%= render_page_title %></title>
     <%= javascript_pack_tag 'frontend' %>
--- a/app/views/resources/_collection.html.erb
+++ b/app/views/resources/_collection.html.erb
@@ -3,7 +3,7 @@
 <% end %>
 
 <%- content_for :meta_tags do %>
-  <%= render 'meta_tags', resource: collection, current_url: current_url %>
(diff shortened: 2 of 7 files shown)


awead (changes requested):
Instead of passing `current_url` all the way from the show page, why not just pas a uuid into meta_tags?

For the collection partial, it'd look like:

``` ruby
<%= render 'meta_tags', resource: collection, uuid: collection.uuid %>
```

And for work versions it'd be:

``` ruby
<%= render 'meta_tags', resource: work_version, uuid: work_version.work.uuid %>
```

Then then meta_tag just renders the resource url with the provided uuid.

rschenk:
@awead That would work, _except_ if you are viewing the resource page for a specific work-version (e.g. an older one) , in which case you're just as hosed. So regardless of whether you pass through the full URL or just the UUID, you still need to pass it in from the main `resoruces/show` view.

awead:
@rschenk yeah, I realized that too... from `_work_version`:

``` ruby
  <%= render 'meta_tags', resource: work_version, uuid: (uuid || work_version.uuid) %>
```

If you're calling the work_version partial directly, `uuid` would be nil since it would have been passed in from the work partial.

Alternatively, what about just using `params[:id]`?

awead (approved):
👊

PR #705, opened 2020-11-12
Guest users should not apply to audits

### LINKED ISSUE #704: Redirects to logins are returning 500
From a "clean" Firefox session, if I visit `/dashboard`, I get **Server Error**. Looking at DataDog, I'm seeing:

`URI::GID::MissingModelIdError: Unable to create a Global ID for User without a model id.`

This is also reproducible locally. It also may be tangentially related to #692. Devise wants to _redirect_ to a login page, but that may ultimately have to be a POST method, which you can't do according to the HTTP spec. We may ultimately have to do something like redirect to the home page and have the user click on the login link.

awead:
There is always a current user, but not all current users can be audited. This avoids attempting to make a guest user a paper trail user.

Fixes #704

### DIFF
--- a/app/controllers/concerns/with_auditing.rb
+++ b/app/controllers/concerns/with_auditing.rb
@@ -12,7 +12,7 @@ module WithAuditing
   # @note Overrides PaperTrail::Rails::Controller to return a global id because our current users are either a User or
   # an ExternalApp.
   def user_for_paper_trail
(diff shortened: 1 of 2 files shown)


rschenk (approved):

awead:
@rschenk thanks, 🐕

PR #711, opened 2020-11-13
Required elements for publishing

### LINKED ISSUE #700: "License is required" error on Review and Publish
When you're creating a new work and get to the Review and Publish form for the first time, you will always see the error, "License is required to publish the work", which is confusing and misleading because the license can only be set on the Review and Publish form. The error should only appear if you try to publish the version without setting the license.

### LINKED ISSUE #707: Review and publish step should say "Access" and not "Visibility"

awead:
License and deposit agreement are now marked required, which will inform users using HTML5 feature and will not longer display the error messages if they are navigating to the page for the first time.

We have also updated the visibility label to read _Access_, making it consistent throughout the application.

Fixes #700 
Fixes #707

### DIFF
--- a/app/controllers/dashboard/work_form/publish_controller.rb
+++ b/app/controllers/dashboard/work_form/publish_controller.rb
@@ -8,9 +8,7 @@ module Dashboard
           .includes(file_version_memberships: [:file_resource])
           .find(params[:work_version_id])
         authorize(@work_version)
-
-        @work_version.publish
-        @work_version.validate
+        prevalidate
       end
 
       def update
@@ -50,6 +48,15 @@ module Dashboard
       end
 
       private
+
+        # @note Validate the work like we're going to publish it, so we can inform the user ahead of time before they
+        # actually attempt it. However, we want to be nice and not yell at them for something they _haven't_ seen yet
+        # like rights.
+        def prevalidate
+          @work_version.publish
+          @work_version.validate
+          @work_version.errors.delete(:rights)
+        end
 
         def publish_work?
           !save_and_exit?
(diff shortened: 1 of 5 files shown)


awead:
@rschenk I can't remember if there was a specific reason we were doing a `publish && validate` at the outset. Maybe it was to show the errors right away? But now we don't want to do that...

rschenk:
@awead @srerickson the point of the `publish && validate` at the beginning was to preemptively show any publication-level validation errors, _but at the time that was written, there was nothing on this form that the user had not already seen before (that had validations)._ I agree that #700 is a major UX issue, but I also don't think we should necessarily throw the baby out with the bath water. 

For example, `description` and `publication_date` are not required to save a draft but are required to publish, so you would see this when the form was first loaded.

Untitled

Another way to accomplish this would be below. It's a little more brittle but I think the UX is better:

```ruby
      def edit
        @work_version = WorkVersion
          .includes(file_version_memberships: [:file_resource])
          .find(params[:work_version_id])
        authorize(@work_version)

        @work_version.publish
        @work_version.validate

        # Remove any errors messages from fields that the user hasn't seen before
        # such as the "Publishing Details" section
        @work_version.errors.delete(:rights)
      end
```

awead:
@rschenk Wouldn't the above solution remove an actual error? If they hit publish, but hadn't selected rights, it would redirect back to edit and remove the error message?

rschenk:
@awead Not quite. If they hit publish but hadn't selected rights, it will _render_ the edit form, but it will not _redirect_ to the edit action, therefore none of the code in the `#edit` method will be executed even though the view is rendered. Kind of confusing.

rschenk (approved):
😎

PR #712, opened 2020-11-15
Use a null indexer for selective indexing of work versions

awead:
In the publish step of the work form, we save the record twice to ensure any changes are persisted even if publishing fails. This was creating an indexing problem because, presumably owing to the quick succession in which the record was being saved and indexed each time, the *second* save process wasn't getting committed to the index. As a result, a work version that was published, wasn't getting correctly indexed as such in Solr.

The fix is to only index on the second call to #save, and not to index on the first call.


rschenk (approved):
Good catch and slick solution!

PR #719, opened 2020-11-17
Disallow Works to be shown if they are not published

### LINKED ISSUE #717: 500 Error for draft work resources
When requesting the uuid for a _Work_ that is still in draft, i.e. it has only one version and that version is in draft status, we're getting a 500 error:

ActionView::Template::Error: undefined local variable or method `model_name' for #<NullWorkVersion:0x00007fc0943c84d8>

Similar to #714

rschenk:
When rendering a Work, the resource page requires/assumes that there will be at least one published version. This was causing trouble when a user _with edit access_ requested the resource page for a draft work, which caused a 500. This commit tweaks the policy to raise a pundit error when requesting the resource page of draft work.

Note that #717 was only ever an issue when the user requesting said work had edit access to it.

Closes #717


app/policies/work_policy.rb
```diff
@@ -8,7 +8,7 @@ def limit
   end
 
   def show?
-    (record.discover_access?(user) && published?) || edit?
+    published? && (record.discover_access?(user) || edit?)
```
awead, line 11: @rschenk If I'm reading this right, nobody (not even admins) can view unpublished _works_. Work _versions_ that are in draft state should still be fine, but I'm wondering if there's a need to redirect to the draft work version if an admin, or someone else, attempts to view the unpublished work uuid.

I may be over-thinking this, because in most cases, we aren't going to be passing around uuids for unpublished works, just for work versions since that's the url in the dashboard.

awead (changes requested):

app/policies/work_policy.rb
rschenk: You're reading that right. I'm not sure if this is the way we want to go, but it does seem that way. The resource page for a _Work_ requires that it have at least one published version. If there is not a published version it will panic and 500. 

We could either make the resource page gracefully handle unpublished works, or we disallow them entirely (what this does).

awead (approved):

PR #762, opened 2020-12-08
i#748 license display

### LINKED ISSUE #748: Fix license displayed on resource page
On resource page, we're showing the raw URL value of the license rather than the human readable version. We should show the human readable version that is linked to the appropriate rights statement. 

In addition the URL used for "rights reserved" needs to be changed to: https://rightsstatements.org/page/InC/1.0/?language=en

srerickson:
Displays the rights as a link to it's corresponding URI.

Fixes #748

### DIFF
--- a/app/components/display_rights_component.rb
+++ b/app/components/display_rights_component.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class DisplayRightsComponent < ApplicationComponent
+  attr_reader :id,
+              :label
+
+  def initialize(id:)
+    @id = id
+    @label = WorkVersion::Licenses.label(id)
+  end
+
+  def render?
+    label.present?
+  end
+end
--- a/config/authorities/licenses.yml
+++ b/config/authorities/licenses.yml
@@ -41,9 +41,12 @@ terms:
   - id: http://creativecommons.org/publicdomain/zero/1.0/
     term: CC0 1.0 Universal
     active: true
-  - id: http://www.europeana.eu/portal/rights/rr-r.html
+  - id: https://rightsstatements.org/page/InC/1.0/
     term: All rights reserved
     active: true
+  - id: http://www.europeana.eu/portal/rights/rr-r.html
+    term: All rights reserved
+    active: false
   - id: http://www.apache.org/licenses/LICENSE-2.0
     term: Apache 2.0
     active: true
--- a/app/components/display_rights_component.html.erb
+++ b/app/components/display_rights_component.html.erb
@@ -1,0 +1,3 @@
+<span class="display-rights">
+    <%= link_to label, id, target: :_blank, rel: 'noopener' %>
+</span>
--- a/app/components/work_version_metadata_component.rb
+++ b/app/components/work_version_metadata_component.rb
@@ -13,7 +13,7 @@ class WorkVersionMetadataComponent < ApplicationComponent
     :creator_aliases,
     :version_number,
     :keyword,
-    :rights,
+    :display_rights,
     :display_work_type,
     :contributor,
     :publisher,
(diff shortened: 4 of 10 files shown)


config/authorities/licenses.yml
```diff
@@ -41,9 +41,12 @@ terms:
   - id: http://creativecommons.org/publicdomain/zero/1.0/
     term: CC0 1.0 Universal
     active: true
-  - id: http://www.europeana.eu/portal/rights/rr-r.html
+  - id: https://rightsstatements.org/page/InC/1.0/?language=en
```
awead: @srerickson I'd remove the `/?language=en` at the end of the url.

app/components/display_rights_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class DisplayRightsComponent < ApplicationComponent
+  attr_reader :label,
+              :id
+
+  def initialize(id:)
+    @id = id
+    @label = WorkVersion::Licenses.label(id) if id.present?
```
awead: @srerickson will there even be a nil id? If so, we could use the `render?` method. This is a built-in feature of ApplicationComponent. Basically, if we were to initialize the component without an id, or a `nil` value, it won't display anything:

``` ruby
def render?
  id.present?
end
```

With the way it's written above, if the id were nil, the span html would still render and you'd get some kind of a broken link, I think.

app/decorators/resource_decorator.rb
```diff
@@ -32,6 +32,12 @@ def display_doi
     MintingStatusDoiComponent.new(resource: resource_with_doi)
   end
 
+  def display_rights
+    return unless respond_to?(:rights)
+
+    DisplayRightsComponent.new(id: rights)
+  end
```
awead: @srerickson I'm curious what errors you were seeing when this was on the WorkVersionDecorator. My guess is that if this is defined on the WorkVersionDecorator, you won't need the return statement, but, I bet you'll need to account for nil ids. That may be why you were seeing errors. See my other comment about the `render?` method. That may solve the problem.

awead (changes requested):
@srerickson Looks good! I think you're replicating the right pattern here. Just a few things to tidy up and we'll be able to push this out.

config/authorities/licenses.yml
srerickson: @awead, it's actually necessary: if you remove the URL param, the link won't resolve.

srerickson: actually this works: https://rightsstatements.org/page/InC/1.0/
but not this: https://rightsstatements.org/page/InC/1.0

bad.

app/decorators/resource_decorator.rb
srerickson: @awead here's an example of the error that happens if I move it to WorkVersionDecorator. (This is with the render? method added).

```rb
  1) WorkVersionMetadataComponent rendering renders a string with label
     Failure/Error: value = format_value(work_version.send(attr))
     
     NoMethodError:
       undefined method `display_rights' for #<ResourceDecorator:0x0000562f16f02800>
     # ./app/components/work_version_metadata_component.rb:58:in `block in attributes'
     # ./app/components/work_version_metadata_component.rb:56:in `map'
     # ./app/components/work_version_metadata_component.rb:56:in `attributes'
     # ./app/components/work_version_metadata_component.html.erb:3:in `call'
     # /usr/local/bundle/gems/view_component-2.18.0/lib/view_component/base.rb:78:in `render_in'
     # /usr/local/bundle/gems/view_component-2.18.0/lib/view_component/render_component_helper.rb:6:in `render_component'
     # /usr/local/bundle/gems/view_component-2.18.0/lib/view_component/test_helpers.rb:27:in `render_inline'
     # ./spec/components/work_version_metadata_component_spec.rb:7:in `block (2 levels) in <top (required)>'
     # ./spec/components/work_version_metadata_component_spec.rb:35:in `block (3 levels) in <top (required)>'
```

awead (approved):

PR #772, opened 2020-12-11
use deposit date in meta tag if publication date is empty

### LINKED ISSUE #768: Set Publication Date Meta Tag to Deposit Date if Publication Date is Empty
For works that have no explicit publication date, we should set the publication date meta tag (`<meta name="citation_publication_date" content="2020-11-28">`) to the deposit date. 

(So that works will be indexed by google scholar)

srerickson:
fixes #768

### DIFF
--- a/app/decorators/resource_decorator.rb
+++ b/app/decorators/resource_decorator.rb
@@ -43,4 +43,11 @@ class ResourceDecorator < SimpleDelegator
       creator_aliases.take(3)
     end
   end
+
+  # @note Google Scholar prefers the year alone
+  def published_date_or_deposited_year
+    return deposited_at.year unless EdtfDate.valid?(published_date)
+
+    Date.edtf(published_date).year
+  end
 end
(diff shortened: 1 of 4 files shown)


srerickson:
fyi, @awead , I want to use just the year-part of the deposit date-

awead:
@srerickson I made some changes and took into account that the dates are going to EDTF, which by themselves, Google may not like. According to [Google Scholar](https://scholar.google.com/intl/en/scholar/inclusion.html#indexing):

> The publication date tag, e.g., citation_publication_date or DC.issued, must contain the date of publication, i.e., the date that would normally be cited in references to this paper from other papers. Don't use it for the date of entry into the repository - that should go into citation_online_date instead. Provide full dates in the "2010/5/12" format if available; or a year alone otherwise. This tag is required for inclusion in Google Scholar. 

We could try doing YYYY/MM/DD, but I opted for just year, which would make it consistent with the other resources that don't have publication dates.

awead (approved):

PR #813, opened 2021-01-15
Fix bug in the ordering of works in the collection edit form

### LINKED ISSUE #790: Collections ordering inconsistency
While working on this [this collection](https://scholarsphere.psu.edu/resources/48e7319c-0321-4094-85ed-cbf83ea1a062) I noticed a problem with collection work ordering: I added a work to the collection (the readme) and set it to be the first item in the collection; that worked, but I noticed that when I go back to the collection edit form, the order is not the same (readme is no longer first).

rschenk:
When the works were being shown on the Collection Edit form, the position ordering wasn't being observed. This fixes that.

Unrelated to the bug, but related to that area of the code, I added some database eager loading to the the controller that renders that form. Again, this was not the cause of the bug, and did not affect said bug, but while I was there I thought I'd add a little performance boost.

Fixes #790

### DIFF
--- a/app/controllers/dashboard/form/members_controller.rb
+++ b/app/controllers/dashboard/form/members_controller.rb
@@ -4,7 +4,9 @@ module Dashboard
   module Form
     class MembersController < BaseController
       def edit
-        @resource = Collection.find(params[:id])
+        @resource = Collection
+          .includes(collection_work_memberships: [:work])
+          .find(params[:id])
         (@member_works, _deprecated_document_list) = search_service.search_results
         authorize(@resource)
       end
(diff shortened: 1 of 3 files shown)


app/controllers/dashboard/form/members_controller.rb
```diff
@@ -4,7 +4,9 @@ module Dashboard
   module Form
     class MembersController < BaseController
       def edit
-        @resource = Collection.find(params[:id])
+        @resource = Collection
+          .includes(collection_work_memberships: [:work])
+          .find(params[:id])
```
rschenk, line 9: Not the cause of the bug, but I figured while I was in here looking around I would boost the db performance a little.

awead (approved):

PR #816, opened 2021-01-15
Add upload endpoint to API

### LINKED ISSUE #815: Upload files via presigned urls in the API
# Feature Request

A client can request a presigned url from Scholarsphere in order to upload a file into our S3 instance.

awead:
Returns a presigned url to a client so that they may upload a file into our S3 instance. The endpoint requires an extension, such as pdf, or txt, so that it can construct a path for the file in the bucket.

Filenames are unique UUIDs, matching the current behavior in the UI via the Shrine/Uppy process.

Fixes #815

### DIFF
--- a/app/controllers/api/v1/uploads_controller.rb
+++ b/app/controllers/api/v1/uploads_controller.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Api::V1
+  class UploadsController < RestController
+    def create
+      render json: {
+        url: signer.presigned_url(:put_object, bucket: ENV['AWS_BUCKET'], key: key),
+        id: id,
+        prefix: prefix
+      }
+    end
+
+    private
+
+      def signer
+        Aws::S3::Presigner.new(
+          client: Aws::S3::Client.new(**s3_options)
+        )
+      end
+
+      def s3_options
+        if ENV.key?('S3_ENDPOINT')
+          base_options.merge(endpoint: ENV['S3_ENDPOINT'], force_path_style: true)
+        else
+          base_options
+        end
+      end
+
+      def base_options
+        {
+          access_key_id: ENV['AWS_ACCESS_KEY_ID'],
+          secret_access_key: ENV['AWS_SECRET_ACCESS_KEY'],
+          region: ENV['AWS_REGION']
+        }
+      end
+
+      def key
+        "#{prefix}/#{id}"
+      end
+
+      def prefix
+        Scholarsphere::ShrineConfig::CACHE_PREFIX
+      end
+
+      def id
+        @id ||= "#{SecureRandom.uuid}.#{extension}"
+      end
+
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/uploads_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    private
+
+      def signer
+        Aws::S3::Presigner.new(client: client)
+      end
+
+      def client
+        Aws::S3::Client.new(
```
whereismyjetpack: I haven't had the chance to pull this down yet, but S3_ENDPOINT, and force_path_style are options that are only used for non-aws buckets. We conditionally add these in the shrine config (https://github.com/psu-stewardship/scholarsphere/blob/develop/lib/scholarsphere/shrine_config.rb#L21)

whereismyjetpack:
Screen Shot 2021-01-18 at 9 29 53 AM
If we only send the extension along with the presign request, will we miss out on this? or is this part of the 3rd metadata step? I realized that you've probably already got that sorted after seeing that. 

I've verified that the presign url is only good for the path given, and will throw an invalid signature if the uploader tries to rename the file. Given that, I'm also curious if we _only_ need to reply with the url in the POST request for a presign url?

I did notice that if an API key is not given, we're given 

```{
  "message": "We're sorry, but something went wrong",
  "errors": [
    "NoMethodError",
    "undefined method `application' for nil:NilClass"
  ]
}
```

 I wonder if we should catch that?

awead:
@whereismyjetpack We're assuming the client has the file locally, and the file name, so it just sends the request using the extension. It'll get back the url, uuid + file extenion, and prefix. Using that information, it'll assemble a path to the file, `prefix/uuid.extension` and send that along to Scholarsphere. So the steps look like:

1) Client sends request for pre-signed url to Scholarsphere
2) Using the returned URL, client uploads file to AWS
3) Assuming step 2 is successful, client sends ingest/create request to Scholarsphere with the work's metadata, and the file metadata created from the response in step 1.

Regarding a missing key, we should be catching that error. How were you able to create that error?

whereismyjetpack:
Gotcha. 


To recreate the error, i just sent ```curl -X POST http://localhost:3000/api/v1/upload``` without the headers.

awead:
@whereismyjetpack I fixed that error. Our tests weren't catching it because RSpec's controller environment doesn't fully replicate a real-world scenario. If this PR looks good, do you want to approve?

whereismyjetpack:
I think the S3_ENDPOINT configuration will need to be fixed yet, force_path_style should be off, or false for aws, and we aren't sending S3_ENDPOINT in real deployments. I wonder if this code should only be in one place?

awead:
@whereismyjetpack this is fixed now.

whereismyjetpack (approved):

Issue #805, opened 2021-01-11
Setting withdrawn state breaks the dashboard


awead:
# Bug Report

## Description
If you set a work version to `withdrawn` it breaks the dashboard because we don't have a label for that state.

## To Reproduce
1. Set a work version to `withdrawn`
2. Visit the dashboard as an admin

## Questions

1. What happens to the work if there is only one published version? Does it remain indexed? It probably ought to be removed since draft versions do not have their parent work indexed. See https://github.com/psu-stewardship/scholarsphere/blob/develop/app/indexers/work_indexer.rb#L7

## WDLL

* fully implement the withdrawn state, with proper test coverage
* include a service that handles the indexing (?)
* display message on resource page indicating withdrawn state (make it look like embargoed message)
* ensure metadata is accessible but files are not

awead:
In answer to question 1: No, the Solr record for both the work and work version will remain. The work is still publicly visible, just like an embargoed work is, but its files cannot be downloaded. The work version record remains as well, and is searchable in the user's dashboard.

PR #824, opened 2021-01-20
DeleteWork service

awead:
Adds a command-line only DeleteWork service to delete works when needed. It is only intended to be run from the console, and as yet, is not linked anywhere else in the application.

### DIFF
--- a/app/services/delete_work.rb
+++ b/app/services/delete_work.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# @abstract Deletes a work, including all of its versions, without regard to its publication data. Note, this service
+# is not intended to be used within the context of the application, but rather as a command-line only method reserved
+# for instances when a specific work needs to be deleted.
+
+class DeleteWork
+  def self.call(uuid)
+    new(uuid: uuid).destroy
+  end
+
+  attr_reader :work
+
+  def initialize(uuid:)
+    @work = Work.find_by(uuid: uuid)
+  end
+
+  def destroy
+    work.versions.map do |version|
+      version.aasm_state = 'withdrawn'
+      version.destroy!
+      IndexingService.delete_document(version.uuid)
+    end
+    work.destroy!
+    IndexingService.delete_document(work.uuid, commit: true)
+  end
+end
--- a/spec/services/delete_work_spec.rb
+++ b/spec/services/delete_work_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DeleteWork do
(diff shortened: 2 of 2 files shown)


app/services/delete_work.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  attr_reader :work
+
+  def initialize(uuid:)
+    @work = Work.find_by(uuid: uuid)
+  end
+
+  def destroy
+    work.versions.map do |version|
+      version.aasm_state = 'withdrawn'
```
rschenk, line 20: Is this line necessary? (is never persisted to db)
If it's necessary, is it OK that we're circumventing the state machine API?

spec/services/delete_work_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  describe '.call' do
+    let!(:work) { create(:work, versions_count: 3, has_draft: true) }
+
+    before { class_double(SolrIndexingJob) }
+
+    specify do
+      expect(Work.count).to eq(1)
+      expect(WorkVersion.count).to eq(3)
+      described_class.call(work.uuid)
+      expect(SolrIndexingJob).not_to receive(:perform_later)
```
rschenk: I think, because this line is located after the `.call` that the rspec matcher needs to be `have_received` not `receive`

rschenk (changes requested):

spec/services/delete_work_spec.rb
awead: I tried that, but I can't seem to get it to work with class doubles.

app/services/delete_work.rb
awead: I put this in 211072880dac71b55656d66ef32aad114962c7e6 that prevents deleting published versions. The whole delete process really needs to be fleshed out a bit more, specifically by fully implementing a withdrawn state. That's ticketed here #805. This PR is to address the short-term need of just deleting something outright, which I get asked to do... and then have to remember to do all these steps.

spec/services/delete_work_spec.rb
awead: If I change it do:

``` ruby
described_class.call(work.uuid)                                                                                      
expect(SolrIndexingJob).not_to have_received(:perform_later) 
```
I get:
```
Failure/Error: expect(SolrIndexingJob).not_to have_received(:perform_later)
       #<SolrIndexingJob (class)> expected to have received perform_later, but that object is not a spy or method has not been stubbed.
```

rschenk: Try using `class_spy`

rschenk (approved):

Issue #828, opened 2021-01-21
Version Endpoint


whereismyjetpack:
# Feature Request
A route/api/or healthcheck that displays version information
## Describe the solution you'd like
A way to tell from the application what version it thinks it is
## Any alternatives you've considered?
i have not.
## What does "done" look like?
a request to an endpoint can display version information to the user. 
How will we know when we've successfully implemented this?

I haven't thought too much about what this might look like, does it return json, html, do we have both formats etc. Opening this up as a way to start the discussion. 

Presently the deployments get injected into them an `APP_VERSION` environment variable, and this matches up with the image tag, (git commit sha, or version tag)

awead:
The easiest would be a health check. That seems to be the least invasive. API would be good to, and perhaps more formal, since it could be included as part of the spec, but currently all API requests require authentication, and it'd be nice to offer this without that.

Also, why not both?

whereismyjetpack:
I think starting with a healthcheck makes sense, all the machinery is already there, and adding to it should be trivial. 

I can start by just pulling APP_VERSION and we an evolve it over time?

PR #833, opened 2021-01-22
adds version_check healthcheck

whereismyjetpack:
First stab at an version endpoint. addresses #828

### DIFF
--- a/lib/healthchecks/version_check.rb
+++ b/lib/healthchecks/version_check.rb
@@ -3,8 +3,7 @@
 module HealthChecks
   class VersionCheck < OkComputer::Check
     def check
-      version = ENV.fetch('APP_VERSION', nil)
-      mark_failure unless version
+      version = ENV.fetch('APP_VERSION', 'unknown')
       mark_message version.to_s
     end
   end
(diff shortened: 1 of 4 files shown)


lib/healthchecks/version_check.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module HealthChecks
+  class VersionCheck < OkComputer::Check
+    def check
+      version = ENV.fetch('APP_VERSION', nil)
```
awead: What about `version = ENV.fetch('APP_VERSION', 'unknown')` and that way there's always something to display? Or is it important that a missing version key be communicated as a failure?

awead (changes requested):
Looks good! One idea, but we can dismiss it if you like.

lib/healthchecks/version_check.rb
whereismyjetpack: I don't think it's important. "unknown" sounds good to me!

whereismyjetpack:
I think this is ready to rock and roll.

awead (approved):

PR #834, opened 2021-01-22
Update creators' positions when null

### LINKED ISSUE #822: Author order of migrated works
# Bug Report

## Description
Author ordering of migrated works is not consistent with the original author order (from SS3). I'm not sure if this is related to recent work on #809.

## To Reproduce
Look at a migrated work version (i.e., the *first version* for any work that was migrated) with multiple authors; compare to the work on ScholarSphere 3.0. The order of authors is reversed. Here is an example (login required since it has an embargo on SS3):

SS4: https://scholarsphere.psu.edu/resources/bfb925f4-3047-4905-beca-cc66f0a23ade
SS3: https://scholarsphere-3.libraries.psu.edu/concern/generic_works/tb2773z02w

## Expected behavior
The orders should be the same

awead:
Due to a migration oversight, migrated works did not have any position indicators. Additionally, there was a bug that was not copying over a creator's positioning indicators to new versions. The bug has since been fixed, but there remained many creator records will null positions.

This is a service class, designed to run in the console only, and add any missing positions. Any first version without positions has them added according to the order of the primary key. This represents the
order in which they were added to the database. Any subsequent versions that don't have positions have them copied over from their previous versions.

The process is idempotent and can be re-run multiple times and never affect creators that have position values. If a version is encountered that somehow has a mix of creators with and without null positions, it
aborts.

Fixes #822 

## Note

I was initially going to throw this code away, but since it might come in handy later, I thought we'd keep it around. I've put it under `app/services` but I'm not sold that's where it should live.

### DIFF
--- a/app/services/creator_order_fix.rb
+++ b/app/services/creator_order_fix.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class CreatorOrderFix
+  class << self
+    def call
+      Work.find_each do |work|
+        work.versions.order(:version_number).each_with_index do |work_version, index|
+          if index.zero?
+            update_first_version(work_version)
+          else
+            update_later_version(work_version, work.versions[index - 1])
+          end
+        end
+      end
+    end
+
+    def update_first_version(version)
+      return unless all_empty?(version.creator_aliases)
+
+      version.creator_aliases.sort_by(&:id).each_with_index do |creator_alias, index|
+        creator_alias.update(position: (index + 1) * 10, changed_by_system: true)
+      end
+    end
+
+    def update_later_version(current, previous)
+      return unless all_empty?(current.creator_aliases)
+
+      previous_ca_lookup = previous.creator_aliases.map { |ca| [ca.actor_id, ca.position] }.to_h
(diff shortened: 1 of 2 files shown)


app/services/creator_order_fix.rb
```diff
@@ -0,0 +1,44 @@
...
+      current.creator_aliases.each do |creator_alias|
+        creator_alias.update(position: previous_ca_lookup[creator_alias.actor_id], changed_by_system: true)
+      end
+    end
+
+    def all_empty?(aliases)
+      positions = aliases.map(&:position)
+
+      if positions.uniq.count > 1 && positions.uniq.any?(nil)
+        raise StandardError, "Work #{aliases.first.work_version.work.uuid} can't be corrected"
```
rschenk, line 45: Everything here looks good, I'm just wondering if we do actually want this to error out. Like, _if_ this is intended to only be run in the console by a human, perhaps we don't want it to die mid-run if it encounters some bogus data. Maybe what we want it to do is log the bad data, and keep chugging through the rest of the database. Then perhaps at the very end (last line of the `.call` method), it could `puts` out all the ones it couldn't automatically fix?

PR #837, opened 2021-01-26
Remodel creators as authors for works and collections

### LINKED ISSUE #818: Update database schema with new authorship model requirements
Makes changes to the database to account for new authorship model

## WDLL

* change WorkVersionCreations
    * rename table to Author (or creator?)
    * add given_name, surname, and email fields in WorkVersionCreations
    * remove foreign key constraint to Actor table
* make polymorphic relationship between Collection, WorkVersion, and Author
    * rename work_version_id to resource_id
    * add resource_type field
* update Ruby models to reflect new changes
* UI interactions should remain unchanged

awead:
This is the first commit of a multi-commit process to remodel creators as authors, removing the requirement that every creator also be tied to a unique actor. This will support the ability to add creators to a work
or collection without requiring a unique identifier such as a Penn State access id or Orcid.

A new, polymorphic Author table is added to replace the existing WorkVersionCreation and CollectionCreation.  An author record has an optional link to an actor so that if the creator does have a unique
identifier, we can link each of their authorships to single, disambiguated person. If the creator has no identifier, then they existing solely in Author and have no corresponding Actor record.

While the database schema is altered, no data should be modified or destroyed with this commit. Migrating existing data into the new schema will be in a future commit.

Fixes #818


app/models/actor.rb
```diff
@@ -80,6 +73,18 @@ class Actor < ApplicationRecord
 
   after_destroy :update_index_async
 
+  def created_work_versions
+    authorships.where(resource_type: 'WorkVersion')
+  end
+
+  def created_works
+    created_work_versions.map(&:work).uniq
```
rschenk: This will be very expensive to do in Ruby and should probably be done as a SQL inner join instead. If 'created_work_versions' is an actual association then you should be able to have Rails write this using a `through:` just like before.

app/models/actor.rb
```diff
@@ -80,6 +73,18 @@ class Actor < ApplicationRecord
 
   after_destroy :update_index_async
 
+  def created_work_versions
+    authorships.where(resource_type: 'WorkVersion')
```
rschenk: I think this is confusing, because unless I am reading this wrong, this method returns back an array of Author objects that belong to Work Versions, not an array of WorkVersions themselves.

I think this should probably be an association instead. 
`has_many :work_versions, through: :authorships, -> { where resource_type: 'WorkVersion' }`

app/models/actor.rb
```diff
@@ -80,6 +73,18 @@ class Actor < ApplicationRecord
...
+  end
+
+  def created_works
+    created_work_versions.map(&:work).uniq
+  end
+
+  def created_collections
+    authorships.where(resource_type: 'Collection')
```
rschenk: Ditto above

app/models/author.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class Author < ApplicationRecord
```
rschenk: I have a hard time with this class name, because this is not a table of Authors in my mind. "Adam Wead" and "Ryan Schenk" (who are authors, without the capital-A) will show up multiple times in this table, once per WorkVersion or Collection that they are listed on. So in my mind this class does not model an Author, it models an author being attached to a particular thing. I like the "Authorship" name that you use on the relation, maybe that would be a better name for this class and table?

rschenk (changes requested):

app/models/actor.rb
awead: Thanks, I didn't know how it would work with the polymorphism. I'm honestly not sure if these associations are even useful. There can't be an inverse on `WorkVersion` because it would exclude those authors that don't have actor references. So it seems it's just a convenient kind of query. Or, am I forgetting something?

awead:
@rschenk This is all ready if you want to poke at it on Friday. However, for some reason, the authorship diff tests are failing in CI but not locally. I won't be able to look at that until Monday.

PR #838, opened 2021-01-27
I#313 collection dois

### LINKED ISSUE #313: Mint Collection DOIs
Collections may have DOIs, but they need to be minted.

## WDLL

1. Create a metadata service for collections that takes an existing collection and formats the metadata for DataCite
2. UI component to trigger the minting process
3. Integration coverage for testing
4. Ensure that it works with the existing DOI job

rschenk:
Previously we had a class `DataCite::Metadata` that took a WorkVersion and turned it into the metadata schema that DataCite requires to publish a DOI.

With the advent of minting DOIs for Collections, I needed to write a corresponding metadata adapter to translate a Collection's metdata into datacite format. To do this, I extracted what previously was done in `DataCite::Metadata` and moved it to `DataCite::MetaData::WorkVersion`. Then I made a corresponding `DataCite::Metadata::Collection`, and extracted the common parts into a superclass `DataCite::Metadata::Base`

The second commit is where I modified our `DoiService` object to be able to handle Collections, in addition to Works and WorkVersions.

This was a big refactor, where ultimately I made two internal strategy classes, one for handling Works-and-Versions, and another for handling Collections.

I also un-DRY'd some of this code, which RuboCop did complain about (and surely CodeClimate will too). The reason I did this was because, when refactoring, the code was so DRY that it was really, really hard to figure out what it was actually doing. Essentially we've got a four-state truth table, which does four similar but slightly different things. I decided to just write this all out rather than DRY it up because I think it's easier to follow.

Closes #313


lib/data_cite/metadata/base.rb
```diff
@@ -0,0 +1,101 @@
...
+        def publication_year
+          publication_date&.year
+        end
+
+        def publication_date
+          parsed_publication_date || resource.created_at
+        end
+
+        def parsed_publication_date
+          Time.zone.parse(resource.published_date)
```
rschenk: Ohohohoho. This was written before `published_date` was an EDTF. I'm going to add an additional commit to handle EDTF here

app/services/doi_service.rb
```diff
@@ -2,83 +2,149 @@
...
-      existing_doi = resource.doi.presence
-      should_save_new_doi = existing_doi.blank?
-
-      publish_doi(
-        doi: existing_doi, # will register a new doi if nil
-        update_resource: should_save_new_doi
-      )
+    def initialize(resource)
+      unless self.class.applicable_to?(resource)
+        raise ArgumentError, "#{self.class.name} expects resource to be a Work or WorkVersion"
```
awead, line 46: This line isn't getting covered in the tests, but that might not be important.

app/services/doi_service.rb
```diff
@@ -2,83 +2,149 @@
...
     end
 
-    def publish_doi(doi:, update_resource:)
-      metadata = metadata_source
-        .call(work_version: work_version, public_identifier: resource.uuid)
-        .tap(&:validate!)
-        .attributes
+    def initialize(collection)
+      unless self.class.applicable_to?(collection)
+        raise ArgumentError, "#{self.class.name} expects resource to be a Collection"
```
awead, line 116: Ditto here.

lib/data_cite/metadata/base.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+module DataCite
+  module Metadata
+    class Base
+      class Error < StandardError; end
+      class ValidationError < Error; end
```
awead: This is nit-picky, so take it or leave it, but could you just move the class definition to Metadata? DataCite::Metadata::Base::Error seems wordy, and having DataCite::Metadata::Error matches up with DataCite::Client::Error.

lib/data_cite/metadata/base.rb
```diff
@@ -0,0 +1,103 @@
...
+        def publication_date
+          parsed_publication_date || resource.created_at
+        end
+
+        def parsed_publication_date
+          return nil unless EdtfDate.valid?(resource.published_date)
+
+          Date.edtf(resource.published_date)
+        rescue ArgumentError, TypeError
+          nil
```
awead, line 89: There's no coverage here. We're covering the `.valid?` case, which is probably fine.

spec/lib/data_cite/metadata/work_version_spec.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-require 'spec_helper'
+require 'rails_helper'
```
awead, line 3: I literally _just_ made the same change yesterday!

awead (dismissed):
I just pointed out a few things. This is all good to go, but if you wanted to address anything I suggested, feel free!

app/services/doi_service.rb
rschenk: It's true that it's not being covered in tests, because `DoiService.call` will never instantiate one of these instances unless it has already passed the `applicable_to?` test when selecting a strategy. This line probably doesn't even need too be there but I thought it was worth being prudent. ¯\_(ツ)_/¯

lib/data_cite/metadata/base.rb
rschenk: That's a good idea, and I think I'll do that. That's the way it was before, but I must have lost it in the refactor madness

lib/data_cite/metadata/base.rb
rschenk: I'm pretty sure .valid? will always catch everything, this rescue was left over from the previous way we were doing it (parsing times without edtf) and I figured it couldn't hurt to have it.

awead (approved):

PR #841, opened 2021-02-01
Return an empty array when graph groups are not present

### LINKED ISSUE #839: Can't login when groups are nil
# Bug Report

## Description
We are getting 
```
User::OAuthError (undefined method `compact' for nil:NilClass
``` 
for @DanCoughlin  when he logs in. I can't be sure, but it appears that the id_token returned from azure has no groups key. I'm not sure under what circumstances this could happen. but apparently it can. In the strategy, maybe we should always return an array in the strategy?

here is where the error is being thrown
https://github.com/psu-stewardship/scholarsphere/blob/develop/app/models/user.rb#L71

## To Reproduce
1. Go to '.Home page'
2. Click on '.Login' (and login as @DanCoughlin )

awead:
If the Azure graph groups flag is not present, then we need to return an empty array to avoid NilClass errors.

This recycles the current VCR responses and re-works the existing tests to add a check for a null environment variable.

Fixes #839


lib/omniauth/strategies/azure_oauth.rb
```diff
@@ -19,14 +19,10 @@ class AzureOauth < OmniAuth::Strategies::OAuth2
       end
 
       info do
-        raw_info['groups'] = graph_groups if ENV['AZURE_GRAPH_GROUPS'].present?
+        raw_info['groups'] = graph_groups
```
whereismyjetpack: raw_info['groups'] should have in it the groups claim, if we always call graph_groups on it we'll always return an empty array which probably isn't what we want. I took a slightly different approach on Friday. https://github.com/psu-stewardship/scholarsphere/pull/840/files

whereismyjetpack (changes requested):

lib/omniauth/strategies/azure_oauth.rb
awead: Right, I saw that. What isn't clear is where or how the groups key is getting populated prior to this, assuming it is...

whereismyjetpack (dismissed):

whereismyjetpack (approved):

PR #844, opened 2021-02-03
Client for the Orcid public API

### LINKED ISSUE #820: Integrate with Orcid to query known ids
Using Orcid's API, we can query with a user-given id and verify the user exists, as well as use the metadata from Orcid's database to populate the form for adding new creators.

## WDLL

* obtain API keys, either the public API or the "members-only" one which Penn State has access to
* libraries and/or service objects to interact with the API (Datacite is a good example here)
* application can query API with a known Orcid ID and receive complete metadata or the appropriate error response

awead:
Adds a client to integration with Orcid's public API. Currently, only querying the `/person` operation with a given orcid is supported.

Fixes #820

### DIFF
--- a/app/models/orcid_id.rb
+++ b/app/models/orcid_id.rb
@@ -4,7 +4,7 @@
 # and can return any of those formats. The unformatted numeric string is preferred for database storage, but this
 # class be used to present the id in alternative, human-readable formats.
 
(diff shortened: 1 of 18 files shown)


rschenk (approved):
Looks good to me

PR #852, opened 2021-02-08
Read-only mode

### LINKED ISSUE #848: Maintenance Page
We need to be able to place Scholarsphere in maintenance mode or "read-only" mode so we can prevent users from making any changes while we perform updates.

## WDLL

* configurable "switch" to turn on a splash page
* ideally, catalog (i.e. Solr) searches would still work
* prevents any database changes
* logins prohibited

awead:
When setting the appropriate environment variable, Scholarsphere will go into "read-only" mode, during which only Solr and database reading operations will work.  Logins are prohibited as well as any other database writing operations.

This is a UI-level change only. Changes via the console are still permissible.

A flash message informs the user that Scholarsphere's content cannot be changed during this time.

Fixes #848 

Screen Shot 2021-02-05 at 2 37 29 PM

### DIFF
--- a/app/components/flash_message_component.rb
+++ b/app/components/flash_message_component.rb
@@ -8,9 +8,11 @@ class FlashMessageComponent < ApplicationComponent
   end
 
   def messages
-    flash.map do |message|
-      Message.new(*message)
-    end
+    read_only_message + flash_messages
+  end
+
+  def render?
+    messages.present?
   end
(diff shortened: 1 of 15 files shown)


srerickson (approved):
This looks good. One additional feature (part of a separate PR, probably) would be to make the banner text configurable.

PR #854, opened 2021-02-08
Migrate Authorships

### LINKED ISSUE #819: Migrate data to new Authorship table
Existing data from WorkCreations and CollectionCreations needs to copied into Authorship.

Blocked by #818 

## WDLL

Consists of a migration script or rake task that does the following:

* for each Authorship:
    * copy given_name, surname, and email from the linked Actor record
    * update resource_type to be WorkVersion, if necessary
* for each CollectionCreations, insert a new record into Authorship:
    * copy CollectionCreations.alias to Authorship.alias
    * copy CollectionCreations.position to Authorship.position
    * copy given_name, surname, and email from the linked Actor record
    * resource_type should be Collection
    * resource_id is the collection id
* drop CollectionCreations
* for each Actor that does not have a PSU id or Orcid:
    * update linked Author record so that actor_id is null
* update the versions table
    * copy `work_version_id` to `resource_id`
    * set `resource_type` to `WorkVersion`

rschenk:
For WorkVersions:
* Loads out all the PaperTrail::Versions corresponding to any WorkVersionCreation (creator) records that have ever been associated with it
* Groups them by their WorkVersionCreation id
* Finds and loads out the associated Actor
* Builds a base set of attributes for the new Authorship object based on that Actor
*Iterates through all the changes in the paper trail version history, and applies the same changes to the Authorship, so that it is built, edited, and destroyed in the same order as the WVC
* Maintains all the original timestamps
* Works even on creators that were deleted, and therefore have no record in the work_version_creations table anymore

For Collections:
* There is no PaperTrail version history for Collections creators
* Load out the associated Actor
* Builds a base set of attributes for the new Authorship object based on that Actor
* Merges in the other attributes that are in the CollectionCreation

Fixes #819

### DIFF
--- a/spec/services/authorship_migration/work_version_creation_migration_spec.rb
+++ b/spec/services/authorship_migration/work_version_creation_migration_spec.rb
@@ -171,20 +171,16 @@ RSpec.describe AuthorshipMigration::WorkVersionCreationMigration, type: :model, 
         @wvc1.versions.first.destroy!
       end
 
-      it 'reports the error, continues to migrate the rest of the creators' do
+      it 'still works' do
         authorship_count_before = Authorship.count
         versions_count_before = PaperTrail::Version.count
 
         errors = perform_call
 
-        expect(errors.length).to eq 1
-        expect(errors.first).to match(/WorkVersion##{work_version.id}/i)
-          .and match(/cannot find the papertrail::version/i)
+        expect(errors).to be_empty
 
-        # It skips the first creator because it errors out, but continues to
-        # migrate the other two (one of which is deleted)
-        expect(Authorship.count - authorship_count_before).to eq 1
-        expect(PaperTrail::Version.count - versions_count_before).to eq 3
+        expect(Authorship.count - authorship_count_before).to eq 2
+        expect(PaperTrail::Version.count - versions_count_before).to eq 5
       end
     end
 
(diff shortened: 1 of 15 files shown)


spec/services/authorship_migration/work_version_creation_migration_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+        expect(authorship1.instance_token).to be_present
+        expect(authorship1.created_at.to_date).to eq Time.zone.local(2021, 1, 1).to_date
+        expect(authorship1.updated_at.to_date).to eq Time.zone.local(2021, 1, 3).to_date
+
+        expect(authorship1.versions.length).to eq 3
+        v1, v2, v3 = authorship1.versions
+        expect(v1.whodunnit).to eq user.to_gid.to_s
+        expect(v1.event).to eq 'create'
+        expect(v1.changeset.slice(:alias, :actor_id, :position, :created_at, :updated_at))
+          .to eq(@wvc1.versions[0].changeset.slice(:alias, :actor_id, :position, :created_at, :updated_at))
```
awead: This one test is failing because the first changeset has `display_name` while the second has `alias`.  We we'll probably need to break this down into

```ruby
expect(v1.changeset[:display_name]).to eq(@wvc1.versions[0].changeset[:alias])
expect(v1.changeset[:actor_id]).to eq(@wvc1.versions[0].changeset[:actor_id])
expect(v1.changeset[:position]).to eq(@wvc1.versions[0].changeset[:position])
expect(v1.changeset[:created_at]).to eq(@wvc1.versions[0].changeset[:created_at])
expect(v1.changeset[:updated_at]).to eq(@wvc1.versions[0].changeset[:updated_at])
```

awead (changes requested):
Looks good. I ran this locally using seed data without any errors.

awead (approved):

PR #858, opened 2021-02-10
Disable saved searches

### LINKED ISSUE #857: Clear out Search History
# Bug Report

## Description
Presently we aren't managing search history, and the searches table grows. 

## Expected behavior
A way to elminate, or clear old search history

## Screenshots
```
 public | searches                    | table | postgres | 447 MB     |
 ```

Related issue
https://github.com/projectblacklight/blacklight/issues/1652

Cronjob seems to be a common pattern. if we had a rake task, we could use a kubernetes cronjob to clear it on a schedule. More desirable would be to turn off the functionality as it doesn't seem we ever use it.

awead:
A feature of Blacklight allows users to save searches for future reference; however, we don't use this and a side-effect is that the searches table can grow excessively large. This disables any searches from being saved into the table.

Fixes #857

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -220,4 +220,13 @@ class CatalogController < ApplicationController
     # default 'mySuggester', uncomment and provide it below
     # config.autocomplete_suggester = 'mySuggester'
   end
+
+  private
+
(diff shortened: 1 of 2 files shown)


awead:
@whereismyjetpack this will prevent any further records from being written, but we'll still need to truncate the table after this is deployed. I figure we can just do that in console.

rschenk (approved):

PR #862, opened 2021-02-18
updates solr schema to be 8.6.3 compatible

whereismyjetpack:


whereismyjetpack:
inspiration taken from https://github.com/samvera/active_fedora/commit/60e4927f14d8b67b84d138e94e7c0afe702d640f

solr/conf/schema.xml
```diff
@@ -107,8 +107,8 @@
 
     <dynamicField name="*_dtri" type="rdate" stored="false" indexed="true" multiValued="false"/>
     <dynamicField name="*_dtrim" type="rdate" stored="false" indexed="true" multiValued="true"/>
-    <dynamicField name="*_dtrs" type="rdate" stored="true" indexed="false" multiValued="false"/>
-    <dynamicField name="*_dtrsm" type="rdate" stored="true" indexed="false" multiValued="true"/>
+    <dynamicField name="*_dtrs" type="rdate" stored="true" indexed="true" multiValued="false"/>
```
awead: We probably want to just remove these. The `_dtrs` and `_dtrsm` suffixes measn `dtr`-rdate type, `s`tored, and `m`ultivalued. We use the `_dtrsi` and `_dtrsim` suffixes because we always index our dates. The `i` stands for indexed.

I don't see us using the _dtrs or _dtrsm options, so can probably just removed them as a dynamic field.

awead (changes requested):

solr/conf/schema.xml
whereismyjetpack: I've removed those two dynamic fields, and tested a full reindex. I think this _should_ be ready to go. it's deployed if we want to look further.

awead (approved):

PR #863, opened 2021-02-18
Update UI with new Authorship workflow

### LINKED ISSUE #821: Update UI to reflect new requirements when adding creators
Because we can now add creators without identifiers, the UI needs to change a bit.

## WDLL

* Orcid is not longer required in new creator form
* require _both_ first and last name (?)
* require email (?)
* adding unidentified creators inserts new record into Author table without a linked Actor record
* entering an Orcid ID will populate the other fields with the resulting metadata

awead:
Changes the UI workflow to remove the modal and integrate searching for creators using an Orcid id. Users may search via a name, email, or access id, as well as an orcid id, and the appropriate record is returned via the Persons authority endpoint.

Screen Shot 2021-02-18 at 3 50 53 PM

Fixes #821

### DIFF
--- a/app/views/dashboard/form/contributors/_authorship_fields.html.erb
+++ b/app/views/dashboard/form/contributors/_authorship_fields.html.erb
@@ -5,18 +5,26 @@
 <div class="nested-fields contributor-wrapper js-contributor-wrapper">
   <div class="row">
     <div class="col-2">
+      <%= form.hidden_field :position, class: 'js-position-index' %>
       <span class="badge badge-dark badge--outline">
         <%= t('dashboard.form.contributors.edit.badge') %>
         <span class="badge--index"><%= index + 1 %></span>
       </span>
     </div>
 
+    <%= form.hidden_field :actor_id %>
+    <%= form.fields_for :actor do |actor_fields| %>
+      <%= actor_fields.hidden_field :email %>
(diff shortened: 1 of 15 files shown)


app/views/dashboard/form/contributors/_authorship_fields.html.erb
```diff
@@ -5,18 +5,26 @@
...
       </span>
     </div>
 
+    <%= form.hidden_field :actor_id %>
+    <%= form.fields_for :actor do |actor_fields| %>
+      <%= actor_fields.hidden_field :email %>
+      <%= actor_fields.hidden_field :given_name %>
+      <%= actor_fields.hidden_field :surname %>
+      <%= actor_fields.hidden_field :psu_id %>
+      <%= actor_fields.hidden_field :orcid %>
```
rschenk, line 21: I can't figure out why we're doing these hidden fields, but I assume this is how we provide this information back to the controller when a new actor is added from the js/autocomplete?

rschenk (approved):

app/views/dashboard/form/contributors/_authorship_fields.html.erb
awead: Yes that's correct. We don't ever update these fields, which makes me wonder if we should somehow _not_ render them in the form if the actor already exists?

PR #864, opened 2021-02-18
Collection settings

### LINKED ISSUE #743: Collection settings page
Create a settings page for collections to allow the user to mint DOIs and add additional editors to a collection. These features should mirror the exact same features in a work.

## WDLL

* collections setting page, styled the same as the one for works
* section for minting DOIs
* section for adding editors to a work
* has parity with the corresponding features in the work settings page

rschenk:
I refactored EditorsForm to be shared between the work and collections settings. 

The rest of the stuff I did not try to go crazy and super DRY, because although similar Work and Collection settings are not identical.

Fixes #743


app/views/dashboard/collections/_editors.html.erb
```diff
@@ -1,11 +1,11 @@
 <p><%= t('.explanation') %></p>
```
awead: Can this file be removed? I think you're calling `editors_form` from `collections/edit`

awead (changes requested):

spec/forms/editors_form_spec.rb
```diff
@@ -84,11 +94,12 @@
...
       it 'does not add the user and reports an error' do
-        expect(work.edit_users).to be_empty
+        expect(resource.edit_users).to be_empty
         form.save
-        expect(work.edit_users).to be_empty
+        expect(resource.edit_users).to be_empty
         expect(form.errors.full_messages).to contain_exactly(
-          'Edit users ' + I18n.t('dashboard.works.editors.not_found', access_id: access_id)
+          'Edit users ' +
+          I18n.t!('activemodel.errors.models.editors_form.attributes.edit_users.not_found', access_id: access_id)
```
rschenk, line 102: @awead here is a pro rails tip, in tests if you use the bang version `I18n.t!` then it will raise an error if the translation can't be found. We should probably do a sweep through everything in the spec directory and fix all these.

awead: Done! #865

awead (approved):

PR #883, opened 2021-02-24
Migrate FileVersionMembership's papertrail versions to the new approach

### LINKED ISSUE #872: Add :changed_by_system flag to FileVersionMembership
This will duplicate the same flagging process we have for authorships. It also necessitate another smaller migration to add the create events to a file's versions that were previously ignored.

While a little tedious, this makes the file class operate the same way as the authorship class, which is overall much better, and can potentially avoid other pitfalls in the future.

## WDLL

* `:changed_by_system` option for FileVersionMembership
* migration script or service object that adds the appropriate create events to the paper trail versions of a given file
* add an instance token attribute to file versions so that they can be tracked from one work version to the next (?)

rschenk:
Updates FileVersionMembership to the new papertrail approach taken by Authorships—where records are written to the db but flagged as `changed_by_system`.

To run the migration, you should use

```ruby
AuthorshipMigration::FileVersionMembershipMigration.migrate_all_file_version_memberships
```

This method does two things:
- Backfills missing "creation" events in papertrail, caused by the old way that `BuildNewWorkVersion` worked
- Updates all FileVersionMembership papertrail events to use the new polymorphic `resource_type` and `resource_id` metadata columns, instead of the depricated `work_version_id` column

Note: This was originally a 2-commit PR, but 559477e7b01307bdcc7a654c0546d4f4bf2ad0c9 was cherry-picked into development directly. This was so we could test it quickly. The remaining commit is in this PR.

Fixes #872

### DIFF
--- a/app/components/work_histories/work_history_component.rb
+++ b/app/components/work_histories/work_history_component.rb
@@ -75,7 +75,11 @@ module WorkHistories
       # @return [Hash<FileMembershipChangePresenter, PaperTrail::Version, User>]
       def changes_to_work_versions_files(work_version)
         PaperTrail::Version
-          .where(item_type: 'FileVersionMembership', work_version_id: work_version.id)
+          .where(
+            item_type: 'FileVersionMembership',
+            resource_id: work_version.id,
+            resource_type: 'WorkVersion'
+          )
           .map do |paper_trail_version|
             {
               component: FileMembershipChangeComponent,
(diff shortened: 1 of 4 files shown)


awead (approved):
Ryan and I worked on this and it'll be removed later, so coverage is a non issue as this was tested with QA data.

PR #896, opened 2021-03-01
Update to Omniauth 2.0

### LINKED ISSUE #262: Non-JS feature tests do no invoke the omniauth login process
When running feature tests using the default, non-js environment, omniauth is not used and there for the `User.from_omniauth` method is not called. This can lead to inconsistent results and should probably be fixed so that either with or without JS, the user's database setup process is the same.

This can probably be fixed in the FeatureHelpers module or by customizing the `login_as`, which is also used in that module.

### LINKED ISSUE #692: Resolving-CVE-2015-9284
It seems the way we are using omniauth, we are vulnerable to CVE-2015-9284. the resolution is to change the auth links to be `:post` requests vs, `:get` requests, and add the csrf protection gem. 

The solution outlined in https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284 won't suffice, since we route through `new_user_session_path`

I'm not sure if we can, in the view just change `new_user_session_path` to `user_azure_oauth_omniauth_authorize_path` and forego `new_user_session_path`?

awead:
Updating to the latest Omniauth gem supports POSTing to our Azure authentication endpoint and resolves CVE-2015-9284.

Fixes #692 
Fixes #262


awead:
@whereismyjetpack and I tested this in a deployment, and all looks good.

Issue #911, opened 2021-03-05
DOIs for migrated work can't be minted


awead:
# Bug Report

## Description

If a work was migrated with invalid metadata, it's still published, but it can't be minted. We should block minting DOIs for these works until their metadata is updated to be valid according to our current standards.

PR #913, opened 2021-03-05
Don't mint DOIs for invalid resources

awead:
Prior to this commit, we were only checking the Work resource when calling the DoiService. Now, we will check both the Work and its latest version, where we get the metadata from. If either of those are invalid, i.e. are missing some requirement, we raise an error.

Ref #911


rschenk (changes requested):
This works, but we've decided to instead skip validations when saving a DOI to the database, thus avoiding potential gotchas. 

See #916

PR #915, opened 2021-03-08
Allow admins to edit published works

### LINKED ISSUE #895: Allow admins to edit published versions
# Feature Request

In some cases, we may need to edit existing published versions. We could allow admins to use the existing edit forms to edit published versions.

I think all we'd need to do is update the policies to simply allow admins to edit published works. That's it. The links to editing, etc. should all work.

We currently do this via the console, which is fine, except the versioning history will show edits by `[unknown user]` which can look misleading.

Admins can edit _any_ version, including published ones, just like they edit any other work version.

awead:
Administrators can edit published works. Changes will appear as they typically do, in chronological order, after publishing.

Fixes #895

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -38,7 +38,7 @@ module Dashboard
         # actually attempt it. However, we want to be nice and not yell at them for something they _haven't_ seen yet
         # like rights.
         def prevalidate
-          @resource.publish
+          @resource.publish unless @resource.published?
           @resource.validate
           @resource.errors.delete(:rights)
         end
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -38,9 +38,22 @@ module Dashboard
         # actually attempt it. However, we want to be nice and not yell at them for something they _haven't_ seen yet
         # like rights.
         def prevalidate
-          @resource.publish unless @resource.published?
-          @resource.validate
-          @resource.errors.delete(:rights)
+          temporarily_publish @resource do
+            @resource.validate
+            @resource.errors.delete(:rights)
+          end
+        end
+
+        # Temporarily mark a resource as published (do not save to the
+        # database), allow some actions to be performed in a block, then reset
+        # the aasm state back to whatever it was previously.
+        def temporarily_publish(resource)
+          initial_state = resource.aasm_state
+          resource.publish unless resource.published?
+
+          yield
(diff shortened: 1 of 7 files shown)


app/views/dashboard/form/publish/edit.html.erb
```diff
@@ -44,7 +44,11 @@
             </div>
             <%= render 'fields', form: form %>
 
-            <%= render 'dashboard/form/shared/action_footer', form: form, primary_action: :publish %>
+            <% if form.object.aasm.from_state == WorkVersion::STATE_DRAFT %>
+              <%= render 'dashboard/form/shared/action_footer', form: form, primary_action: :publish %>
+            <% else %>
+              <%= render 'dashboard/form/shared/action_footer', form: form, primary_action: :finish %>
+            <% end %>
```
rschenk: This is confusing to me, and I think it's a result of our `#prevalidate` method. What I think might be a better design is for `prevalidate` to: publish, validate, roll back to the original state. Then this code here can say `if form.object.published?` rather than using the from_state

app/views/dashboard/form/shared/_action_footer.html.erb
```diff
@@ -1,7 +1,8 @@
 <% primary_action = local_assigns.fetch(:primary_action, :save_and_continue) %>
+<% button_key = current_user.admin? ? :admin_save : param_key %>
```
rschenk: How about we call this `secondary_action` instead of `button_key` since there are three buttons in this footer, and this will match up better with `primary_action` above?

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -38,7 +38,7 @@ def update
         # actually attempt it. However, we want to be nice and not yell at them for something they _haven't_ seen yet
         # like rights.
         def prevalidate
-          @resource.publish
+          @resource.publish unless @resource.published?
```
rschenk: I think I would like to refactor this method to temporarily publish the @resource, and then roll it back to whatever its initial state was. That would, I think, prevent some of the weirdness in the views.

rschenk (changes requested):

app/views/dashboard/form/shared/_action_footer.html.erb
awead: Yep, that was my first choice, but I named it button_key because it didn't follow the _exact_ same pattern with primary_action, but that doesn't matter.... `secondary_action` makes more sense.

rschenk (approved):

PR #916, opened 2021-03-08
Save DOI to database without invoking validations

### LINKED ISSUE #911: DOIs for migrated work can't be minted

rschenk:
Fixes #911 

Replaces / Closes #913

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -37,6 +37,12 @@ Rails/ApplicationController:
   Exclude:
     - 'app/controllers/api/v1/rest_controller.rb'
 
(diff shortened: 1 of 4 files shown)


awead (changes requested):
This is fine. The only thing I'd like is a comment or description in the DoiService that explains why we don't validate, etc. I can see future me not getting the diff between `update_attribute` and `update`.

awead (approved):

PR #918, opened 2021-03-09
Remove AuthorshipMigration

awead:
This was temporary code to enable us to update our data model. Now we need to remove it because upcoming changes to data validations will break these tests and the code no longer needs to be maintained.

Sorry, @rschenk...


rschenk:
😭

rschenk (approved):

awead:
giphy

PR #919, opened 2021-03-09
BuildNewActor service

### LINKED ISSUE #910: Make creating actors easier and less error-prone
Actors must be identifiable, either by Orcid or Penn State access id. Since we can query each of those sources for the information, let's make that easier to do.

## WDLL

* update Actor validations to prevent creating an actor without either an access id or an orcid
* create `ActorRegistrationService` that builds or creates a new actor when provided with either kind of id
* raise errors if the id is invalid or does not exist

awead:
To support API integrations, we need to be able to build new or existing Actor objects given an identifier. This way, a client can provide either a Penn State access id or an Orcid id, and we'll be able to create the required Actor without any additional information.

This pattern assumes the client knows the id and that it is valid; otherwise, an error will (ultimately) be returned to them.

Fixes #910


app/models/actor.rb
```diff
@@ -66,11 +64,9 @@ class Actor < ApplicationRecord
               with: /\A\d{15,15}[0-9X]{1,1}\z/,
               message: 'must be valid',
               allow_nil: true
-            }
-
-  validates :orcid,
+            },
             presence: true,
-            on: :from_user
+            unless: -> { psu_id.present? }
```
rschenk, line 73: I'm not sure this `unless psu_id.present?` is entirely in the right place. As I'm reading this, the intended behavior is, "orcid is required if psu_id is missing". This accomplishes that goal, but it _also_ affects the format regex too. I believe what will happen with this change is that, if psu_id is present, then I can type anything I want into the orcid field and its format will be accepted.

rschenk (changes requested):

app/models/actor.rb
rschenk: I think what we want is

```ruby
  validates :orcid,
            uniqueness: {
              case_sensitive: false,
              allow_nil: true
            },
            format: {
              with: /\A\d{15,15}[0-9X]{1,1}\z/,
              message: 'must be valid',
              allow_nil: true
            }
  validates :orcid,
            presence: true,
            unless: -> { psu_id.present? }
```

awead: Yep. That makes sense to me.

app/services/build_new_actor.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class BuildNewActor
+  class << self
+    def call(psu_id: nil, orcid: nil)
+      if psu_id.present?
+        build_with_psu_id(psu_id)
+      elsif orcid.present?
+        build_with_orcid(orcid)
+      else
```
awead: There's another case I missed here: what if _both_ are present. The Penn State id would take precedence, but I'm not sure we would want that.

app/models/actor.rb
awead: @rschenk I've updated this and added a test case to reflect it.

app/models/actor.rb
```diff
@@ -51,11 +51,9 @@ class Actor < ApplicationRecord
             uniqueness: {
               case_sensitive: false,
               allow_nil: true
-            }
-
-  validates :psu_id,
+            },
             presence: true,
-            on: :from_omniauth
+            unless: -> { orcid.present? }
```
rschenk: We have the same situation here as we did on `orcid`. I think we need to separate to the uniqueness validation to run all the time, and the presence validation to run `unless orcid.present?`

rschenk (changes requested):

app/models/actor.rb
awead: yep, sure enough!

rschenk (dismissed):

rschenk (approved):

PR #921, opened 2021-03-10
Drop support for migration features in the API

### LINKED ISSUE #909: Update API validations to match UI validations
The API was initially created to support migration, which required a _less_ restrictive metadata validation. Now that migration is complete, we need to update the API to align with the actual metadata restrictions imposed by the application.

## WDLL

* remove any validations associated with `:migration_api`

awead:
Removes the API migration context, which we used to validate works that were being migrated from Scholarsphere 3. Also drops support for Noids, which were only a feature of migrated work as well.

Fixes #909

### DIFF
--- a/app/controllers/api/v1/collections_controller.rb
+++ b/app/controllers/api/v1/collections_controller.rb
@@ -2,8 +2,6 @@
 
 module Api::V1
   class CollectionsController < RestController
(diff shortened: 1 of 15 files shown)


rschenk (approved):
That was a lot to take in, but I think 👍

Issue #898, opened 2021-03-03
Update PublishNewWork service


awead:
In order to integrate with RMD, the `PublishNewWork` service needs to align with the current UI pattern.

## WDLL

* accepts creators without identifiers and does not create Actor records for them
* adds position numbers for creators
* remove default license and work type, require the client to provide them
* remove support for legacy identifiers (this was only for Scholarsphere 3 noids)
* do not create works unless they can be successfully published

awead:
@srerickson Should the API have a default for work type and license, or should the client always _explicitly_ provide them?

srerickson:
The work type and license should be specified in the request, and they should be required.

PR #925, opened 2021-03-16
Update Ingest API to publish only

awead:
The ingest API now no longer creates draft works. It will only publish new works, or return an error if there isn't sufficient metadata. Additionally, all parameters are now explicit. The client must provide values for work type, visibility, and rights.

Ref #898


rschenk (approved):
OK, if I'm reading this correctly (and remembering yesterday's conversation correctly!), if you give it a value for `visibility` that is not included in the list, then it doesn't return an error, it instead sets it to the default value for that class. 

...And then when the WorkVersion is marked as published, then the "visiblity must be open or psu-only" validation kicks in?

PR #929, opened 2021-03-18
Delete unpublished works and collections

### LINKED ISSUE #867: Delete unpublished works
# Feature Request

Many times a user may create a work that doesn't get published, or we create on to "test" things. We should be able to delete this.

If the user has edit access, there's a delete button on the resource page.

* withdraw the work instead of deleting (Disadvantage: you still have to delete it at some point)
* grant specific `delete` privileges to the only the depositor/admin

## WDLL

* delete any draft version for a work
* if there's only 1 draft, then it deletes the work as well
* single button on the work settings page
* available to anyone with edit access
* admins can delete a version regardless of whether it's been published or not
* may only delete the latest version (applies to admins only)
* if deleting a later version, take the user back to the work settings page upon success; if deleting the only version, take the user back to the dashboard upon success
* must play sound effect of Strong Bad saying [deleted!](https://www.youtube.com/watch?v=hm_yHwnySX8) 👍 "HAHA, that felt great"

rschenk:
Users can delete draft works and collections from the edit form.

Admins can delete them from the edit form, too. They also have a special section of the work/collection settings page that allows them to delete published versions too. 

At the moment there's a minor bug where:
1. If the user is an admin
2. They delete a second version from the work settings page
3. They are redirected to the resource page for that work, but not back to the settings page

"Fixing" number 3 above will be rather annoying so I'm not sure it's worth the effort.

There's also a neato feature on the edit form, where large screen sizes the button says "Delete Draft" and at smaller screen sizes where there's not enough room, it automagically just says "Delete"

Closes #867

### DIFF
--- a/app/policies/work_version_policy.rb
+++ b/app/policies/work_version_policy.rb
@@ -19,12 +19,21 @@ class WorkVersionPolicy < ApplicationPolicy
   end
   alias_method :update?, :edit?
 
+  # Work versions cannot be destroyed if they haven't been persisted.
+  # Admins can delete a work version in any state, as long as it's the latest one
+  # Regular users can only delete draft versions that they can edit
   def destroy?
+    return false if record.new_record?
+    return true if user.admin? && record == record.work.latest_version
+
+    !record.published? && editable?
+  end
+
+  def publish?
     return false if record.published?
 
     editable?
   end
-  alias_method :publish?, :destroy?
 
   def download?
     return true if editable?
--- a/spec/models/work_version_spec.rb
+++ b/spec/models/work_version_spec.rb
@@ -384,24 +384,40 @@ RSpec.describe WorkVersion, type: :model do
     end
   end
 
+  describe '#force_destroy' do
+    it { is_expected.to respond_to(:force_destroy).and respond_to(:force_destroy=) }
+  end
+
   describe '#destroy' do
     context 'with a published version' do
       let(:work_version) { create(:work_version, :published) }
 
-      it 'raises an error' do
-        expect {
-          work_version.destroy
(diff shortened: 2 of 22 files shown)


app/policies/work_version_policy.rb
```diff
@@ -19,12 +19,21 @@ def edit?
   end
   alias_method :update?, :edit?
 
+  # Work versions cannot be destroyed if they haven't been persisted.
+  # Admins can delete a work version in any state, as long as it's the latest one
+  # Regular users can only delete draft versions that they can edit
   def destroy?
+    return false if record.new_record?
```
awead, line 26: Is this the same as `persisted? == false`

spec/models/work_version_spec.rb
```diff
@@ -384,24 +384,40 @@
...
 
     context 'with a draft version' do
-      let(:work_version) { create(:work_version, :draft) }
+      let!(:work_version) { create(:work_version, :draft) }
 
       it 'deletes the version' do
         expect {
           work_version.destroy
-        }.to change(described_class, :count).by(1)
+        }.to change(described_class, :count).by(-1)
```
awead, line 420: 😆 I think I did this.

awead (dismissed):
This looks good to me. I don't think your #3 is necessarily a bug. The flash notice is there, and Rails apps will often times redirect to the root path when deleting something, so I don't think it's completely unexpected.

app/policies/work_version_policy.rb
rschenk: Yup, `persisted?` and `new_record?` are inverses of each other

awead (approved):

PR #938, opened 2021-03-23
Authorship requires position

### LINKED ISSUE #928: Authorships can have a null position
# Bug Report

## Description

Although we ensure it in the UI, the position of the creator, via the Authorship, can be nil. We should make it a required attribute at the model level. If it is null, we could have weird behaviors crop up.

rschenk:
* Authorship model validates that position is present
* Collection API now accepts the position parameter
* Service object to migrate old data

To run the service object:

```ruby
# rails console

AuthorshipMigration::CollectionAuthorshipPositionFix.call
# => true if success, false if errors and errors will be printed to the stdout
```

Once the service object has been run and our old data is ported correctly on production, we should follow this up with a second release/deployment that makes the `authorships.position` field in the database schema non-nullable

Fixes #928 

## To Do
- [x] depoy this to production as quickly as possible (hotfix, whatever) #941 
- [x] run the migration code above on production
- [x] also get this merged into develop
- [x] let Ryan know when all of the above have been done, so he can begin work on #939

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -42,6 +42,7 @@ Rails/SkipsModelValidations:
     - 'app/models/api_token.rb'
     - 'app/services/doi_service.rb'
     - 'spec/services/doi_service_spec.rb'
+    - 'spec/services/authorship_migration/collection_authorship_position_fix_spec.rb'
 
 RSpec/DescribeClass:
   Exclude:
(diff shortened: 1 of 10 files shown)


awead (dismissed):

awead:
This has been deployed to a test branch, and I've run the script. No errors. You can check it out here:
https://scholarsphere-positions.dsrd.libraries.psu.edu/dashboard

rschenk:
I've updated the description with a list of todos

awead (approved):

Issue #939, opened 2021-03-24
Make Authorship position a non-null field


awead:
**NOTE!** This is a follow-up to #938 and cannot be done until that PR is merged, deployed to production, and the authorship migration script has been run.

## WDLL

* write a database migration setting Authorship position column to NOT NULL

PR #941, opened 2021-03-24
Hotfix release 4.4.1

awead:
We need plug the nil Authorship position leak. This does so by requiring the field in the model, AND includes a script to update existing nil positions in collections. This all paves the way for #939, which can't be done until any existing nil fields are cleaned up.

We decided it's best to do this sooner rather than later. The database fix in #939 can wait until  a later release, but this initial fix which will prevent any further nil values should be pushed out now.


srerickson (approved):
Fine with me.

PR #944, opened 2021-03-24
DOI API

### LINKED ISSUE #930: API endpoint for DOIs
# Feature Request

We need to know if a work exists in Scholarsphere with a given DOI. This is so an application such as RMD can determine if the work exists prior to ingesting it.

## Example

When a work exists:

`curl -X GET https://scholarpshere/api/v1/dois/10.18113/S18D2W`

Returns `200` with an array of json representations for _each_ of the resources that contain this DOI.

When a work does not exist:

`curl -X GET https://scholarpshere/api/v1/dois/10.18113/asdf`

Returns 404.

## WDLL

* API endpoint for `/dois`
* Doi schema class that does something like: `doi_ssim` field containing formatted dois from both our `doi` attribute and `identifier` attribute
    * Ex: `["doi:10.33542/g0pt-q976", "doi:10.1080/02763877.2014.879035"]`
    * Solr field should NOT collide with existing `doi_tesim` field
* array response:

## Example

```
[
  { url: "/resources/uuid1" },
  { url: "/resources/uuid2" }
]
```

### LINKED ISSUE #933: Multiple segment DOIs will not stored properly
# Bug Report

## Description

If a DOI has multiple segments, ex. `doi:10.21345/asdf/qwer`, the Doi model will not validate it correctly.

## To Reproduce

``` ruby
>   doi = Doi.new('doi:10.21345/asdf/qwer')
=> #<Doi:0x00007fc76ee8b588 @doi=#<URI::Generic doi:10.21345/asdf/qwer>>
>   doi.to_s
=> "doi:10.21345/asdf"
```

## Expected behavior

It should be `doi:10.21345/asdf/qwer`

Refer to https://www.crossref.org/blog/dois-and-matching-regular-expressions/ for a regex that should match all cases.

rschenk:
Closes #930 
Closes #933

### DIFF
--- a/app/models/doi_search.rb
+++ b/app/models/doi_search.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class DoiSearch
+  attr_reader :doi
+
+  def initialize(doi:)
+    @doi = Doi.new(doi)
+  end
+
+  # @returns an array of uuids for the matching resources
+  def results
+    return [] unless doi.valid?
+
+    doi_query = escape(doi.to_s)
+
+    solr_response = Blacklight.default_index.search(
+      q: %(all_dois_ssim:"#{doi_query}"),
+      fl: ['id'],
+      rows: max_documents
+    )
+
+    solr_response
+      .response
+      .fetch('docs', [])
+      .map { |d| d['id'] }
+  end
+
+  private
+
+    def max_documents
+      1_000
+    end
+
+    def escape(value)
+      RSolr.solr_escape(value)
+    end
+end
--- a/app/models/doi_search.rb
+++ b/app/models/doi_search.rb
@@ -13,16 +13,15 @@ class DoiSearch
 
     doi_query = escape(doi.to_s)
 
-    solr_response = Blacklight.default_index.search(
-      q: %(all_dois_ssim:"#{doi_query}"),
-      fl: ['id'],
-      rows: max_documents
-    )
-
-    solr_response
-      .response
-      .fetch('docs', [])
-      .map { |d| d['id'] }
+    Blacklight
+      .default_index
+      .search(
+        q: %(all_dois_ssim:"#{doi_query}"),
+        fl: ['id'],
+        rows: max_documents
+      )
+      .docs
+      .map(&:id)
   end
 
   private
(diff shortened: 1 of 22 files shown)


awead (dismissed):

awead:
Looks good. Would it make sense to add #933 into this as well?

app/models/doi_search.rb
```diff
@@ -0,0 +1,37 @@
...
+    solr_response = Blacklight.default_index.search(
+      q: %(all_dois_ssim:"#{doi_query}"),
+      fl: ['id'],
+      rows: max_documents
+    )
+
+    solr_response
+      .response
+      .fetch('docs', [])
+      .map { |d| d['id'] }
```
awead: You could reduce this to:
``` ruby
Blacklight
  .default_index
  .search(q: "all_dois_ssim:#{doi_query}", rows: 1_000, fl: ['id'])
  .docs
  .map(&:id)
```

Unpacking this a bit:
``` ruby
> response = Blacklight.default_index.search(q: 'all_dois_ssim:*', rows: 1_000, fl: ['id']) # => Blacklight::Solr::Response
> docs = response.docs => Array of SolrDocument
> docs.first.id
=> [uuid_of_resource]
```

SolrDocument is really just a hash, with some method names for certain solr fields such as `id`.

awead (commented):
This looks good. I made one suggestion; otherwise, 👍 . When it's ready, let me know and I'll do the formal approval thingy.

app/models/doi_search.rb
rschenk: @awead Is blacklight guaranteed to have a `.docs` that will always return an array?

rschenk: @awead The reason why I did it this way was to access the hash directly, without having it mapped to a `SolrDocument`. Since we are simply grabbing a single key verbatim, I didn't think it was worth the performance hit of  mapping those to a SolrDocument

```ruby
> response = Blacklight.default_index.search(q: 'all_dois_ssim:*', rows: 1_000, fl: ['id']) # => Blacklight::Solr::Response

> response.docs.first.class => SolrDocument

> response.response.fetch('docs').first.class => Hash
```

awead: This is way all the other searches work, so by running it through Blacklight::Solr::Response, etc. you're guaranteed to be interfacing with Solr the same way. The hash response from Solr could change, too, and RSolr/Blacklight would take care of all that.

rschenk: OK, I'll update that method to your version 😎

awead (approved):

PR #946, opened 2021-03-25
Hotfix 4.4.2 Release

awead:
Enforces a _not_ null position value in the Authorship table. The 4.4.1 release add this enforcement at the code level, and also included a script to update any existing null values. We can now add a database-level enforcement.


srerickson (approved):
ship it!

PR #950, opened 2021-03-26
Valid URLs in the Related URL metadata field are clickable

### LINKED ISSUE #947: Make Related URLs clickable
make "Related URLs" "clickable" on resource page. 

The should also be presented as a list:

- link1 ..
- link2 .. 

Don't validate links
Open links in a new tab

rschenk:
Closes #947

This simple addition led me to go on a refactoring bender.

- Added a [standardized way](https://api.rubyonrails.org/v6.0.0/classes/ActiveModel/Conversion.html#method-i-to_model) to un-decorate our decorators
- Refactored `WorkVersionMetadataComponent` and `CollectionMetadataComponent` to share a base class and most of the logic
- Removed `version_number` from the Work Version's metadata... @srerickson doesn't want it and I was in this code already.

Screen Shot 2021-03-29 at 12 05 15 PM


app/components/base_metadata_component.rb
```diff
@@ -0,0 +1,75 @@
...
+      return nil if value.empty?
+
+      list_items =
+        value
+          .map { |member| format_value(value: member, attr: attr) }
+          .compact
+          .map { |member| %(<li class="multiple-member">#{member}</li>) }
+          .join
+
+      %(<ol class="multiple-values">#{list_items}</ol>).html_safe
```
awead, line 60: I'm not crazy about including the html in here, but I understand this is probably the simplest path.

My original idea for dealing with formatting individual fields differently, was to make separate components for each of them. See d953b965acf03c2cadf8e9c74283cd8c174719ec

So, you could make `DisplayMultiValueComponent` and `DisplayLinkComponent` classes for your `format_multi_value` and `format_link` methods respectively, and then the `format_value` method gets a little less branchy. Each case of a doi, multiple field, or link field, would be handled by the `elsif value.is_a? ApplicationComponent` case.

You'd also have to define the respective methods on the decorators to assign each field to its given component.

However, this does make things a bit more complicated... certainly more files. So, if we want to pass on this and try another direction in the future, I'm okay with that.

app/decorators/resource_decorator.rb
```diff
@@ -10,6 +10,23 @@ def self.decorate(resource)
...
+  # Rails convention for "un-decorating" objects:
+  # https://api.rubyonrails.org/v6.0.0/classes/ActiveModel/Conversion.html#method-i-to_model
+  def to_model
+    undecorated = __getobj__
+
+    # It's possible to compose decorators on top of each other, so keep digging
+    # down the stack unitl we are no longer getting SimpleDelegators.
+    # Also define a max number of nesting so we don't get an infinite loop.
+    100.times do
+      break unless undecorated.respond_to?(:__getobj__)
```
awead, line 22: What if we raised an error here? If we're nesting decorators 100 levels deep, we've got bigger problems.

awead (approved):
Overall, this is great. I like the refactor! I've posed alternatives in two places. Let me know what you think.

app/components/base_metadata_component.rb
rschenk: The problem with doing that is, the items _within_ each multi-value field, could themselves be any number of datatypes which need to be formatted. The `format_multi_value` method recursively calls `format_value` on each of its items, so it needs to be able to call back into `BaseMetadataComponent.format_value`. Now, `format_value` is a pure function, so we could make that a class level method and allow other classes to call into it, but that seems like a major WTF in six months

awead: Ah, good point. Well, we may need to rethink this eventually. As different presentation options crop up, I don't want to end up putting all these different html formats in this component.

PR #954, opened 2021-03-29
Fix bug where work was not indexed when hitting cancel on form

### LINKED ISSUE #860: Cancel button prevents works from showing up on the dashboard.
# Bug Report

## Description
Clicking the "cancel" button during initial deposit (creating fist version of a work) causes the work to not show up in the dashboard even when the work version exists in draft state.

## To Reproduce
1. Go to the "create new work" form, fill out required fields on work details tab, and click "save and continue"
2. Click "Cancel" from the contributors tab. (You will be redirected to the work page, and see that the draft version exists.)
3. Go to your dashboard: you will not see the work because it wasn't indexed.

## Expected behavior
Two possibilities: 
- The draft version should appear on the dashboard (i.e., it should have been indexed)
- OR: The "cancel" could be changed to "delete draft", in which case the draft version should be deleted. (After step 2, above, the user would be directed to their dashboard).

rschenk:
Fixes #860

This was easy but made the work form feature tests explode, specifically the end-to-end ones, which were testing the indexing at every single step. 

I ended up removing some of these expectations from the "let's test the form end to end" specs, because they were already thoroughly tested in unit tests of each individual page on the form, and it was kind of confusing because the "this method was called X times" number keeps accumulating through the test. @awead I'll let you give the 👍 or 👎 to that.

### DIFF
--- a/spec/features/dashboard/collection_form_spec.rb
+++ b/spec/features/dashboard/collection_form_spec.rb
@@ -10,7 +10,20 @@ RSpec.describe 'Creating and editing collections', :inline_jobs, with_user: :use
   let(:new_collection) { Collection.last }
 
   before do
+    mock_solr_indexing_job
+  end
+
+  # When stepping through each page of the form, we want to test whether the
+  # collection was indexed, and how, based on the buttons that were pressed.
+  # Because there can be multiple of these steps within each test, and because
+  # RSpec mocks _cumulatively_ record the number of times they've been called,
+  # we need a way to say "from this exact point, you should have been called
+  # once." We accomplish this by tearing down the mock and setting it back up.
+  def mock_solr_indexing_job
+    RSpec::Mocks.space.proxy_for(SolrIndexingJob)&.reset
+
     allow(SolrIndexingJob).to receive(:perform_now).and_call_original
+    allow(SolrIndexingJob).to receive(:perform_later).and_call_original
   end
 
   context 'when creating a collection with only the required metadata' do
@@ -19,6 +32,7 @@ RSpec.describe 'Creating and editing collections', :inline_jobs, with_user: :use
 
       visit dashboard_form_collections_path
 
+      mock_solr_indexing_job
       FeatureHelpers::DashboardForm.fill_in_minimal_collection_details(metadata)
       FeatureHelpers::DashboardForm.save_and_exit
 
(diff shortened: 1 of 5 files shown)


awead (dismissed):

rschenk:
@awead please take a gander at `spec/features/dashboard/collection_form_spec.rb`, particularly the kind of strange helper method I wrote called `mock_solr_indexing_job` and see if it makes sense to you. The idea there was to be able to provide a clean slate around each button press, saying "ok, when you press this button, X job should have been called once". I needed this both because our factories hit the indexing job, and because we are testing multiple steps of the form within a single test, and I didn't want to accumulate the number of times  job had been called, e.g. "at this point you should have been called once in total, then at this next point you should have been called twice in total". I wanted to provide a clean slate for the cause and effect of, "this button = this action"

spec/features/dashboard/collection_form_spec.rb
```diff
@@ -10,7 +10,20 @@
...
+  end
+
+  # When stepping through each page of the form, we want to test whether the
+  # collection was indexed, and how, based on the buttons that were pressed.
+  # Because there can be multiple of these steps within each test, and because
+  # RSpec mocks _cumulatively_ record the number of times they've been called,
+  # we need a way to say "from this exact point, you should have been called
+  # once." We accomplish this by tearing down the mock and setting it back up.
+  def mock_solr_indexing_job
+    RSpec::Mocks.space.proxy_for(SolrIndexingJob)&.reset
```
awead, line 23: @rschenk OK, this makes sense. So basically any time this method is invoked, the count is reset and the expectation can be "this was called once" or "this thing was called at least once" or it's always going from zero to 1 or more.

spec/features/dashboard/work_version_form_spec.rb
```diff
@@ -68,7 +69,22 @@
...
+        visit dashboard_form_work_versions_path
+
+        FeatureHelpers::DashboardForm.fill_in_work_details(metadata)
+        FeatureHelpers::DashboardForm.save_and_continue
+
+        FeatureHelpers::DashboardForm.cancel
+
+        visit dashboard_root_path
+
+        expect(page).to have_content metadata[:title]
```
awead, line 87: So here there's no expectation check because we want it to actually index an check it in the dashboard?

awead (changes requested):
I think is basically fine. If you can double-check my questions, then I'll approve. I also didn't look _too_ closely at expect now versus later. I'd can pull down the branch and read through the whole test to see if we're expecting the right things at the right time.

spec/features/dashboard/work_version_form_spec.rb
rschenk: Yeah I figured here we should actually to a real blackbox end-to-end test to truly replicate the bug, rather than say "meh, that method was called so good enough ¯\_(ツ)_/¯"

spec/features/dashboard/collection_form_spec.rb
rschenk: Exactly.

Otherwise, the factories can alter the count, as can previous pages in the form if testing the entire form end-to-end. This always resets it to 0 so you can be sure that your exact action is affecting the method call counts

awead (approved):

PR #958, opened 2021-03-30
Reindex works when updating

### LINKED ISSUE #940: Can't add works that you have edit access to (but didn't create) to collections
# Bug Report

## Description
Users can only add works they have created to collections. Works they have *edit access to* (but didn't create) do not appear in the list of works to add to a collection on the collection edit form.

## To Reproduce
1. With admin privileges enabled, find a work you didn't create and give yourself edit permissions for the work.
2. Turn off admin privileges
3. Create a collection and try to add the work from step 1. (You can't)

## Expected behavior
You should be able to add works you didn't create but have edit access to.

### LINKED ISSUE #955: Inconsistency between access setting that appears on the catalog page and the work page.
On the catalog, a work appears to have Penn State-only access [[link](https://scholarsphere.psu.edu/catalog?search_field=all_fields&q=Acceptability+of+electronic+nicotine+delivery+systems+%28ENDS%29+among+HIV+positive+smokers)]. However, the work actually is public [[link](https://scholarsphere.psu.edu/resources/c2ad3853-93d2-44c2-9098-521972c4f6a6)]. Looks like the setting was updated without a re-index?

### LINKED ISSUE #957: Works are not reindex when they are updated
# Bug Report

## Description

When works are updated via the work settings page in `Dashboard::WorksController#update`, the work isn't reindexed after it is saved. This is the underlying issue with two other issues:

* Inconsistency between access setting that appears on the catalog page and the work page. #955 
* Can't add works that you have edit access to (but didn't create) to collections #940

awead:
Adds a callback to Dashboard::WorksController so that works are reindexed when they are updated. Note that the callback is not added to the Work model because this could cause unintended side effect when reindexing work versions, which also reindex their associated works.

The #destroy endpoint is removed from the controller as well since it is not used. Works are removed via calls to their work version.

Fixes #957
Fixes #955 
Fixes #940

### DIFF
--- a/app/controllers/dashboard/works_controller.rb
+++ b/app/controllers/dashboard/works_controller.rb
@@ -4,6 +4,8 @@ module Dashboard
   class WorksController < BaseController
     layout 'frontend'
 
+    after_action :update_index, only: [:update]
+
     def edit
       @undecorated_work = Work.find(params[:id])
       authorize(@undecorated_work)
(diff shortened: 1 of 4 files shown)


awead:
@rschenk Indexing is getting a lot of attention. I opted for the controller callback because I wasn't certain a model callback, like we're doing with WorkVersion, would cause a bunch of other problems. Ideally, we should just be able to call `save` and be done with it, but I can't see a solution the 100% problem-free ATM.

rschenk (approved):
Looks good to me

PR #961, opened 2021-03-31
Multiple fields' add button is at the bottom, not the top

### LINKED ISSUE #507: Switch the buttons for adding new fields
The (+) button should appear on the last (bottom-most) field instead of the first. This enables the user to add new fields starting with the last, most recent one.

rschenk:
Closes #507 

## Before
Screen Shot 2021-04-02 at 10 34 12 AM

## After
Screen Shot 2021-04-02 at 10 33 44 AM

### DIFF
--- a/app/javascript/controllers/multiple_fields_controller.js
+++ b/app/javascript/controllers/multiple_fields_controller.js
@@ -16,32 +16,68 @@
 import { Controller } from 'stimulus'
 
 export default class extends Controller {
+  buttonClass = 'js-mutiple-field-action'
+  addButtonClass = 'js-multiple-fields-action--add'
+  removeButtonClass = 'js-multiple-fields-action--remove'
+
   connect () {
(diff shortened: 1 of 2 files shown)


rschenk:
@awead You'll need to tell Code Climate to settle down

awead (changes requested):
This is stretching my limits in terms of JS comprehension. It looks like we're adding the buttons from the cloning process and then rearranging them? While the existing feature tests that exercise this aren't failing, I don't see tests that actually test what the controller is doing, or a feature the verifies the buttons are getting moved around to where we want them.

Additionally, I'm uncertain of any a11y issues that this may create. It may not, or actually fix some! I'd like some more time to look into this. For now, some testing might help ensure this lives on a bit longer. I've put in another PR with some Jest testing features, if you want to try to use that. See #969

awead (approved):

PR #963, opened 2021-03-31
Resource description supports markdown

### LINKED ISSUE #959: support paragraph breaks & links in work description
# Feature Request

The description should field should render paragraph breaks and links using markdown syntax.

A stripped-down markdown rendering of description text on the resource page. Only paragraph breaks and links should be supported. (User-entered html tags should be stripped-out or ignored).

* paragraph breaks (a blank line b/w text)
* links in the description are rendered on the resource page

## Implementation Details

Create a new markdown renderer for descriptions.  See `config/initializers/markdown_template_handler.rb`

rschenk:
I'm not sure if this is exactly what @awead had in mind regarding the decorators

Closes #959 

Screen Shot 2021-03-31 at 2 08 25 PM
Screen Shot 2021-03-31 at 2 08 45 PM

## Plain-texterization

`og:description` should maybe be plain text? There's a plain text method in the decorator to do this:

Screen Shot 2021-04-02 at 11 03 40 AM
---
Screen Shot 2021-04-02 at 11 03 46 AM
---
Screen Shot 2021-04-02 at 11 04 00 AM

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -47,6 +47,7 @@ Rails/SkipsModelValidations:
 Rails/OutputSafety:
   Exclude:
     - 'app/components/base_metadata_component.rb'
+    - 'app/decorators/resource_decorator.rb'
 
 RSpec/DescribeClass:
   Exclude:
--- a/app/decorators/resource_decorator.rb
+++ b/app/decorators/resource_decorator.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ResourceDecorator < SimpleDelegator
+  include ActionView::Helpers::SanitizeHelper
+
   # @returns an appropriately-decorated object. This is a factory method
   def self.decorate(resource)
     return WorkVersionDecorator.new(resource) if resource.is_a? WorkVersion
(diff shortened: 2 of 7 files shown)


awead (commented):
I'm hesitant to approve this for a couple of reasons.

1. Opening up user-entered data that we process in real time opens us up to some potential problems. Some bad data could kill the whole page. Maybe we should wrap this in a service, or private method that will rescue and display and error message to the user, give them the opportunity to fix it.
2. `description_html` is getting called twice for each page. Maybe memoize it?
3. I'm not sure we should return plain text at all. Stripping the html out could result in a meaningless description, such as bullet points w/out context. I'm more inclined to just return the markdown directly when we're not in the UI, and let the client decide what it wants to do with it.

I'd like to poke at this a bit more. We've been burned in the past with markdown rendering in SS3, and I want to make sure we're not setting ourselves up for the same problem.

rschenk:
@awead I just pushed up a new commit

> 1. Some bad data could kill the whole page

I trapped any errors and just return back the original string

> 2. Memoize

||= :done

> 3. just return markdown directly when we're not in the UI

The only place in the entire app where the plain text version is used is in the `og:metadata` header. Check out the screenshots in the description

awead (approved):

PR #968, opened 2021-04-01
API endpoint to return all the known DOIs

### LINKED ISSUE #967: API Endpoint for all DOIS
# Feature Request

RMD needs an API endpoint that returns a list of all DOIs of our resources (`doi` and `identifier` field). This is very similar to #930 and uses the same fields,  but instead of providing a DOI as a query param and returning back yes or no, it returns back a list of everything.

> my use case is that if I have a representation of an article in RMD with a publisher DOI, I’d like to be able to keep track of
most importantly - whether or not that DOI matches something in ScholarSphere
but also, exactly what thing(s) it matches in ScholarSphere
> We’re already doing that whenever someone deposits in ScholarSphere via RMD, but I’d like to do the same thing for stuff that’s already in ScholarSphere.
-- @EricDurante 

Eventually I don't think this will scale, but for now it will. 

```
GET /api/vi/dois
{
  "doi:10.1/123": ["uuid1", "uuid2"],
  "doi:10.1/456": ["uuid3"],
  ...
}
```

rschenk:
Closes #967

### DIFF
--- a/app/controllers/api/v1/dois_controller.rb
+++ b/app/controllers/api/v1/dois_controller.rb
@@ -2,6 +2,11 @@
 
 module Api::V1
   class DoisController < RestController
+    def index
+      results = DoiSearch.all
+      render json: results
+    end
+
     def show
       results = DoiSearch.new(doi: params[:doi]).results
       if results.any?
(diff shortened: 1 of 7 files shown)


awead (approved):
If this is good for @EricDurante, it's good for me.

PR #981, opened 2021-04-09
Index works async when saved

awead:
Works will always index into Solr when they are saved. This makes it consistent with work versions and collections.

### DIFF
--- a/app/controllers/dashboard/works_controller.rb
+++ b/app/controllers/dashboard/works_controller.rb
@@ -4,8 +4,6 @@ module Dashboard
   class WorksController < BaseController
     layout 'frontend'
 
-    after_action :update_index, only: [:update]
-
     def edit
       @undecorated_work = Work.find(params[:id])
       authorize(@undecorated_work)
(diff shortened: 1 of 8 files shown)


awead:
@whereismyjetpack My concern for this change was that saving a work version might trigger a save on a work, and thusly trigger multiple indexing jobs: work version save causes version and work indexing,  work saving also causes work indexing. I pushed this up and indexed some things while I watched the Sidekiq queue. I didn't see anything odd, like an extra proliferation of indexing jobs.... is there anything else you can think of?

whereismyjetpack:
we discussed this in chat. We think after comparing with qa we're 🏅

awead:
## Indexing Changes

The only consequence seems to be two additional work indexing jobs near the beginning of the workflow.

### Publishing a new work

#### Starting page

No changes.

#### Save and Continue to Contributors

4 work indexing jobs:

    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4 --> fails with record not found exception
    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4 --> fails with record not found exception
    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4
    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4

No further changes to the form from there on through publication.

### Creating a new version of a published work

#### Starting page

2 work indexing jobs:

    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4
    [ActiveJob] [SolrIndexingJob] gid://scholarsphere/Work/4
  
No further changes to the form from there on through publication.

whereismyjetpack:
This looks, and sounds good to me.. I think we should be good to roll with how it's written 🚀

awead:
I'm closing this in favor of #986. It's a smaller change and introduces less "noise" to the Sidekiq queue. It still leaves us open to work indexing problems in the future, but we can re-examine that when the time comes.

PR #983, opened 2021-04-12
terms of use page: added to page footer and about page; change column…

### LINKED ISSUE #826: Terms of use page
The terms of use page never made it over to ss4.0:

https://scholarsphere-3.libraries.psu.edu/terms

srerickson:
Fixes #826:

- A new Terms of Use page (based on SS3)
- Links to the new page on the "about" page and in the site footer
- A few small updates to other markdown pages
- Change the markdown layout so it's max-width instead of 100%


whereismyjetpack (commented):
@srerickson "ITS" isn't a unit at Penn State anymore. I think they prefer to be "Penn State IT"

app/views/layouts/markdown.html.erb
```diff
@@ -3,7 +3,7 @@
 <% end %>
 
 <% content_for :content do %>
-  <div class="container-fluid">
+  <div class="container">
```
awead, line 6: Switching to the fluid-less class may have implications for responsiveness. Did you test the layout at various screen sizes, from phone all the way up to full screen?

awead (changes requested):

app/views/layouts/markdown.html.erb
srerickson: I dragged the browser window width from the narrowest (mobile width) to the widest (two monitors wide) and everything looks the way I'd expect.

awead (approved):

PR #989, opened 2021-04-19
Analytics: use a custom formatting function to fix overlapping tick labels on analytics chart

### LINKED ISSUE #648: Fix overlapping month names in analytics plot
Maybe we should abbreviate months to fix overlapping text here: 

Screen Shot 2020-10-29 at 12 37 50 PM

nicktiberi:
Fixes #648.

This PR adds a custom `tickFormat` function to render the tick labels for the analytics chart. It fixes a rendering issue where in certain circumstances, full month names would overlap each other.

The formatting code is based on an example in the **d3-time-format** repo (https://github.com/d3/d3-time-format#d3-time-format) with a slight modification to abbreviate the month names.

**Before:**
image

**After:**
Screen Shot 2021-04-16 at 4 10 22 PM

### DIFF
--- a/app/javascript/frontend/scripts/view_statistics_chart.js
+++ b/app/javascript/frontend/scripts/view_statistics_chart.js
@@ -42,6 +42,7 @@ function drawChart (selection, data) {
     .ticks(5)
     .tickSizeOuter(0)
     .tickSizeInner(4)
+    .tickFormat(multiFormat)
 
   const line = d3.line()
     .x(d => x(d[0]))
@@ -149,4 +150,32 @@ function hideOverflowingTickLabels (svg, tickTextSelection) {
       return (tickLeft < svgLeft || svgRight < tickRight)
     })
     .style('visibility', 'hidden')
+}
+
+/**
+ * Custom multi-format override for the x-axis (time/date) tick labels.
+ *
+ * This is used to fix some rendering issues encountered when using the default tick formatter.
+ *
+ * Based on example from https://github.com/d3/d3-time-format#d3-time-format
+ * @param {Date} date
+ * @returns {String} Formatted string representation of the date
+ */
+export default function multiFormat (date) {
+  const formatMillisecond = d3.timeFormat('.%L')
+  const formatSecond = d3.timeFormat(':%S')
+  const formatMinute = d3.timeFormat('%I:%M')
+  const formatHour = d3.timeFormat('%I %p')
+  const formatDay = d3.timeFormat('%a %d')
+  const formatWeek = d3.timeFormat('%b %d')
+  const formatMonth = d3.timeFormat('%b')
+  const formatYear = d3.timeFormat('%Y')
+
+  return (d3.timeSecond(date) < date ? formatMillisecond
+    : d3.timeMinute(date) < date ? formatSecond
+      : d3.timeHour(date) < date ? formatMinute
+        : d3.timeDay(date) < date ? formatHour
+          : d3.timeMonth(date) < date ? (d3.timeWeek(date) < date ? formatDay : formatWeek)
+            : d3.timeYear(date) < date ? formatMonth
+              : formatYear)(date)
 }
--- a/spec/javascript/view_statistics_chart.test.js
+++ b/spec/javascript/view_statistics_chart.test.js
@@ -1,0 +1,21 @@
+import multiFormat from '../../app/javascript/frontend/scripts/view_statistics_chart'
+
+describe('ViewStatisticsChart', () => {
+  it('formats a day tick properly', () => {
+    const formattedDate = multiFormat(new Date('2021-04-13 00:00:00'))
+
+    expect(formattedDate).toBe('Tue 13')
+  })
+
+  it('formats a month tick properly', () => {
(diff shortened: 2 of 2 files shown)


app/javascript/frontend/scripts/view_statistics_chart.js
```diff
@@ -38,10 +38,33 @@ function drawChart (selection, data) {
...
+  const formatDay = d3.timeFormat('%a %d')
+  const formatWeek = d3.timeFormat('%b %d')
+  const formatMonth = d3.timeFormat('%b')
+  const formatYear = d3.timeFormat('%Y')
+
+  // Custom multi-format override for the x-axis (time/date) tick labels.
+  // This is used to fix some rendering issues encountered when using the default tick formatter.
+  // Based on example from https://github.com/d3/d3-time-format#d3-time-format
+  const multiFormat = function (date) {
+    return (d3.timeSecond(date) < date ? formatMillisecond
```
nicktiberi: @awead - I am guessing Code Climate doesn't like this function very much. 😁 It is kinda brutal to read.

Should I convert it to a conventional `if/else`?

```
const multiFormat = function (date) {
  if (d3.timeSecond(date) < date) {
	return formatMillisecond(date)
  } else if (d3.timeMinute(date) < date) {
	return formatSecond(date)
  } else if (d3.timeHour(date) < date) {
	return formatMinute(date)
  } else if (d3.timeDay(date) < date) {
	return formatHour(date)
  } else if (d3.timeMonth(date) < date) {
	if (d3.timeWeek(date) < date) {
	  return formatDay(date)
	} else {
	  return formatWeek(date)
	}
  } else if (d3.timeYear(date) < date) {
	return formatMonth(date)
  } else {
	return formatYear(date)
  }
}

Comment (awead) [reply app/javascript/frontend/scripts/view_statistics_chart.js]: @nicktiberi Actually, the original example isn't _too_ bad. It's basically just a big switch operation. If it were me, I'd probably pull the whole `multiFormat` thing out into its own function so that it was set apart (visually) from the rest of the `drawChart` function. See `findNearestDataPoint` for an comparison example. If it isn't clear from reading the funtion, I'll put in some YARD docs to explain what's coming in and out:

``` js
// @param [Date]
// @return [String] ??
// @notes Custom multi-format override for the x-axis (time/date) tick labels.
function multiformat (date) {
...
```

When this was written, we didn't have a testing framework setup, but we do now. Would you consider adding a jest test for this one function? Don't worry about this rest of the code... It's also possible that getting it strapped into a jest setup may be more trouble that it's worth.

awead: If CodeClimate still barks at us, we can just tell it to sit down.

nicktiberi: @awead - I think CodeClimate is still mad about the nested ternary operators, but it didn't like my refactored `if/else` either (too many lines in a function + too many return statements). So maybe nested ternary is OK?

Wrote a few tests for this which seem to cover the cases that we would encounter with the Scholarsphere analytics data, but please let me know if there are some more scenarios I should account for.

app/javascript/frontend/scripts/view_statistics_chart.js
```diff
@@ -150,3 +151,31 @@ function hideOverflowingTickLabels (svg, tickTextSelection) {
...
+ */
+export default function multiFormat (date) {
+  const formatMillisecond = d3.timeFormat('.%L')
+  const formatSecond = d3.timeFormat(':%S')
+  const formatMinute = d3.timeFormat('%I:%M')
+  const formatHour = d3.timeFormat('%I %p')
+  const formatDay = d3.timeFormat('%a %d')
+  const formatWeek = d3.timeFormat('%b %d')
+  const formatMonth = d3.timeFormat('%b')
+  const formatYear = d3.timeFormat('%Y')
```
awead, line 172: I wonder if we can just do:

``` js
const foo,
          bar,
          baz
```

but, you know, spaced correctly... github is being weird.

app/javascript/frontend/scripts/view_statistics_chart.js
```diff
@@ -150,3 +151,31 @@ function hideOverflowingTickLabels (svg, tickTextSelection) {
...
+
+  return (d3.timeSecond(date) < date ? formatMillisecond
+    : d3.timeMinute(date) < date ? formatSecond
+      : d3.timeHour(date) < date ? formatMinute
+        : d3.timeDay(date) < date ? formatHour
+          : d3.timeMonth(date) < date ? (d3.timeWeek(date) < date ? formatDay : formatWeek)
+            : d3.timeYear(date) < date ? formatMonth
+              : formatYear)(date)
+}
```
awead, line 181: yeah, this is fine. The alternatives make it less-readable, or more complicated than it needs to be. One nit-picky thing: can you line up the colons like it is in the example? I think that will help with readability.

awead (changes requested):
Awesome! yay tests! This is look so much better. I had just two really minor things. Otherwise, this is ready to go.

awead (approved):

PR #994, opened 2021-04-23
Release 4.6.1: Query for S3 etags using our Graphql endpoint

### LINKED ISSUE #995: View a file's etag from S3
# Feature Request

We need to be able to view the etag that was created for a file when it was uploaded to Amazon's S3 storage bucket.

Because this will be needed to verify file integrity for migrated files, it needs to be available in our API. Graphql makes the most sense because it's a simple query for information and adding the etag to the existing queries would be very easy.

We could to do a separate API endoint just for the etags, but this is more work.

awead:
In order to verify that files were migrated successfully, without corruption from Scholarsphere 3, we can check their integrity using the etag that Amazon calculated when it was uploaded into S3. The main advantage to doing it this way is that we don't have to calculate new checksums for each existing file right now.

To enable this, we first grant ExternalApp classes "admin" rights so they can perform file queries using graphql. Because querying for _any_ file using its old NOID could let someone view the metadata of a potentially private or embargoed file, we want to restrict this query to administrators only.

The graphql file query will return the etag for the file, which the client (i.e. Scholarsphere 3) can compare with another etag that it will calculate locally.

Fixes #995

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,7 @@ RSpec/LeakyConstantDeclaration:
     - 'spec/lib/healthchecks/queue_latency_check_spec.rb'
     - 'spec/system/active_job_system_spec.rb'
     - 'spec/models/concerns/all_dois_spec.rb'
+    - 'spec/models/concerns/null_object_pattern_spec.rb'
 
 RSpec/ExpectActual:
   Exclude:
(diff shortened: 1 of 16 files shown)


whereismyjetpack:
lgtm, I'll want to poke around at this a bit, but don't want to be a blocker for merging.

whereismyjetpack (dismissed):

awead:
@whereismyjetpack we can leave it on the branch for a little while. Unfortunately, we can't really test it until it's merged to production because we don't copy files from the bucket.

whereismyjetpack (approved):

PR #996, opened 2021-04-26
Creators search UX improvements

### LINKED ISSUE #891: Hitting enter on creator search box submits the form
Hitting enter when search for creators, submits the form (the page redirects to the work page).  

Note - this should be solved by moving help text to above the edit field. Screen reader user will know to arrow down for search results rather than hitting enter

nicktiberi:
Fixes #891.

- hitting the Enter key in the creators search field no longer submits the form
- the first search result is automatically selected

### DIFF
--- a/app/javascript/controllers/autocomplete_controller.js
+++ b/app/javascript/controllers/autocomplete_controller.js
@@ -7,13 +7,21 @@ export default class extends Controller {
   static targets = ['field', 'suggestionTemplate', 'emptyResultTemplate', 'lastOptionTemplate']
 
   connect () {
+    // Prevent Enter key from submitting the form
+    this.fieldTarget.addEventListener('keypress', function (e) {
+      if (e.key === 'Enter') {
+        e.preventDefault()
+      }
+    })
+
     const url = this.data.get('search')
     this.ac = autocomplete(
       this.fieldTarget,
       {
         hint: false,
         clearOnSelected: true,
-        openOnFocus: true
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/autocomplete_controller.js
```diff
@@ -7,13 +7,21 @@ export default class extends Controller {
   static targets = ['field', 'suggestionTemplate', 'emptyResultTemplate', 'lastOptionTemplate']
 
   connect () {
+    // Prevent Enter key from submitting the form
+    this.fieldTarget.addEventListener('keypress', function (e) {
+      if (e.key === 'Enter') {
+        e.preventDefault()
+      }
+    })
+
```
awead, line 16: @nicktiberi Is this somehow "scoped" to the autocomplete? If my cursor is focused someplace else on the page, this _shouldn't_ pick it up, right?

awead (changes requested):
Fantastic! One small question, otherwise, 👍

app/javascript/controllers/autocomplete_controller.js
nicktiberi: @awead  - yep, it is scoped to `fieldTarget` aka the creators search `<input>` tag, so it should not affect other fields on the page. If your cursor is focused on any other "normal" field, hitting Enter will still produce the default action of attempting to submit the form.

awead (approved):

PR #1003, opened 2021-05-03
Autopopulate display name when user adds a custom creator

### LINKED ISSUE #945: prepopulate display name with the search value from "add another creator" input
On the contributors form, if the creator isn't found through the "add another creator" input field and you select the "create a new creator" option, the display name field of the new creator entry should be prepopulated with the value from the search field.

nicktiberi:
Fixes #945.

In the creators search, if the user selects one of the options to create their own creator, prepopulate the custom creator's display name with the content of the user's search query.

Screen Shot 2021-05-03 at 3 50 28 PM

Screen Shot 2021-05-03 at 3 50 40 PM

### DIFF
--- a/app/javascript/controllers/autocomplete_controller.js
+++ b/app/javascript/controllers/autocomplete_controller.js
@@ -50,9 +50,11 @@ export default class extends Controller {
     return (q, callback) => {
       axios.get(url, { params: { q } }).then((response) => {
         const results = response.data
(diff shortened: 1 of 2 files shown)


awead (approved):
Wow. That was way easier than I thought it would be.

nicktiberi:
🙂

PR #1011, opened 2021-05-06
Update Dois when works are updated

### LINKED ISSUE #657: Update DOIs when works are published
Re-allow DOI minting for draft works and ensure that when the work is published, the DOI is update to reflect that.

awead:
The metadata associated with a doi in DataCite need to be updated when the work is updated in Scholarsphere.  For works, this only needs to happen when a new version is published, or an existing published version is updated via an admin. For collections, this will happen anytime the collection is updated.

In order to achieve this, we needed to stop supporting synchronous indexing. Due to the nature of the callbacks for both indexing and updating a doi, a synchronous index job was blocking an asynchronous update doi job. As a result, index job would somehow block the doi from getting the updated metadata from the database and the doi was updated with incorrect information.

Fixes #657

### DIFF
--- a/app/controllers/dashboard/form/base_controller.rb
+++ b/app/controllers/dashboard/form/base_controller.rb
@@ -80,6 +80,7 @@ module Dashboard
 
         def save_resource(index: true)
           @resource.indexing_source = null_indexer if !index
+          @resource.update_doi = (publish? || finish? || save_and_exit?)
           @resource.save
         end
 
--- a/app/controllers/dashboard/form/base_controller.rb
+++ b/app/controllers/dashboard/form/base_controller.rb
@@ -79,16 +79,7 @@ module Dashboard
         end
 
         def save_resource(index: true)
-          index_synchronously = (create? || publish? || finish? || save_and_exit?)
-
-          @resource.indexing_source = case
-                                      when !index
-                                        null_indexer
-                                      when index_synchronously
-                                        SolrIndexingJob.public_method(:perform_now)
-                                      else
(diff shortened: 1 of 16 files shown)


app/controllers/dashboard/form/base_controller.rb
```diff
@@ -79,16 +79,8 @@ def param_key
...
-
-          @resource.indexing_source = case
-                                      when !index
-                                        null_indexer
-                                      when index_synchronously
-                                        SolrIndexingJob.public_method(:perform_now)
-                                      else
-                                        SolrIndexingJob.public_method(:perform_later)
-                                      end
+          @resource.indexing_source = null_indexer if !index
```
nicktiberi, line 82: This is more a question for my own clarification, but `indexing_source` is either going to be `null_indexer` (basically a no-op) or `nil`, right? And if it's `nil`, then `SolrIndexingJob.public_method(:perform_later)` will be used as `indexing_source`. Am I understanding that correctly?

nicktiberi (approved):
I just have one question for my own knowledge, but otherwise this LGTM! 👍

app/controllers/dashboard/form/base_controller.rb
awead: Yes. This is we can _not_ index if we need to, and there's only one case where that's happening right now. In the publish controller, we want to save any changes to the work version, even if they still result in an unpublishable work: https://github.com/psu-libraries/scholarsphere/blob/develop/app/controllers/dashboard/form/publish_controller.rb#L19-L30

PR #1013, opened 2021-05-07
Add a public message for PSU-only works

### LINKED ISSUE #662: Public message for Penn State works
Display an information message to the user, similar to the embargo message, if a public user access a Penn State work. It should let them know they don't have access to the files.

## WDLL

On a resource page that has 'Penn State Only' access settings, a message should appear to unauthenticated users:

> Files are only accessible to users logged-in with a Penn State Access ID.

nicktiberi:
Fixes #662.

When a public user views a work that is PSU-only, a message appears indicating that the files are not viewable.

### Embargoed Work
image

### Embargoed, PSU-Only Work
image

### PSU-Only Work
Screen Shot 2021-05-07 at 1 45 19 PM


spec/components/files_visibility_detail_component_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        expect(content).to include('You have permission to access the files as an editor of the resource.')
+        expect(content).to include('Public link')
+      end
+    end
+
+    context 'when the work is PSU only and the user is not authorized' do
+      let(:work_version) { build(:work_version, :published, work: work) }
+      let(:work) { build(:work, visibility: Permissions::Visibility::AUTHORIZED) }
+
+      it 'displays a message' do
```
nicktiberi, line 58: @awead - this test is failing, probably because I'm not initializing the work and/or work version properly. Am I missing something obvious here?

spec/components/files_visibility_detail_component_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    specify do
+      expect(content).to be_empty
+    end
+
+    context 'when the work is PSU only and the user is not authorized' do
+      let(:work_version) { build(:work_version, :published, work: work) }
+      let(:work) { build(:work, visibility: Permissions::Visibility::AUTHORIZED) }
+
+      it 'displays a message' do
```
nicktiberi, line 78: @awead - this test is also failing, probably for the same reason as the test above.

spec/components/files_visibility_detail_component_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        expect(content).to include("Embargoed until #{embargo_date.strftime('%Y-%m-%d')}")
+        expect(content).to include('Files are not accessible during the embargo. After the embargo period, ')
+        expect(content).to include('files will be accessible to users logged-in with a Penn State Access ID.')
+      end
+    end
+  end
```
nicktiberi: @awead - this test is failing, probably because I'm not initializing the work and/or work version properly. Am I missing something obvious here?

spec/components/files_visibility_detail_component_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    context 'when the work is PSU only and the user is not authorized' do
+      let(:work_version) { build(:work_version, :published, work: work) }
+      let(:work) { build(:work, visibility: Permissions::Visibility::AUTHORIZED) }
+
+      it 'displays a message' do
+        expect(content).to include('Penn State Only')
+        expect(content).to include('Files are only accessible to users logged-in with a Penn State Access ID.')
+      end
+    end
```
nicktiberi: @awead - this test is also failing, probably for the same reason as the test above.

app/components/files_visibility_detail_component.rb
```diff
@@ -0,0 +1,37 @@
...
+      I18n.t('files_message.unauthorized.heading')
+    end
+  end
+
+  def i18n_key
+    if embargoed?
+      if unauthorized? then 'embargo_unauthorized' else 'embargo' end
+    elsif unauthorized?
+      'unauthorized'
+    end
```
awead, line 23: The nested `if` is a bit dodgy. I have two alternatives:

Snazzy:
``` ruby
    keys = Array.new                                                                                                       
    keys << 'embargo' if embargoed?                                                                                        
    keys << 'unauthorized' if unauthorized?                                                                                
    keys.join('_')
```

Safe:
``` ruby                                                                                                                                                                                                                  
    if unauthorized? && embargoed?                                                                                         
      'embargo_unauthorized'                                                                                               
    elsif embargoed?                                                                                                       
      'embargo'                                                                                                            
    elsif unauthorized?                                                                                                    
      'unauthorized'                                                                                                       
    end

File: app/views/shared/_work_version_files.html.erb [hunk 6]
```diff
@@ -1,4 +1,4 @@
-<%= render(EmbargoDetailComponent.new(work_version: work_version)) %>
+<%= render(FilesVisibilityDetailComponent.new(work_version: work_version)) %>
 <%= render(WithdrawnDetailComponent.new(work_version: work_version)) %>
```
awead, line 2: Were we combining the withdrawn component? I can't remember... if the work is withdrawn _and_ embargoed/PSU, we'd see two alert boxes. Is that what we want? I apologize if we already covered this.

awead (changes requested):

app/views/shared/_work_version_files.html.erb
nicktiberi: @awead - no worries! That was our original plan, but after talking with Seth in Slack on Friday, he preferred to keep it as two separate components and was OK with 2 alert boxes:

> Seth Erickson 9:50
> it might make sense to treat withdrawn in a separate component
> 
> Nick Tiberi  9:51 AM
> ok. the current implementation is that withdrawn and embargoed are 2 separate components, and both could possibly display at the same time. and there is no component for “penn state only”
> 
> Seth Erickson  9:59 AM
> right. so one component for penn state only & embargo, and separate component for withdrawn. This makes sense to me because they relate to slightly different things (works and versions, respectively)

awead: Cool, thanks!

app/components/files_visibility_detail_component.rb
nicktiberi: Yeah, I totally agree. IMO your "safe" option is the most easily readable out of the 3.

awead (approved):

PR #1016, opened 2021-05-10
Check for draft DOIs

### LINKED ISSUE #1015: DOI health check
We can check the status of the DOIs in Scholarsphere via our healthcheck system. Currently, this can just support checking for draft DOIs, which aren't allowed currently, but could be expanded later to check for out-of-sync DOIs or DOIs that either don't resolve or have other issues.

awead:
Custom OKComputer check that queries DataCite for any draft dois under the current client. A configurable threshold will allow a given number of drafts, in case there are some legitimate ones in the future.

Fixes #1015


lib/data_cite/client.rb
```diff
@@ -87,25 +91,28 @@ def put(doi:, data:)
...
+      end
+
+      def process_body(response)
         parsed_body = begin
                         JSON.parse(response.body)
                       rescue JSON::ParserError
                         {}
                       end
 
+        return parsed_body if response.success?
```
nicktiberi, line 107: I'm assuming that the `response` would have no `errors` in the response body if `success?` is true (i.e. it would only return a 200 status code if there are no errors, and if there were errors, it would throw some error status code and `success?` would be `false`).

nicktiberi (approved):
LGTM! 👍 

Thanks for adding me on these reviews. It's good to see how some of the moving pieces work here (DataCite integration, OkComputer health checks, etc).

lib/data_cite/client.rb
awead: Good point. I'm assuming that "success" is anything other than a non-200. I'm not 100% certain about that and just letting Faraday figure it out for me:

https://github.com/lostisland/faraday/blob/main/lib/faraday/response.rb#L73

But, we can just go with that. If there's something else, like a 404 or 500, we'd raise a DataCite::Error below. However, if the response is "success" but there some other kinds of errors, it's up to the calling service to figure that out. We haven't been too concerned with that. But, it does pose some problems. For example, on line 95 above, we're assuming that a successful response would have the data and id keys in it. It's possible they may not. I think in that case, `dig` would return null. It's probably some we ought to thing about.

PR #1018, opened 2021-05-11
Update webpacker, postcss, and is-svg

awead:
Addresses security vulnerabilities CVE-2021-28092 and CVE-2021-27290.


awead:
@nicktiberi is-svg and postcss needed to be updated. There doesn't seem to be a way to do this via just yarn.lock, and I put them in pack.json directly, even though these are transitive dependencies.

nicktiberi (approved):

PR #1019, opened 2021-05-12
Update the visibility badge to indicate that a work is embargoed

### LINKED ISSUE #441: badge for embargoed works in search results
When listing works on the browse/search, we should indicate if a work is under embargo or withdrawn. In both cases the page for the work is public (the metadata is public) but associated files are not accessible. It would make sense to show this status as a "badge" that replaces the visibility badge (open access, psu-only).

nicktiberi:
Fixes #441.

If a work is embargoed, we now display an embargoed badge instead of the normal access level badge.

Screen Shot 2021-05-13 at 10 49 05 AM

### DIFF
--- a/app/components/visibility_badge_component.rb
+++ b/app/components/visibility_badge_component.rb
@@ -7,35 +7,52 @@ class VisibilityBadgeComponent < ApplicationComponent
   end
 
   def render
-    work.respond_to?(:visibility)
+    work.respond_to?(:visibility) && work.respond_to?(:embargoed?)
   end
 
   private
 
     attr_reader :work
 
+    def label
+      details[details_key][:label]
+    end
+
+    def color
+      details[details_key][:color]
+    end
+
+    def icon
+      details[details_key][:icon]
+    end
+
     def html_class
       [
         'badge',
         'badge--icon',
-        "badge--icon-#{details[visibility][:color]}"
+        "badge--icon-#{color}"
       ].join(' ')
     end
 
-    def visibility
+    def tooltip
+      return unless work.embargoed?
+
+      I18n.t('visibility_badge_component.tooltip.embargoed', date: embargo_release_date)
+    end
+
+    def details_key
+      return 'embargoed' if work.embargoed?
+
       work.visibility || Permissions::Visibility::PRIVATE
-    end
-
-    def label
-      details[visibility][:label]
-    end
-
-    def icon
-      details[visibility][:icon]
     end
 
     def details
       {
+        'embargoed' => {
+          label: i18n_label('embargoed'),
+          color: 'red',
+          icon: 'lock_clock'
+        },
         Permissions::Visibility::OPEN => {
           label: i18n_label(Permissions::Visibility::OPEN),
           color: 'orange',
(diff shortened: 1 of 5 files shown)


app/decorators/resource_decorator.rb
```diff
@@ -67,6 +67,10 @@ def visibility_badge
     VisibilityBadgeComponent.new(work: self)
   end
 
+  def visibility_badge_mini
+    VisibilityBadgeComponent.new(work: self, mini: true)
+  end
+
```
nicktiberi: @awead - I don't love declaring a second method for the visibility badge, but wasn't sure if there was a way to pass the `mini` flag into the resource decorator and conditionally set it in the `visibility_badge` method above.

awead: I think this is stemming from the fact that we only want the embargo badge to show in search results, which are use the mini component? We should revisit this decision...

app/models/solr_document.rb
```diff
@@ -56,6 +56,16 @@ def deposited_at
...
 
+  def embargoed?
+    return false if self[:embargoed_until_dtsi].nil?
+
+    embargoed_until = Time.zone.parse(self[:embargoed_until_dtsi])
+
+    return false if embargoed_until.blank?
+
+    embargoed_until > Time.zone.now
+  end
```
awead, line 68: I think I would opt for two methods here to reduce the if-y-ness:

``` ruby
def embargoed_until
  Time.zone.parse(self.fetch(:embargoed_until_dtsi, ''))
rescue ArgumentError
end
```
The `fetch` method will take a default value. This is handy when dealing with null values in hashes. In this case, we can return an empty string, which Time will evaluate to nill. The rescue handles the other cases. Ex:

```
> Time.zone.parse('')
=> nil
> Time.zone.parse('asdf')
=> nil
> Time.zone.parse('2002')
ArgumentError: argument out of range
```

Then, our `embargoed?` method becomes a lot simpler:

``` ruby
def embargoed?
  return false if embargoed_until.nil?

   embargoed_until > Time.zone.now
end
```

app/components/visibility_badge_component.rb
```diff
@@ -2,36 +2,57 @@
...
     end
 
     def icon
-      details[visibility][:icon]
+      if show_embargoed?
+        'lock_clock'
+      else
+        details[visibility][:icon]
+      end
     end
```
awead, line 47: I think you should be able to keep icon, label, and html_class methods the same by using the visibility and details methods:

``` ruby
def visibility
  'embargo' if work.embargoed?

  work.visibility || Permissions::Visibility::PRIVATE 
end

def details
  {
    'embargo' => {
      label: i18n_label('embargoed'),
      color: 'red',
      icon: 'lock_clock'
...
```

We are conflating visibility and embargo here, which is probably ok within the context of just this one component. However, I could also renaming the visibility method to `details_key` or just `key` for clarity.

awead (changes requested):
We can pair over these if it's confusing.

app/decorators/resource_decorator.rb
nicktiberi: Per our conversation with Seth in standup, we are going to show the embargoed badge everywhere and not make a distinction between the mini and non-mini version.

nicktiberi:
@awead - I forgot to add the tooltip. 🤦 Please hold off on re-reviewing until I get that in. Should be a pretty quick change.

awead (dismissed):

nicktiberi:
@awead - OK - ready to go now, tooltip and all!

awead (approved):
Nicely done!

PR #1020, opened 2021-05-12
Add Publisher's Statement

### LINKED ISSUE #960: New field "statement_required"
# Feature Request

Publishers often require the inclusion of a "set statement" when authors upload pre-publication versions of articles to institutional repositories (i.e., for new OA policy). Currently, the only way to include a set statement is to put it in the description, which may cause confusion in the long run. We should create a new field that is part of the work version's meta block.

- a new field called `publisher_statement`

- just adding the set statement to the description

- A a field (`publisher_statement`, with label: "Publisher Statement") that is part of the work version's `meta` block. 
- The field input should appear under "optional metadata" in the deposit form
- <s>Change deposit validations so that if a publisher's statement is included, the description is not required.</s>

awead:
Adds a publishers_statement field to the work version, and moves the keyword field down to "Optional Metadata" because it was never required before.

Fixes #960

### DIFF
--- a/app/decorators/resource_decorator.rb
+++ b/app/decorators/resource_decorator.rb
@@ -34,10 +34,9 @@ class ResourceDecorator < SimpleDelegator
   end
 
   def description_html
-    return unless respond_to?(:description)
-    return '' if description.blank?
+    return '' if combined_description.blank?
 
-    @description_html ||= render_markdown(description)
+    @description_html ||= render_markdown(combined_description)
   end
 
   def description_plain_text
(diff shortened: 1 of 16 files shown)


app/decorators/resource_decorator.rb
```diff
@@ -77,6 +76,10 @@ def first_creators
 
   private
 
+    def combined_description
+      [try(:description), try(:publisher_statement)].join(' ')
```
nicktiberi, line 80: I wonder if we should put a couple line breaks before the `publisher_statement` if it is preceded by the `description`. But that is probably a decision for Seth to make after he plays around with it.

nicktiberi (approved):
LGTM!

app/decorators/resource_decorator.rb
awead: I mentioned that to him, and he seemed ok with the tacking it together. We'll see at retro...

PR #1027, opened 2021-05-14
Update Rails and related gems

awead:
Updates us to the latest compatible version of Rails, which is 6.1.


Gemfile
```diff
@@ -80,7 +78,6 @@ group :test do
   gem 'rspec'
   gem 'rspec-its'
   gem 'rspec-rails', '~> 4.0.0.beta3'
-  gem 'scholarsphere-client'
```
nicktiberi, line 83: Is `scholarsphere-client` now deprecated in favor of the API methods now built into Scholarsphere?

Did that need to be done in order to upgrade Rails to 6.1? I guess I'm not sure why the file upload process was changed in an upgrade Rails commit.

awead: @nicktiberi I should have put this is a separate commit... It's tangentially related. I'll push up a new commit.

nicktiberi (approved):
Looks good! 🎉

Gemfile
awead: @nicktiberi I pushed up a different commit, and combined one of the previous ones. The `marcel` gem went to 1.0, which blocked the Rails update, but also required a new release of scholarsphere-client. Scholarsphere was using an older version. Updating the client gem caused the one ingest controller test to break because we were using the presigned url business to upload files. Prior to this, we were uploading files to S3 basically by sharing all of the access keys, which isn't a good idea security-wise.

Anyway, the only reason we were using the gem in this case was to craft a POST request to the controller for testing. It didn't seem worthy to go through all the trouble of re-hashing uploads with the new client gem version  just to test this one controller. So I decided to just remove the gem.

PR #1034, opened 2021-05-18
Home page SEO improvements

### LINKED ISSUE #720: Meta Description, Link Descriptive Text
I was running lighthouse against ScholarSphere, and it flagged 2 issues that seem pretty quick to fix. 

1.) Document does not have meta description
https://web.dev/meta-description/?utm_source=lighthouse&utm_medium=devtools

2.) "read more" /about link does not have descriptive text
https://web.dev/link-text/?utm_source=lighthouse&utm_medium=devtools

## WDLL 

Add description meta tags to the home page.

nicktiberi:
Fixes #720.

- meta description added to home page
- "read more" link is now more descriptive

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -1,3 +1,7 @@
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('home.intro.content') %>">
+<% end %>
+
 <% content_for :navbar_heading do %>
   <h1 class="navbar-brand navbar-brand--center slab-font"><%= t('navbar.heading.home') %></h1>
 <% end %>
@@ -8,7 +12,7 @@
       <h2 class="h4 mb-2"><%= t('home.intro.heading') %></h2>
       <div class="surface">
         <%= t('home.intro.content') %>
-        <%= link_to 'read more', about_path %>
+        <%= link_to t('home.intro.read_more'), about_path %>.
       </div>
     </div>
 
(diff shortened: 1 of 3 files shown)


app/views/layouts/frontend.html.erb
```diff
@@ -8,6 +8,7 @@
     <!-- Internet Explorer use the highest version available -->
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <%= yield(:meta_tags) if content_for? :meta_tags %>
+    <%= render partial: 'resources/home_meta_tags' if current_page?('/') -%>
```
awead: @nicktiberi On line 10, there's a yield, which is basically saying, if you have a content block that you've called `meta_tags`, I'm going to let it through right here. You can read up on on it if you like https://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method

So, what you could do here is edit `app/views/pages/home.html.erb` and add something like:

``` ruby
<% content_for :meta_tags do %>
  <meta name="description" content="<%= t('home.intro.content') %>">
<% end %>
```

And that ought to do the trick. If this doesn't make sense, let me know and we can pair up on it.

awead (changes requested):

app/views/layouts/frontend.html.erb
nicktiberi: Learning something new about Rails every day. 🤓 Thanks!

awead (approved):
👏

awead:
@nicktiberi could you verify with the lighthouse site that this fixes the problem?

nicktiberi:
@awead - yep, I verified with Chrome devtools that our SEO score is now 💯 :

Screen Shot 2021-05-18 at 3 03 15 PM

PR #1037, opened 2021-05-18
Change indexing of work and work versions

### LINKED ISSUE #765: Work / WorkVersion Link Confusion
Some users are stumbling into a problem where they are using and circulating links to a _work version_ when they mean to be using links to _the work_. One reason this happens is that the links used on the dashboard are version links but the links on the search page are work links.

I'd like to make it harder for people to accidentally use version links when they probably want a work link. To that end, can we make the links used on the dashboard work level links?

### LINKED ISSUE #1010: Changes to published/draft status badge in search results
# Feature Request

In search results and on the dashboard, each entry has a "badge" indicating the work versions's published/draft status: 
badge

Currently, this badge reflects the state of the most recent work version in a work. This feature requests changes the badge behavior to reflect the state of the parent work instead of the individual work version:

- An entry should appear as "published" if the parent work has at least one published version.
- An entry should appear as "withdrawn" if the parent work has no published versions and at least one withdrawn version.
- An entry should appear as "draft" if the parent work only has versions in draft state. (Because a work may only have one draft version at a time, there should be only one version in this case).

Relates to discussion in #441 
Blocked by #1026

### LINKED ISSUE #1026: Changes to indexing of work and work versions
Presently, the public and dashboard searches are driven by a distinction between work and work version. Works appear in the public search, while work versions appear in the dashboard. The original impetus behind this was so a user could see what drafts they could edit. Because editing and creating drafts has now part of the resource page, it's time to clear up the work/work version distinction.

Related to #1010 
Related to #765 

## Proposed Change

Index works (even in draft state) and return works in _both_ the public and dashboard searches. What gets returned where depends on the status of the work:

|Status|Public Search|Dashboard Search|
 --- | --- | --- |
| Published work | ✅ | ✅ |
| Draft work | ❌ |  ✅ |
| Withdrawn work | ❌ |  ✅ |

The criteria for each works' status is outlined in #1010.

## WDLL

* delete or rename LatestPublishedVersionSchema to select the relevant work version to index for a given work:
    * draft work = current draft version
    * withdrawn work = latest withdrawn version
    * published work = latest published version
* update SearchBuilder to return only works that are in `published` state
* update Dashboard::SearchBuilder to return all works, regardless of state, but ensure they're limited to works the user can edit
* update WorkIndexer
    * works are _always_ indexed
    * each work version is _always_ indexed
    * don't remove withdrawn works from the index 

## Notes

* admin status would lift any restrictions, making both public and dashboard searches effectively the same (although facets are still configured differently)
* work versions will _still_ be indexed for now

awead:
Both public and dashboard search interfaces now return only Work resources for search results. All works are now indexed, regardless of what kinds of versions they have. Any given work will have a "state" that is determined by its versions:

* published work = work with any published version
* withdrawn work = work with any withdrawn versions and no published versions
* draft work = work with no published or withdrawn versions

All the relevant search builder models were updated to reflect this, as well as changes to work policies so that draft and
withdrawn works will resolve for users who have access to them.

Fixes #1026
Fixes #1010 
Fixes #765


app/models/dashboard/search_builder.rb
```diff
@@ -1,81 +1,17 @@
 # frozen_string_literal: true
 
+# @abstract Builds a search query of all resource the user has edit access to. This is used to power the dashboard
+# search interface. Draft, emabargoed, and withdrawn works are eligible if the has edit access to them.
```
nicktiberi: Minor typos in comment:

```suggestion
# @abstract Builds a search query of all resources the user has edit access to. This is used to power the dashboard
# search interface. Draft, embargoed, and withdrawn works are eligible if the user has edit access to them.
```

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -25,7 +25,8 @@ def update
         # save again--this time using the published validations. That way the
         # appropriate error messages will appear on the form when it's re-rendered
         if publish?
-          save_resource(index: false)
+          @resource.indexing_source = Proc.new { nil }
+          @resource.save
```
nicktiberi, line 29: What is the significance of this change? Looks like it now skips the `@resource.update_doi` stuff from `save_resource` now.

nicktiberi (changes requested):
Some of the specifics of what is going on here are a little hazy to me, but overall this makes sense and LGTM. Just a question and a minor comment.

app/controllers/dashboard/form/publish_controller.rb
awead: This is the bug you were noticing earlier. We save the resource twice. The first time is to catch validation errors that occur during publishing, and to make sure that changes are saved to the database, even if it can't be published. There's a comment above this method that explains that, it's just not showing up in the diff.

I'm not sure why this wasn't a problem earlier, but the `Proc.new { nil }` is so that the work version is not indexed during this initial save-check.

nicktiberi (approved):

PR #1042, opened 2021-05-19
Always show works and work versions

### LINKED ISSUE #970: Draft works' urls are not resolving for the public
# Bug Report

## Description

If a work has only one version that is in draft, its url does not resolve to the public.

## To Reproduce
1. Create a work
2. Save as draft
3. Visit the url in a non-authenticated session

## Expected behavior

It should be a 200.

### LINKED ISSUE #1022: Work's withdrawn version should not 404
# Bug Report

## Description

If the work url points to a withdrawn version, it should not 404.

## To Reproduce
1. publish a new work
2. withdrawn the only version
3. visit the work resource url

## Expected behavior

Display the work's metadata. It should be as if it's a restricted resource. We should not be able to see or download files, and see the withdrawn message.

awead:
Draft works and withdrawn works should always resolve. However, their files are still inaccessible, and draft works are not discoverable when navigating existing versions in the resource page.

This ensures that any user who has a url for a draft or withdrawn work will be able to see it, but not discover it via search or browsing other existing versions.

Fixes #1022 
Fixes #970

### DIFF
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -8,7 +8,7 @@ class WorkPolicy < ApplicationPolicy
   end
 
   def show?
-    published? || editable?
+    true
   end
 
   def edit?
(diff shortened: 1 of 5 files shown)


nicktiberi (approved):
LGTM! 👌

PR #1043, opened 2021-05-20
Allow creation of a new work version from a withdrawn work version

### LINKED ISSUE #1021: Create a new version from a withdrawn one
# Bug Report

## Description

If the latest version is withdrawn, we can't create a new version.

## To Reproduce
1. Publish new work
2. withdraw published version
3. goto display page

## Expected behavior

Create new version link in the header should be available.

nicktiberi:
Fixes #1021.

In the case where a work only has 1 version and it is withdrawn, that version is considered the "representative version". The user can now create a new work version from that withdrawn version.

Previously, if there was only 1 work version and it was withdrawn, the user would be unable to create any new versions of that work.

### DIFF
--- a/app/policies/work_version_policy.rb
+++ b/app/policies/work_version_policy.rb
@@ -51,7 +51,7 @@ class WorkVersionPolicy < ApplicationPolicy
 
   def new?
     Pundit.policy(user, record.work).create_version? &&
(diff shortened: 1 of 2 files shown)


awead (approved):
NOICE!!!!

Issue #1059, opened 2021-05-28
Message for withdrawn versions that newer version exists


srerickson:
# Feature Request

When viewing a __withdrawn__ work version for a work with a more recent published version, a message should appear at the top of the work:

> This version has been withdrawn. [View the current version](*link to work*).

srerickson:
can this be closed?

ajkiessl:
@srerickson I think this might be completed already considering what Adam posted here: https://github.com/psu-libraries/scholarsphere/pull/1061 .

However, I'm having trouble recreating a "Withdrawn work version" scenario to confirm.  The only option I can find as an admin is to delete the most recent version.

ajkiessl:
Okay, I just messed with some things in the backend locally to reproduce.  Just like what Adam shows in #1061, this is how a withdrawn version with a newer published version looks:

![Screen Shot 2022-01-27 at 2 52 50 PM](https://user-images.githubusercontent.com/32677188/151433950-66361c7e-59a1-4e35-bfb0-26dac13537c4.png)

I think this is probably sufficient.

srerickson:
Sorry @ajkiessl I missed my own note to check if this can be closed -- I think it can.

PR #1061, opened 2021-06-01
WorkVersionDetailComponent

### LINKED ISSUE #1058: Message to edit draft when viewing work
# Feature Request

When a logged-in user views a *published work* that they have edit access to, and a draft version for the work exists, the user should see a message with links to the draft:

> An updated draft version for this work is [available](*link to draft*). You can access the draft version using the version selection menu above.

Note, this is a short-term fix. Ultimately, we should replace this message with buttons in the blue bar that takes users to the draft version form.

awead:
Replaces in the in-view logic with a separate component to determine messages displayed to the user. If the user can edit the work, and a draft exists, then the message will give them a link to it.

If the user does not have edit rights, the message only appears if another published version exists.

Fixes #1058

### DIFF
--- a/app/components/work_version_detail_component.rb
+++ b/app/components/work_version_detail_component.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+class WorkVersionDetailComponent < ApplicationComponent
+  attr_reader :work_version
+
+  def initialize(work_version:)
+    @work_version = work_version
+  end
+
+  def render?
+    !current_draft_version? && i18n_key.present?
+  end
+
+  def i18n_key
+    if draft_version.present?
+      'draft_version'
+    elsif representative_version.published? && (work_version.uuid != representative_version.uuid)
+      'old_version'
+    end
+  end
+
+  def linked_version
+    if draft_version.present?
+      work_version.work.draft_version.uuid
+    elsif representative_version.published?
+      work_version.work.uuid
+    end
+  end
+
+  private
+
+    def representative_version
+      @representative_version ||= work_version.work.representative_version
+    end
+
+    def draft_version
+      @draft_version ||= begin
+                           unless Pundit.policy(controller.current_user, work_version.work).edit?
+                             return NullWorkVersion.new
+                           end
+
+                           work_version.work.draft_version || NullWorkVersion.new
+                         end
+    end
+
+    def current_draft_version?
+      work_version.uuid == draft_version.uuid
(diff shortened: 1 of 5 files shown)


awead:
There's some conflict with #1059 that may render it moot.

## Examples

Published work with available draft

Screen Shot 2021-06-01 at 3 31 20 PM

Withdrawn work with available  draft

Screen Shot 2021-06-01 at 3 31 44 PM

Public work with a more recent version

Screen Shot 2021-06-01 at 3 42 52 PM

Withdrawn version with a current published one

Screen Shot 2021-06-01 at 3 43 02 PM

app/views/resources/_work_version.html.erb
```diff
@@ -47,14 +47,7 @@
...
 
-  <% unless (work_version.latest_published_version? || work_version.draft? || work_version.withdrawn?) %>
-    <div class="alert alert-warning" role="alert">
-      <%= t 'resources.old_version.message' %>
-      <%= link_to t('resources.old_version.link'),
-                  resource_path(work_version.work.uuid),
-                  class: 'alert-link' %>
-    </div>
-  <% end %>
+  <%= render WorkVersionDetailComponent.new(work_version: work_version) %>
```
nicktiberi, line 50: Nice cleanup! Make everything into components! 🙂

app/components/work_version_detail_component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def initialize(work_version:)
+    @work_version = work_version
+  end
+
+  def render?
+    return false if current_draft_version?
+
+    i18n_key.present?
+  end
```
nicktiberi: Is this a Ruby way of writing this? Just curious. I think more typically I would see something like:

```
def render?
  !current_draft_version? && i18n_key.present?
end
```

Obviously what you have works; I am mainly just wondering if the reason you wrote it that way is due to some convention. Thanks!

nicktiberi (dismissed):
LGTM! Just have one quick question.

srerickson:
@awead - this makes sense to me! Thanks!

app/components/work_version_detail_component.rb
awead: @nicktiberi Yeah, that's probably better. This went through a lot of iterations and I wrote it that way so it would look really obvious to me.

nicktiberi (approved):

PR #1063, opened 2021-06-03
Force UTF-8 encoding for extracted text files

### LINKED ISSUE #1055: Encoding::UndefinedConversionError when indexing extracted text
# Bug Report

## Description

Certain pdfs will throw a Encoding::UndefinedConversionError when attempting to index their extracted text into Solr. Message will look something like:

    Traceback (most recent call last):
            3: from (irb):18
            2: from app/models/file_resource.rb:50:in `update_index'
            1: from app/services/indexing_service.rb:9:in `add_document'
    Encoding::UndefinedConversionError ("\xE2" from ASCII-8BIT to UTF-8)

## To Reproduce

1. Create a work with `Recommendations for Supporting ORCID in Repository Platforms FINAL (4).pdf` as the file
2. Attempt to update the index of the file in the console

## Expected behavior

The file should index successfully without any errors.

## To Fix

We can probably just ignore or strip out the incompatible ASCII-8BIT characters or attempt a better UTF-8 conversion. It's also possible Solr might be misconfigured?

awead:
Shrine is downloading extracted text derivatives as ASCII-8BIT, but we need them to be UTF-8 so that they'll index properly. Because MetadataListener is created the files, we can reasonably assume that they'll almost always be UTF-8.

Several of the existing file helper classes were refactored and combined into a new FileHelpers class for better organization in the spec tests.

Fixes #1055


nicktiberi (approved):
Nice! LGTM 👌

PR #1067, opened 2021-06-07
Validate format of DOIs in the doi field

### LINKED ISSUE #1035: DOI problems with DataCite
Since 4.7, we can update existing DOIs in DataCite, however there are discrepancies with the existing ones in Scholarsphere versus those in DataCite.

### DOI has already been taken

Ex: https://doi.org/10.26207/agge-kh83

``` ruby
> resource = FindResource.call('c227dd00-a2e6-46c4-82e6-44aed7864a02')
> metadata = DataCite::Metadata::WorkVersion.new(resource: resource.latest_published_version, public_identifier: resource.uuid)
> client.update(doi: resource.doi, metadata: metadata.attributes)
DataCite::Client::Error ([{"source"=>"doi", "uid"=>"10.26207/agge-kh83", "title"=>"This DOI has already been taken"}])
```

### 403 Forbidden

This seems to only affect those with the 10.18113 prefix, which were migrated. Also, they appear to have links to work version URLs and not works.

Example: https://doi.org/10.18113/s1ks3v

``` ruby
> resource = FindResource.call('d660e2d9-2650-4cc6-a6e0-8e923fd2ba7e')
> resource.class
=> WorkVersion
> metadata = DataCite::Metadata::WorkVersion.new(resource: resource, public_identifier: resource.work.uuid)
> client.update(doi: resource.work.doi, metadata: metadata.attributes)
DataCite::Client::Error ([{"status"=>"403", "title"=>"You are not authorized to access this resource."}])
```

## Discrepancies

DataCite lists 546 DOIs. These do not mach 957 in Scholarsphere.

| Prefix | DataCite | Scholarsphere |
| ------ | --------  | --------------|
| 10.26207 | 401 | 350 |
| 10.18113 |  145 | 607 |

awead:
Models that have the :doi field are required to have dois that are valid, managed, and formatted so that they can be updated by the Datacite API. Previously, any valid and manged DOI would suffice, but the API was unable to interpret DOIs with "doi:" or "http:" URIs, so now the required formate is simply "[prefix]/[suffix]".

Fixes #1035

### DIFF
--- a/app/models/doi.rb
+++ b/app/models/doi.rb
@@ -6,6 +6,14 @@ class Doi
 
   MANAGED_PREFIXES = ['10.26207', '10.18113'].freeze
 
+  # @note If the DOI is valid from Datacite's point of view, it will be valid from our perspective, as well as have one
+  # of our managed prefixes, _and_ have a particular format that the Datacite API can interpret.
+  def self.valid_datacite_doi?(doi)
+    formatted_doi = new(doi)
+
+    formatted_doi.valid? && formatted_doi.managed? && doi == formatted_doi.to_datacite
+  end
+
   # @param [String] doi
   def initialize(doi)
     @doi = begin
@@ -39,6 +47,10 @@ class Doi
 
   def uri
     URI("https://doi.org/#{prefix}/#{suffix}")
+  end
+
+  def to_datacite
+    "#{prefix}/#{suffix}"
   end
 
   private
--- a/app/models/concerns/all_dois.rb
+++ b/app/models/concerns/all_dois.rb
@@ -16,6 +16,11 @@ module AllDois
       symbolized_args = args.map(&:to_sym)
       define_method(:fields_with_dois) { symbolized_args }
     end
+
+    # Validates the Datacite DOI field to ensure it's in the correct format for the API
+    validates :doi,
+              datacite_doi: true,
+              unless: -> { doi.blank? }
   end
 
   # Returns an arary of all valid dois found within the methods specified by the
--- a/app/validators/datacite_doi_validator.rb
+++ b/app/validators/datacite_doi_validator.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class DataciteDoiValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if Doi.valid_datacite_doi?(value)
+
+    record.errors.add(attribute, (options[:message] || :invalid_doi))
+  end
+end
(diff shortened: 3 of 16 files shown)


app/models/doi.rb
```diff
@@ -6,6 +6,14 @@ class Doi
...
   MANAGED_PREFIXES = ['10.26207', '10.18113'].freeze
 
+  # @note If the DOI is valid from Datacite's point of view, it will be valid from our perspective, as well as have one
+  # of our managed prefixes, _and_ have a particular format that the Datacite API can interpret.
+  def self.from_datacite?(doi)
+    formatted_doi = new(doi)
+
+    formatted_doi.valid? && formatted_doi.managed? && doi == "#{formatted_doi.prefix}/#{formatted_doi.suffix}"
+  end
+
```
nicktiberi: @awead - this is probably just a case of me not understanding the full context, but why is this written as a static method rather than an instance method (sorry; prob wrong terminology for Ruby) such as `valid?`, `managed?` etc?

Also, maybe it makes sense to call this something like `datacite_valid?` or something. The name `from_datacite?` to me implies that this method indicates the DOI originated from Datacite, and not that the DOI is valid from Datacite's perspective.

If I'm understanding correctly, all the DOIs in the system should now have this new format that is valid in Datacite, but that does not necessarily mean that all of the DOIs were created by Datacite.

Again, I may be super wrong about some of these assumptions haha. I don't have a super strong opinion on this, but wanted it to throw it out there for your consideration.

awead: No, those are good questions. 

To answer the first question, the `from_datacite?` method is a class method. So we'd call is like:

``` ruby
Doi.from_datacite?('doi')`
```

Within the context of the method, we're in the Doi _class_ scope. To get to instance methods, we create an _instance_ of the Doi class (formatted_doi), and then call instance-level methods on it. If we just called `valid?` it would assume the class scope and look for `Doi.valid?`.

Now to your second question... `from_datacite` is misleading. I was originally thinking the "from" part is because that's what is returned from the API. But, that's isn't very clear. Here's what I propose:

An new instance method:

``` ruby
def to_datacite
  "#{prefix}/#{suffix}"
end
```

And rename the class method to:

``` ruby 
def valid_datacite_doi?(doi)
  formatted_doi = new(doi)

  formatted_doi.valid? && formatted_doi.managed? && doi == formatted_doi.to_datacite
end
```

nicktiberi: Sounds good to me! I think `valid_datacite_doi?` makes its purpose much clearer.

nicktiberi (approved):

PR #1068, opened 2021-06-07
Authorize the user to destroy any collection

### LINKED ISSUE #1045: Admins can't delete collections
# Bug Report

Admins can't delete collection

## Description
When an admin tries to delete a collection they didn't create, it doesn't work.

## To Reproduce
1. Enable admin privs on QA server
2. On QA, find a collection you didn't create
3. Try to delete it.

awead:
Users may have rights to destroy a collection they don't own, such as an admin or another user with edit access. We use Pundit policies to check this, just as we are with the other actions in the controller.

Fixes #1045


nicktiberi (approved):
Nice fix!

PR #1069, opened 2021-06-07
Users should not edit withdrawn works

### LINKED ISSUE #1053: Withdrawn versions are editable
# Bug Report

## Description
Withdrawn work versions are editable. They shouldn't be.

## To Reproduce
1. Find or create a withdrawn work version and open it.
2. The edit button is active

## Expected behavior
- The edit button shouldn't be active
- Withdrawn versions should behave like published versions with respect to edit permissions: admins can edit, but users can't.

awead:
An oversight in the WorkVersionPolicy was allowing users edit access to their withdrawn works. Now, users cannot edit a withdrawn work; however, admins still can.

Fixes #1053

### DIFF
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -14,10 +14,7 @@ class WorkPolicy < ApplicationPolicy
   def edit?
     editable?
   end
-
-  def update?
-    edit?
-  end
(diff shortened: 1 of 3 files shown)


nicktiberi (approved):
🆒

PR #1070, opened 2021-06-08
Return the deposited year for uncertain dates

### LINKED ISSUE #1038: Valid EDTF Publication Date Causes Server Error
# Bug Report

## Description

The publication date field must be EDTF but some EDTF dates cause a server error when saving/rendering. For example, the date "2010/2020" is valid but causes error:

```ruby
    error: ActionView::Template::Error: undefined method `year' for #<EDTF::Interval:0x00007fa7651a6800> 
```

## To Reproduce
1. Create a new work
2. Set the publication date to an interval "2010/2020"
3. Try to publish the work

## Expected behavior
Valid EDTF dates shouldn't cause a server error.

awead:
If EDTF can't determine a year for a given date, then return the deposited year. Valid EDTF dates, such as intervals like `2010/2020` don't have exact years, so we won't attempt to guess one.

Fixes #1038

### DIFF
--- a/app/components/google_scholar_metadata_component.rb
+++ b/app/components/google_scholar_metadata_component.rb
@@ -28,7 +28,7 @@ class GoogleScholarMetadataComponent < ApplicationComponent
   def citation_publication_date
     return deposited_at.year unless EdtfDate.valid?(published_date)
 
-    Date.edtf(published_date).year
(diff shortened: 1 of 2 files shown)


nicktiberi (approved):
LGTM! 🥇

PR #1073, opened 2021-06-11
Create new versions from the representative

### LINKED ISSUE #1054: Can't create new version from withdrawn version
# Bug Report

## Description

If a work has a version history like this:
- v1: published
- v2: withrdawn

 ... when creating a new draft (v3), it will be based on v1 instead of v2.

## To Reproduce

1. Create a new work with a version history as above
2. Create a draft
3. The draft will be based on the published version, not the withdrawn version

## Expected behavior
The draft should be based on the most recent version even if it's withdrawn.

awead:
New versions of works should come from the representative version, which could either be withdrawn or published, but will always be the version that the user sees in their dashboard.

Fixes #1054

### DIFF
--- a/app/controllers/dashboard/work_versions_controller.rb
+++ b/app/controllers/dashboard/work_versions_controller.rb
@@ -6,7 +6,7 @@ module Dashboard
       work = Work.find(params[:work_id])
       authorize(work, :create_version?)
 
-      @work_version = BuildNewWorkVersion.call(work.latest_version)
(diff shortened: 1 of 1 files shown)


nicktiberi (commented):
LGTM, but should we write a test to verify this works for the scenario in #1054? Not sure if that is easy or tricky.

awead:
@nicktiberi I thought about that, but figured if nothing exiting broke, it was ok, and the upcoming UI changes will require tests for this same thing.

nicktiberi (approved):

PR #1090, opened 2021-06-16
I#1031 change depositor

### LINKED ISSUE #1031: Ability for admins to change depositor
# Feature Request

The `depositor` field should be editable by admins through the work settings page (in the "Danger Zone").

- A new field in the danger zone of the work settings page (only appears for admins) for changing the depositor username.

awead:
Adds a additional form to the work and collection settings page to switch the depositor to a different actor. Existing actors or current Penn State users are allowed via the BuildNewActor service. If the actor does not exist, and the user's access id is not currently active at Penn State, then the user is not allowed to be the depositor.

Fixes #1031


nicktiberi (approved):
LGTM! I tested it out locally and it works well.

Not for this PR, of course, but I do wonder whether we can share more code between the work and collection edit forms. They are super similar.

PR #1091, opened 2021-06-17
match bundler

whereismyjetpack:


Dockerfile
```diff
@@ -14,7 +14,7 @@ USER app
 
 COPY Gemfile Gemfile.lock /app/
 COPY --chown=app vendor/ vendor/
-RUN gem install bundler:2.1.4
+RUN gem install bundler:2.2.20
```
awead: Could we do something like:

``` Dockerfile
RUN gem install bundler -v "$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tail -n 1)"
```

That way we should always get the version that's specified in the Gemfile.lock.

awead (changes requested):

awead (approved):

PR #1092, opened 2021-06-17
FindByMimeType and UpdateExtractedText

### LINKED ISSUE #1057: Extract text from existing files in Scholarsphere
After the 4.8.0 release, we need to extract text from existing files in Scholarsphere and reindex them.

##  WDLL

* identify file resources will text-base files
* run a sub-set of the files, ~1000 or so, to test
* run the remainder of the files

We should use the same process we did with virus checks to reduce download costs from Amazon.

awead:
Creates two new service classes allowing us to update a file's extracted text based on its mime type. Two different classes can be used, together if needed, or separately for other purposes.

FindByMimeType allows us to do

``` ruby
FileResource.find_by_mime_type('image/png')
```

much like we would any other column-level query in ActiveRecord, only in this case, the service object builds the necessary sql query to perform the search.

UpdateExtractedText will update any FileResource that doesn't have existing extracted text. Forcing it, with `force: true`, will update it even if it already has text. Resulting jobs are always run async.

Fixes #1057

### DIFF
--- a/app/services/find_by_mime_type.rb
+++ b/app/services/find_by_mime_type.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+# @abstract Returns or yields a set of FileResource objects based on their mime types.
+#
+# Examples
+#
+# Return a set of file resources for a given mime type
+#
+#   resources = FindByMimeType.call(mime_types: 'image/png')
+#
+# Return a set of file resources for a set of mime types
+#
+#   resources = FindByMimeType.call(mime_types: ['image/png', 'image/jpeg'])
+#
+# Iterate over all text-based file resources. This is helpful when needing to batch process a large set of files
+#
+#   FindByMimeType.call(mime_types: :text) do |resource|
+#     resource.update(...)
+#   end
+#
+
+class FindByMimeType
+  TEXT = %w(
+    application/msword
+    application/pdf
+    application/postscript
+    application/rtf
+    application/vnd.ms-excel
+    application/vnd.oasis.opendocument.spreadsheet
+    application/vnd.oasis.opendocument.text
+    application/vnd.openxmlformats-officedocument.presentationml.presentation
+    application/vnd.openxmlformats-officedocument.spreadsheetml.sheet
+    application/vnd.openxmlformats-officedocument.wordprocessingml.document
+    application/vnd.wordperfect
+    application/x-mobipocket-ebook
+    html
+    text
+    xml
+  ).freeze
+
+  class << self
+    def call(mime_types: [])
+      types = mime_types == :text ? TEXT : Array.wrap(mime_types)
+      sql = build_statement(types)
+
+      ActiveRecord::Base
+        .connection
+        .execute(sql)
+        .values
+        .flatten
+        .map do |id|
+          FileResource.find(id) do |resource|
+            block_given? ? yield(resource) : resource
+          end
+        end
+    end
+
+    private
+
+      def build_statement(types)
+        inner_sql = types.map do |type|
+          "file_data->'metadata'->>'mime_type' LIKE '%#{type}%'"
+        end.join("\n            OR    ")
+
+        %(
+        SELECT DISTINCT
+          file_version_memberships.file_resource_id
+        FROM
+          file_version_memberships
+        WHERE
+          file_version_memberships.work_version_id IN (
+            SELECT id
+            FROM work_versions
+            WHERE aasm_state = 'published'
+          )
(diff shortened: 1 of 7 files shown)


app/services/find_by_mime_type.rb
```diff
@@ -0,0 +1,85 @@
...
+            FROM work_versions
+            WHERE aasm_state = 'published'
+          )
+        AND
+          file_version_memberships.file_resource_id IN (
+            SELECT id
+            FROM file_resources
+            WHERE #{inner_sql}
+          )
+        )
```
nicktiberi, line 82: Not sure how much of a performance impact this would have, but a join might be faster than a subquery. Might be worth testing it out to see if there is any meaningful difference. Don't let it hold this up, though. Just a suggestion.

Caution: untested! ☠️ ☠️ ☠️ 

```
SELECT DISTINCT
	file_version_memberships.file_resource_id
FROM
	file_version_memberships
JOIN 
	work_versions 
ON 
	file_version_memberships.work_version_id = work_versions.id
JOIN 
	file_resources 
ON 
	file_version_memberships.file_resource_id = file_resources.id
WHERE 
	work_versions.aasm_state = 'published'
AND 
	#{inner_sql}
```

app/services/update_extracted_text.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# @abstract Updates the extracted text for a give FileResource.
+class UpdateExtractedText
+  def self.call(resource:, force: false)
+    return if resource.extracted_text.present? && (force != true)
```
nicktiberi, line 6: I guess we explicitly have to say `force != true` to protect against passing some random non-boolean value? Otherwise we could just use `!force`.

nicktiberi (approved):
Looks good! Just a couple minor comments that should not hold up this PR.

app/services/find_by_mime_type.rb
awead: @nicktiberi Good idea! I tried that just now in the postgres shell, and the join query is ~10 ms _slower_ than my subquery approach. I wasn't expecting that because I would assume joins would be faster. My guess is that the join is operating on all the rows, whereas the subqueries reduce the number of rows on each table a bit. 🤷

nicktiberi: @awead - interesting! Yeah, I would have expected the join to be faster, but your theory makes sense. Subqueries are easier to read, anyway, IMO.

awead: FWIW, I did:

``` sql
SELECT DISTINCT
        file_version_memberships.file_resource_id
FROM
        file_version_memberships
JOIN
        work_versions
ON
        file_version_memberships.work_version_id = work_versions.id
JOIN
        file_resources
ON
        file_version_memberships.file_resource_id = file_resources.id
WHERE
        work_versions.aasm_state = 'published'
AND
        file_data->'metadata'->>'mime_type' LIKE '%pdf%'
OR
        file_data->'metadata'->>'mime_type' LIKE '%text%'
```

And got `Time: 49.560 ms`

Whereas 

``` sql
SELECT DISTINCT
  file_version_memberships.file_resource_id
FROM
  file_version_memberships
WHERE
  file_version_memberships.work_version_id IN (
    SELECT id
    FROM work_versions
    WHERE aasm_state = 'published'
  )
AND
  file_version_memberships.file_resource_id IN (
    SELECT id
    FROM file_resources
    WHERE file_data->'metadata'->>'mime_type' LIKE '%pdf%'
    OR file_data->'metadata'->>'mime_type' LIKE '%text%'
  )
;
```

Comes in at  `Time: 33.869 ms`

nicktiberi: Thanks for benchmarking! ⏱️

app/services/update_extracted_text.rb
awead: I like to be explicit here. There are some cases where `"false"` would evaluate as `true`. Ex:

``` ruby
> force = "false"
=> "false"
> puts "hi" if force
hi
```

But,

``` ruby
> puts "hi" if force == true
=> nil
```

So while `"false"` will _sometimes_ be ok, I err on the side of caution and allow the only possible true case, where you definitely use a TrueClass. It doesn't read very well, but the user will _have_ to say `force: true`.

app/services/find_by_mime_type.rb
awead: Numbers don't lie... unless your math is wrong.

PR #1093, opened 2021-06-17
Move version navigation to right column

### LINKED ISSUE #1080: Move version list to the right-hand page
List versions after metadata and remove the dropdown from the blue bar. This is borrowed from Zenodo.

## Examples

Screen Shot 2021-06-14 at 9 43 06 AM
Screen Shot 2021-06-14 at 9 42 32 AM

## WDLL

* update the component that renders the drop down
* create a new section between metadata and analytics
* create a fixed-height or scrollable section

nicktiberi:
Fixes #1080.

Removes the version dropdown from the blue bar and places it in the righthand sidebar. The version selector is scrollable, and will automatically scroll to put the active version in view on initial page load.

This functionality can be previewed at https://scholarsphere-version-navigation.dsrd.libraries.psu.edu/

Screen Shot 2021-06-16 at 3 05 55 PM


app/components/work_versions/version_navigation_component.rb
```diff
@@ -37,7 +38,13 @@ def path_for(version)
       end
     end
 
-    def classes_for(version)
+    def list_classes_for(version)
+      classes = []
+      classes << 'active' if current?(version)
+      classes.join(' ')
```
awead: I'm not seeing any other classes, so the return for this function would either be `nil` or `'active'`. Is that right? If so, you can just rewrite it as:

``` ruby
def list_classes_for(version)
  return unless current(version)

  'active'
end

File: app/components/work_versions/version_navigation_component.rb [hunk 2]
```diff
@@ -37,7 +38,13 @@ def path_for(version)
...
+    end
+
+    def link_classes_for(version)
```
awead, line 45: You can make the same change here, too.

app/javascript/controllers/version_navigation_controller.js
```diff
@@ -0,0 +1,11 @@
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+  connect () {
+    // auto-scroll to the active version so it's visible on initial page load
+    const versionNavigation = this.element
+    const activeVersion = versionNavigation.querySelector('.active')
```
awead, line 7: It's worth mentioning, but probably _not_ worth changing, that we'll often pass in class names where the controller is initialized. Ex:

``` html
<div data-controller="version-navigation" data-version-naviation-active-value="active">
```

And then reference the value in the controller:
``` js
static values = { active: String }

connect () {
  const versionNavigation = this.element
  const activeVersion = versionNavigation.querySelector(this.activeValue)
}
```

It also looks like this has changed slightly in Stimulus 2.0. I think you're contribution here is the first 2.0 code. Everything prior to this was 1.x. But, this is all just FYI. I think having `.active` in there directly is fine.

awead (changes requested):
Looks good! There's only one thing about the classes method that might be refactored. I'm wondering if there were additional classes in previous iterations that are now no longer used.

app/components/work_versions/version_navigation_component.rb
nicktiberi: @awead - Good call! The original version of this did the same, but also had one class. I thought that was kinda strange but I guess in the moment "preserve original style" won out over "refactor this" haha. 🙂

awead: No, I think "preserve original style" is the best first step. I often forget that and jump straight to refactoring, when really it's better to have the refactoring emerge from the keeping-things-looking-the-same.

awead (dismissed):
I think this is ready to go. Do we want another round of 👓 from @srerickson and others?

nicktiberi:
> I think this is ready to go. Do we want another round of 👓 from @srerickson and others?

Yeah, I'd just like to get the seal of approval from @srerickson before we merge it up.

awead (approved):

PR #1095, opened 2021-06-18
Combine Edit and Create Version buttons into an Update Work button

### LINKED ISSUE #1078: Combine edit and create buttons in the resource page
Depending on the state of the resource, you can either edit or create. We don't need separate buttons for this.

## WDLL

* a single button that states either "Edit" or "Create" depending on what's available to the user
* admins will have a separate button to edit versions that have already been published

### LINKED ISSUE #1082: Make collection edit button consistent with work edit button
Corresponds to changes in #1078. Ensure that buttons for collection edit look the same as work edit.

## WDLL

* move collection edit button to the left of collection settings
* use the same view component ?

nicktiberi:
Fixes #1078.
Fixes #1082.

The functionality of the edit and create new version butons has been combined into a single button whose action changes depending on whether a draft version exists.

If a draft version exists, the update button acts as "edit draft," and if no draft version exists, the update button acts as "create new draft."

The button to edit a specific version (draft or otherwise) is now an admin-only feature.

This is available for preview at https://scholarsphere-update-work-button.dsrd.libraries.psu.edu

Screen Shot 2021-06-24 at 10 45 08 AM

### DIFF
--- a/app/components/edit_resource_button.html.erb
+++ b/app/components/edit_resource_button.html.erb
@@ -1,0 +1,8 @@
+<%= link_to path,
+            method: method,
+            class: 'btn btn-outline-light btn--squish btn--edit mr-lg-2',
+            title: t('resources.edit_button.tooltip'),
+            data: { toggle: 'tooltip', placement: 'bottom' } do %>
+            <%= label %>
+            <i class="material-icons" aria-hidden="true">edit</i>
+<% end %>
--- a/app/components/edit_resource_button.rb
+++ b/app/components/edit_resource_button.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class EditResourceButton < ApplicationComponent
+  attr_reader :resource, :policy
+
+  def initialize(resource:, policy:)
+    @resource = resource
+    @policy = policy
+  end
+
+  def path
+    if collection?
+      dashboard_form_collection_details_path(resource.id)
+    elsif has_draft?
+      # update the existing draft version
+      dashboard_form_work_version_details_path(resource.work.draft_version.id)
+    elsif policy.new?
+      # create a new draft work version
+      dashboard_work_work_versions_path(resource.work)
+    end
+  end
+
+  def label
+    I18n.t('resources.edit_button.text', type: type)
+  end
+
+  def method
+    return if collection? || has_draft?
+
+    'post'
+  end
+
+  private
+
+    def has_draft?
+      return false if collection?
+
+      resource.work.draft_version.present?
+    end
+
+    def collection?
+      resource.is_a?(CollectionDecorator)
+    end
+
+    def type
+      collection? ? 'Collection' : 'Work'
+    end
+end
--- a/app/views/resources/_collection.html.erb
+++ b/app/views/resources/_collection.html.erb
@@ -8,18 +8,18 @@
 
 <%= content_for :navbar_items do %>
   <% if policy(collection).edit? %>
-    <li class="nav-item">
-      <%= link_to(edit_dashboard_collection_path(collection),
-                  class: 'btn btn-outline-light btn--squish mr-lg-2 btn--settings') do %>
-          <%= t('resources.settings_button.text', type: 'Collection') %>
-          <i class="material-icons">settings</i>
-      <% end %>
+    <li class="nav-item py-1">
+      <%= render EditResourceButton.new(resource: collection, policy: policy(collection)) %>
     </li>
 
     <li class="nav-item py-1">
-      <%= link_to t('resources.collection.edit_button'),
-                  dashboard_form_collection_details_path(collection.id),
-                  class: 'btn btn-outline-light btn--squish mr-lg-2' %>
+      <%= link_to edit_dashboard_collection_path(collection),
+                  class: 'btn btn-outline-light btn--squish mr-lg-2 btn--settings',
+                  title: t('resources.settings_button.tooltip'),
+                  data: { toggle: 'tooltip', placement: 'bottom' } do %>
+          <%= t('resources.settings_button.text', type: 'Collection') %>
+          <i class="material-icons" aria-hidden="true">settings</i>
+      <% end %>
     </li>
   <% end %>
 <% end %>
--- a/app/views/resources/_work_version.html.erb
+++ b/app/views/resources/_work_version.html.erb
@@ -7,33 +7,27 @@
 <% end %>
 
 <%= content_for :navbar_items do %>
+  <% if current_user.admin? %>
+    <li class="nav-item py-1">
+      <a class="btn btn-outline-light btn--squish mr-lg-2 qa-edit-version"
+          href="<%= dashboard_form_work_version_details_path(work_version.id) %>">
+        <%= I18n.t('resources.work_version.admin_edit_button', version: work_version.display_version_short) %>
+      </a>
+    </li>
+  <% end %>
+
   <% if policy(work_version.work).edit? %>
     <li class="nav-item py-1">
-      <%= render LinkDisabledByTooltipComponent.new(
-            enabled: policy(work_version).edit?,
-            text: I18n.t('resources.work_version.edit_button.text', version: work_version.display_version_short),
-            path: dashboard_form_work_version_details_path(work_version.id),
-            tooltip: I18n.t('resources.work_version.edit_button.tooltip'),
-            class_list: 'btn btn-outline-light btn--squish mr-lg-2 qa-edit-version'
-          ) %>
+      <%= render EditResourceButton.new(resource: work_version, policy: policy(work_version.work.representative_version)) %>
     </li>
 
     <li class="nav-item py-1">
-      <%= render LinkDisabledByTooltipComponent.new(
-            enabled: policy(work_version).new?,
-            text: I18n.t('resources.work_version.create_button.text', version: work_version.display_version_short),
-            path: dashboard_work_work_versions_path(work_version.work),
-            method: :post,
-            tooltip: I18n.t('resources.work_version.create_button.tooltip'),
-            class_list: 'btn btn-outline-light btn--squish mr-lg-4 qa-create-draft'
-          ) %>
-    </li>
-
-    <li class="nav-item py-1">
-      <%= link_to(edit_dashboard_work_path(work_version.work),
-                  class: 'btn btn-outline-light btn--squish mr-lg-2 btn--settings') do %>
+      <%= link_to edit_dashboard_work_path(work_version.work),
+                  class: 'btn btn-outline-light btn--squish mr-lg-2 btn--settings',
+                  title: t('resources.settings_button.tooltip'),
+                  data: { toggle: 'tooltip', placement: 'bottom' } do %>
           <%= t('resources.settings_button.text', type: 'Work') %>
-          <i class="material-icons">settings</i>
+          <i class="material-icons" aria-hidden="true">settings</i>
       <% end %>
     </li>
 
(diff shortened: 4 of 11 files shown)


nicktiberi:
@awead - the following test is failing, and I am wondering if we need to consider how making "Edit Work" an admin-only feature affects the dashboard.

`rspec ./spec/features/dashboard/catalog_spec.rb:94 # Dashboard catalog page when the user has only editable works and collections, none of which they created displays the user's editable works`

Is this test still relevant? If so, what should we be testing in this scenario? If a user is not an admin, the "Edit Work" button is no longer visible.

awead (changes requested):
Overall, this is good. I'm having two thoughts:

1) Rename it to UpdateResource or EditResourceButton, enable it with collections, you've just fixed #1082 as well.

2) Combine the admin button feature into the same component, and the button becomes a drop-down with two options: a) edit existing version; or, b) create new version. This could be pushing this out of scope, so I'm not super pushy on it.

nicktiberi:
@awead - so #1082 is basically just to move the Edit Collection button to the left of the collection settings. Maybe we can tackle that + combining the Edit Work and Edit Collection buttons into a single EditResourceButton component in a separate PR, once this one gets merged up? How do you feel about that?

Re: the dropdown, I think that could be cool, but we are already making a bunch of UI changes so maybe we should run that by Seth before we go that route.

awead:
@nicktiberi You can do them separately. I would make sure to use the class name EditResourceButton, otherwise the file will change names from one PR to the other and create some unneeded churn.

nicktiberi:
@awead - I guess some clarification is in order here. The terminology is starting to get a little murky. There are two action buttons on the work version page:
- Update Work (new to this PR)
- Edit Work Version (changed to admin-only in this PR)

My thought was that Update Work button (`UpdateWorkComponent`) would stay as-is, and then in the separate PR, I would add `EditResourceComponent`, which would combine the Edit Work and Edit Collection buttons into a single component.

Does that sound OK to you? The naming is a bit confusing, so I'm happy to re-evaluate how we name these things if you think that's helpful.

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def method
+      return if collection? || has_draft?
+
+      'post'
+    end
+
+    def has_draft?
+      resource.work.draft_version
```
awead: We can simplify this to `resource.draft?`, but resource might be a collection. There are a couple of ways to deal with that:

Use `try`:
```
resource.try(:draft?)
```

This would return true/false if it's a work version, or nil if it's a collection. Technically, that violates the principle that "?" methods always return true/false, but since nil evaluates to false, it's not too bad. Alternatively, you can just use a guard clause:

``` ruby
return false if collection?

resource.draft?
```

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def type
+      resource.class.name
+    end
+
+    def collection?
+      type == 'CollectionDecorator'
```
awead: Use `resource.is_a?(CollectionDecorator)` and remove the type method.

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def work_version?
+      type == 'WorkVersionDecorator'
```
awead: same here.

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,54 @@
...
+        # update the existing draft version
+        dashboard_form_work_version_details_path(resource.work.draft_version.id)
+      elsif policy.new?
+        # create a new draft work version
+        dashboard_work_work_versions_path(resource.work)
+      end
+    end
+
+    def label
+      collection? ? I18n.t('resources.collection.edit_button') : I18n.t('resources.work_version.edit_button')
```
awead: There's a feature in ActiveRecord models where you can get the i18n key. It's also how it builds all the `edt_*_path` methods and other things. See https://api.rubyonrails.org/classes/ActiveModel/Name.html

Try `I18n.t("resources.#{resource.model_name.i18n_key}.edit_button")`

Even though the resource is a decorated version of the active record object, any methods not explicitly defined on the decorated model, just get pushed up to the model that it's decorating. So we have access to all the same methods we do if resource was just a Collection or WorkVersion.

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,54 @@
...
+
+class EditResourceButton < ApplicationComponent
+  attr_reader :resource, :policy
+
+  def initialize(resource:, policy:)
+    @resource = resource
+    @policy = policy
+  end
+
+  private
```
awead: I'm surprised you can put all the methods as private. I would have thought that calling path, method, and label in the component template would raise "can't call private method" errors. 🤷 Anyway, I was going to suggest that those three methods be public since you're invoking them from outside the class, i.e. in the template.

app/views/resources/_collection.html.erb
```diff
@@ -8,19 +8,17 @@
 
 <%= content_for :navbar_items do %>
   <% if policy(collection).edit? %>
```
awead, line 10: We could get rid of this `if` block and add:

``` ruby
def render?
  policy.edit?
end
```

to our component. This is a nice way of controlling when to display your components without having to put any logic in your views.

app/views/resources/_work_version.html.erb
```diff
@@ -13,25 +13,17 @@
       ) %>
 
   <% if policy(work_version.work).edit? %>
```
awead, line 19: Same thing here. The `render?` method would control whether or not to show the button.

app/views/resources/_work_version.html.erb
```diff
@@ -13,25 +13,17 @@
...
-    <li class="nav-item">
-      <%= render LinkDisabledByTooltipComponent.new(
-            enabled: policy(work_version).edit?,
-            text: I18n.t('resources.work_version.edit_button.text', version: work_version.display_version_short),
-            path: dashboard_form_work_version_details_path(work_version.id),
-            tooltip: I18n.t('resources.work_version.edit_button.tooltip'),
-            class_list: 'btn btn-outline-light btn--squish mr-lg-2 qa-edit-version'
-          ) %>
-    </li>
+    <% if current_user.admin? %>
```
awead: Note that this `if` block doesn't need to be inside the policy block, since an admin can always edit resource.

app/views/resources/_work_version.html.erb
```diff
@@ -13,25 +13,17 @@
...
+      <li class="nav-item">
+        <a class="btn btn-outline-light btn--squish mr-lg-2 qa-edit-version"
+           href="<%= dashboard_form_work_version_details_path(work_version.id) %>">
+          <%= I18n.t('resources.work_version.edit_version_button', version: work_version.display_version_short) %>
```
awead: `edit_version` threw me for a little bit. Would `admin_edit_button` make more sense?

awead (changes requested):
This is great! I've left a bunch of comments that are refinements and about using some Rails features that can tighten this up a bit, but overall this is exactly what it needs to be.

app/components/edit_resource_button.rb
nicktiberi: 👍 - FWIW, this code was copied from `DeleteResourceButtonComponent`. I agree that this approach is cleaner and simpler.

app/components/edit_resource_button.rb
nicktiberi: TIL! I was wondering where those paths magically came from. 😁

app/components/edit_resource_button.rb
nicktiberi: I'm guessing `private` in this case means private to the class and any templates using the class. So if you were to subclass `EditResourceButton` you would not get access to `path`, `method`, and `label`. Just a guess. But yeah, I'll move those 3 methods out of `private`.

app/components/edit_resource_button.rb
awead: Just goes to show, you can't trust @rschenk. (joke) Actually, he used `type` elsewhere in the class, so that makes sense. I don't think we're doing that here? So we can take a short-cut.

app/components/edit_resource_button.rb
awead: I'm intrigued by this now and wonder if this is a bug or if it's by design....

app/views/resources/_collection.html.erb
nicktiberi: The only issue with this is that then we would have an empty `<li>` hanging out in our HTML since the `<li>` would still render but the button would not.

We could include the `<li>` in the component, but I don't know if I like that, because it arbitrarily assumes that this button will only ever be in a list.

The other thing is that the `if` block also controls the rendering of the settings below, so we would still need it anyway unless we refactored things further.

app/components/edit_resource_button.rb
nicktiberi: Actually `resource.draft?` doesn't work, because that returns if the current work version is a draft. What we are trying to determine is if the parent work has any draft versions.

So how about:

```
return false if collection?
      
resource.work.draft_version != nil
```

app/views/resources/_collection.html.erb
awead: I think solve the `<li>` issue by using slots (https://viewcomponent.org/guide/slots.html) which would enable us to put the button in different html surroundings. Also, I think the settings button should be refactored into its own component as well, and then could be treated the same as the edit resource. However, this is getting further afield from the ticket, so let's leave it as is for now.

nicktiberi:
@awead - this should be ready to go.

awead (approved):

app/components/edit_resource_button.rb
```diff
@@ -0,0 +1,48 @@
...
+      return false if collection?
+
+      resource.work.draft_version.present?
+    end
+
+    def collection?
+      resource.is_a?(CollectionDecorator)
+    end
+
+    def work_version?
```
awead: @nicktiberi Sorry, I just noticed this method doesn't get called anywhere. Am I missing something?

awead (changes requested):

app/components/edit_resource_button.rb
nicktiberi: No, you're not missing something. Good catch! Another copy-paste thing. Just pushed an update that removes this method.

awead (dismissed):

app/components/edit_resource_button.rb
rschenk: hacking in progress

app/components/edit_resource_button.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= link_to path,
+            method: method,
+            class: 'btn btn-outline-light btn--squish btn--edit mr-lg-2',
+            title: t('resources.edit_button.tooltip'),
+            data: { toggle: 'tooltip', placement: 'bottom' } do %>
+            <%= label %>
+            <i class="material-icons">edit</i>
```
awead: We need to add `aria-hidden="true"` here otherwise the screen reader with pronounce "edit".

awead (changes requested):
One last tiny addition...

app/components/edit_resource_button.html.erb
nicktiberi: Added `aria-hidden="true"` to the edit component as well as the settings buttons. Making all of these changes in 2 places really underscores the need to componentize the settings button as well...

awead: Ha. That's funny, @nicktiberi  because I added a comment to the settings button too, but changed my mind and removed it because I thought it was out of scope. But, yes, absolutely. If we had a central "button component" that did all of that for us....

awead (approved):

PR #1099, opened 2021-06-24
Add vim to Dockerfile

awead:
I've found that I needed to edit some files in the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,10 @@ ARG UID=2000
 COPY bin/vaultshell /usr/local/bin/
 USER root
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


whereismyjetpack:
This is kind of an anti pattern, any file that is edited in the container will be reverted on the next container restart, leaving the application in an undesirable state. If I understood the usecase a bit I might be able to come up with a more stable solution

awead:
I'm going off of what I saw in the catalog Dockefile. I've needed to use a editor in the deployed container before but there hasn't been one.

awead:
It's been for debugging or doing some one-off work, so I wouldn't expect the changes to persist between deployments.

whereismyjetpack:
Oof that was a mistake.

PR #1106, opened 2021-06-28
Create reusable ResourceButton component

### LINKED ISSUE #1105: Make Resource Settings Button a Component
# Feature Request
We've already moved the Edit Resource button into a component, so let's do the same for Resource Settings.

The Resource Settings button is moved into a reusable component, and that component is used on the Work Version and Collection details pages.

nicktiberi:
Fixes #1105.

We now have an "abstract" component called `ResourceButton`, with two components extending its functionality:

- `EditResourceButton`
- `ResourceSettingsButton`

These two components are currently used in the Work Version and Collection details views.

### DIFF
--- a/app/components/resource_button.rb
+++ b/app/components/resource_button.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class ResourceButton < ApplicationComponent
+  attr_reader :resource, :policy
+
+  def initialize(resource:, policy: nil)
+    @resource = resource
+    @policy = policy
+  end
+
+  def method; end
+
+  private
+
+    def has_draft?
+      return false if collection?
+
+      resource.work.draft_version.present?
+    end
+
+    def collection?
+      resource.is_a?(CollectionDecorator)
+    end
+
+    def type
+      collection? ? 'Collection' : 'Work'
+    end
+end
--- a/app/components/resource_edit_button.rb
+++ b/app/components/resource_edit_button.rb
@@ -1,13 +1,6 @@
 # frozen_string_literal: true
 
-class EditResourceButton < ApplicationComponent
-  attr_reader :resource, :policy
-
-  def initialize(resource:, policy:)
-    @resource = resource
-    @policy = policy
-  end
-
+class ResourceEditButton < ResourceButton
   def path
     if collection?
       dashboard_form_collection_details_path(resource.id)
@@ -24,25 +17,17 @@ class EditResourceButton < ApplicationComponent
     I18n.t('resources.edit_button.text', type: type)
   end
 
+  def tooltip
+    I18n.t('resources.edit_button.tooltip')
+  end
+
+  def icon
+    'edit'
+  end
+
   def method
     return if collection? || has_draft?
 
     'post'
   end
-
-  private
-
-    def has_draft?
-      return false if collection?
-
-      resource.work.draft_version.present?
-    end
-
-    def collection?
-      resource.is_a?(CollectionDecorator)
-    end
-
-    def type
-      collection? ? 'Collection' : 'Work'
-    end
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -103,8 +103,8 @@ RSpec/MultipleSubjects:
 
 RSpec/VerifiedDoubles:
   Exclude:
-    - 'spec/components/edit_resource_button_spec.rb'
     - 'spec/components/mintable_doi_component_spec.rb'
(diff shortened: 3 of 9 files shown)


nicktiberi:
@awead - I am wondering if we should rename `EditResourceButton` -> `ResourceEditButton` for consistency. Then we would have:

- `ResourceButton`
- `ResourceEditButton`
- `ResourceSettingsButton`
- `Resource(some other action)Button`

app/components/resource_button.rb
```diff
@@ -0,0 +1,30 @@
...
+class ResourceButton < ApplicationComponent
+  attr_reader :resource, :policy
+
+  def initialize(resource:, policy: nil)
+    @resource = resource
+    @policy = policy
+  end
+
+  def method
+    nil
```
awead: use `def method; end`

`nil` is implied. This is typically the _only_ place we're you'll find one-line method invocations.

awead (changes requested):
We're inching towards something like

``` ruby
render ResourceSettingsButton.new(
  resource: work_version.work,
  icon: 'edit',
  tooltip: 'resources.edit_button.tooltip'
)
```
But I don't want to get too far ahead of myself.

nicktiberi:
> We're inching towards something like
> 
> ```ruby
> render ResourceSettingsButton.new(
>   resource: work_version.work,
>   icon: 'edit',
>   tooltip: 'resources.edit_button.tooltip'
> )
> ```
> 
> But I don't want to get too far ahead of myself.

If we make `icon` and `tooltip` params, that implies that they could vary for different resource settings buttons, and we will have to pass those params in every time we create a new `ResourceSettingsButton`. IMO it is cleaner to leave these properties internal to the component and only make them params if/when we need to customize them from outside the component. I feel like adding duplicate required params starts to defeat the purpose of the component a little bit.

I'll defer to you, but those are my thoughts, FWIW.

awead:
@nicktiberi Having default with parameter overrides would work, but at this point we only have 2 use cases. If we get to more than three, we can refactor it then.

app/components/resource_button.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class ResourceButton < ApplicationComponent
+  attr_reader :resource, :policy
+
+  def initialize(resource:, policy: nil)
+    @resource = resource
+    @policy = policy
```
awead, line 8: @nicktiberi we don't appear to be using policy anywhere. I recall that we were at one point. Do we need to?

awead (changes requested):

app/components/resource_edit_button.rb
```diff
@@ -1,13 +1,6 @@
...
-class EditResourceButton < ApplicationComponent
-  attr_reader :resource, :policy
-
-  def initialize(resource:, policy:)
-    @resource = resource
-    @policy = policy
-  end
-
+class ResourceEditButton < ResourceButton
   def path
```
nicktiberi, line 4: @awead - we are using `policy` in the `path` method of `ResourceEditButton`:
```
elsif policy.new?
  # create a new draft work version
  dashboard_work_work_versions_path(resource.work)
end
```
Though I wonder if we can either derive it from within the component instead of passing it in, or if we even need it at all, since the app would probably 403 or something if the user tried to create a new draft version and they somehow did not have the permissions to do so. As it is now, if they somehow didn't have permissions to create a new draft, `path` would return `nil`, so would kind of be a different bug.

Maybe we can just change the code to an `else` only, and stop passing in `policy`?

```
else
  # create a new draft work version
  dashboard_work_work_versions_path(resource.work)
end
```

Let me know what you think.

awead: @nicktiberi No, I like being explicit with `.new?`. I was just missing it.

app/components/resource_button.rb
awead: Nm... I just wasn't seeing it.

awead:
@nicktiberi Sorry, I missed your earlier comment about renaming. Yes, I like the Resource(thing)Button pattern throughout. Feel free to change it! But I won't get in the way if you're tired of this. :)

awead (approved):

nicktiberi:
> @nicktiberi Sorry, I missed your earlier comment about renaming. Yes, I like the Resource(thing)Button pattern throughout. Feel free to change it! But I won't get in the way if you're tired of this. :)

Renaming is already done! So I think this should be good to go. 🙂

PR #1161, opened 2022-01-10
updates solr in docker to match what's in qa

whereismyjetpack:
We used to maintain our own solr, it was just upstream solr, with the start-solr script in it. so we can have one less thing to manage, we mount in the script and pull from upstream


nicktiberi:
Hmm... looks like Solr becomes unavailable in CI with this change:

```
Waiting for Solr to become available
Solr never became ready. Exiting
ERROR: 1

Exited with code exit status 1

CircleCI received exit code 1
```

https://app.circleci.com/pipelines/github/psu-libraries/scholarsphere/830/workflows/8c1ec756-43f4-4bd7-855b-04b45e566819/jobs/1229

whereismyjetpack:
I forgot to come back and check CI on this one. that most certainly should not have happened. I'll look into it. One thing I have to do this year is refactor all of CI -- It's slow, and error prone :(

whereismyjetpack:
This won't work in CI. i'm trying to find another way to do this. I might just have to inline the script. Sorry for the mess.

nicktiberi:
> This won't work in CI. i'm trying to find another way to do this. I might just have to inline the script. Sorry for the mess.

No worries! Thank you for the valiant attempt! 🙂

whereismyjetpack:
We're getting closer, now we're running into an issue where `solr:init` is being run before the authentication container has completed. solr will happily accecpt this intialization but then mark it unsafe because it was uploaded without auth (cool, dude) 

I'm going to close this and refactor CI so that it runs all the steps in series to avoid this chicken and egg problem.

PR #1165, opened 2022-01-12
update solr

whereismyjetpack:
- updates solr in docker to match what's in qa

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,7 @@ x-web_env: &web_env
       REDIS_HOST: redis
       SOLR_USERNAME: scholarsphere
       SOLR_PASSWORD: scholarsphere
+      BUNDLE_DEFAULT_INSTALL_USES_PATH: vendor/bundle
       REDIS_PASSWORD: redispassword
       OAUTH_APP_ID: ${OAUTH_APP_ID}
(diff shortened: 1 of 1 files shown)


whereismyjetpack:
@nicktiberi I finally found a way to get this to work

docker-compose.yml
```diff
@@ -81,18 +82,20 @@ services:
     - -c 
     - mkdir -p /data/scholarsphere; minio --compat server --address ':9000' /data
   solr:
-    image: psul/solr:8.5.2-slim
+    image: solr:8.11.1-slim
     restart: always
     environment:
-      SOLR_STOP_WAIT: 5
+      SOLR_STOP_WAIT: 10
```
nicktiberi, line 88: Just for my own info, what is the significance of this environment variable and why did it change?

whereismyjetpack: SOLR_STOP_WAIT is the time solr will wait before forcefully killing it. it defaults to something crazy like 180. as it turns out that variable is also used for the start_wait time, and i had to increase it.

nicktiberi (approved):
LGTM! Something seems to be weird with Codeclimate, but this change wouldn't affect any of those results.

whereismyjetpack:
I'll put in a ticket to look into code climate

PR #1169, opened 2022-01-14
removes unused docker-compose services. refactors startup for performance

whereismyjetpack:
I'm going through the docker-compose setup for Monday, and came across some items that needed fixing. sidekiq doesn't need to be running all the time, so i took that service out. if we need to run sidekiq for testing we can do that as a one off (i'll put that in the docs) I have also updated the docs to go along with this PR (https://github.com/psu-libraries/scholarsphere/wiki/Docker)

I moved the startup items around, to only run in development mode, in kubernetes we are going to handle this with a cron job. I removed the db:seed task, because it ran at every startup, is slow, and will fill up the database over time. In the docs there's a seperate command to run to seed the database if you want

### DIFF
--- a/.envrc.sample
+++ b/.envrc.sample
@@ -1,0 +1,7 @@
+# Get values from vault
+# vault kv get secret/devteam/scholarsphere
+export OAUTH_APP_ID=
+export OAUTH_APP_SECRET=
+export OAUTH_APP_URL=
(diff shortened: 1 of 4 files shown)


nicktiberi (dismissed):
Lean and mean! Thank you! 🙂

nicktiberi (approved):

PR #1172, opened 2022-01-18
Markdown support for field hints' translations

### LINKED ISSUE #1156: "read more" links for field help text
# Feature Request

The "hint" text that appears under some field elements (as configured in [`en.yml`](https://github.com/psu-libraries/scholarsphere/blob/42298b892e98edd83eb60db38cd04336edd4c9d8/config/locales/en.yml#L196)) doesn't support arbitrary html or links. This is a problem in cases where we'd like to refer users to additional documentation relating to the field. We should fine a way to allow "read more" links in the hint text. 

*This is blocking #793*

I can think of two approaches:

- interpret the `helpers.hint` values in `en.yml` as markdown or html and add links when necessary.
- Add new helper for setting a `readmore_url` for a field. If the url is set then a "read more here" links is added to the end of the hint text.

I think I prefer the first option but I could be convinced otherwise.

rschenk:
For field hints (and only field hints) you can use either Markdown in the en.yml file and it will be parsed and rendered. You can also use raw HTML if you like, but the MD syntax is cleaner so I suggest you use that.

I did this by refactoring all of our input field partials to use a common View Component for rendering the hint. This view component has the smarts to parse the markdown that may be present in the en.yml file. Note that I did not add MD support to everything in the en.yml file, only this one component in our application logic. 

This had the side benefit of really cleaning up our form field partials too.
Screen Shot 2022-01-18 at 2 00 31 PM

Closes #1156

### DIFF
--- a/app/components/field_hint_component.rb
+++ b/app/components/field_hint_component.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+class FieldHintComponent < ApplicationComponent
+  attr_reader :form,
+              :attribute
+
+  class << self
+    def markdown
+      @markdown ||= Redcarpet::Markdown.new(
+        Redcarpet::Render::HTML.new(filter_html: false),
+        {}
+      )
+    end
+  end
+
+  def initialize(form:, attribute:)
+    @form = form
+    @attribute = attribute
+  end
+
+  def render?
+    I18n.exists?(i18n_key)
+  end
+
+  def dom_id
+    return unless render?
+
+    "#{form_field_id}-hint"
+  end
+
+  def hint_text
+    render_markdown I18n.t(i18n_key)
+  end
+
+  def i18n_key
+    @i18n_key ||= ['helpers', 'hint', form.object.class.model_name.i18n_key, attribute].join('.')
+  end
+
+  private
+
+    def render_markdown(text)
+      rendered = self.class.markdown.render(text)
+
+      # remove the surrounding <p> tag that redcarpet insists on inserting
+      rendered
+        .strip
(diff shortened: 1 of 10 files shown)


srerickson (approved):

app/components/field_hint_component.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+class FieldHintComponent < ApplicationComponent
+  attr_reader :form,
+              :attribute
+
+  class << self
```
nicktiberi, line 7: Just for my own knowledge, what is the purpose of putting the `markdown` method inside `class << self`?

nicktiberi (approved):
This looks awesome — great cleanup! 🥇

app/components/field_hint_component.rb
rschenk: Good question. The markdown library that we're using, redcarpet, suggested doing such a thing in their [readme](https://github.com/vmg/redcarpet#and-its-like-really-simple-to-use): 

> The `Redcarpet::Markdown` object is encouraged to be instantiated once with the required settings, and reused between parses.

I interpreted that as meaning that we should instantiate it once for the entire `FieldHintComponent` class, rather than at the instance level every time another FieldHintComponent is rendered. 

I agree that it's a bit nonstandard, but that's why I did it.

Edit: btw the `class << self` syntax is a Ruby way of making methods that exist once at the class level rather than at the instance level. Ruby tends to call these "class methods" but other languages call these static methods (I know Java, ES6 javascript, and Python all use the "static" lingo).

PR #1173, opened 2022-01-18
Update ci

whereismyjetpack:
- Runs CI outside of docker, to allow for parallel builds / better caching
- uses circleci headless browser for js tests to reduce complexity

Code climate works again, and while tests are still slow, they are ~44% faster with this method.


.circleci/config.yml
```diff
@@ -22,144 +21,164 @@ jobs:
...
+        command: |
+          gem install bundler
+          bundle check --path vendor/bundle || bundle install --deployment
+    - run:
+        name: "Yarn"
+        command: |
+          yarn install --cache-folder ~/.cache/yarn --frozen-lockfile
+    # - run:
+    #     name: "Wait for Redis"
+    #     command: dockerize -wait tcp://localhost:6379 -timeout 1m
```
nicktiberi: Should this just be removed since it's commented out?

bin/ci-rspec
```diff
@@ -50,6 +50,10 @@ if [ ! -f ${REPORTER_BIN} ]; then
   chmod 755 ${REPORTER_BIN}
 fi
 
+# update chromedriver
+bundle exec rails webdrivers:chromedriver:update
```
nicktiberi, line 54: Is this something that might prevent tests from sometimes failing due to chromedriver needing an update?

nicktiberi (commented):
Thanks so much for this refactor, @whereismyjetpack! Just one question and a minor comment. ~44% speedup in tests is nothing to 🤧 at.

bin/ci-rspec
whereismyjetpack: webdrivers will do this for free, but VCR freaks out because it's not allowed to connect to the host, so I download them here first

.circleci/config.yml
whereismyjetpack: good catch. i was testing if we needed  redis or not for the tests to run (i wasn't sure if they were mocked out in RAILS_ENV=test, i forgot to readd this.

nicktiberi (approved):
🚀

PR #1178, opened 2022-01-19
Work factory keeps the same list of actors for all versions

### LINKED ISSUE #827: Factories are not duplicating creators correctly
# Bug Report

## Description

When we use FactoryBot to create a works with multiple versions, the creators the later versions have different actor ids. While this is ok from a testing standpoint, this doesn't accurately reflect what typically happens in the application. Usually, subsequent versions will have an updated creator display name, or no change at all, but they should be referencing the same actor id.

The current behavior reflects a "delete and replace" type of scenario where we are removing the original creator and adding a new one in a second version. While this is still ok, we need to have the option of building a work that represents the more typical scenario of a second version with the same original creator.

rschenk:
In our data model, a Work requires at least one associated WorkVersion. So, our factory creates one or more versions when creating a work. You can tell the work factory to create multiple versions if you wish. Previously, each of these versions was created in isolation and therefore was created with its own unique list of authors. This is not a violation of our data model, but for some reason @awead didn't like that and wrote #827 

This PR does a little legwork to maintain the same list of authors within each version created by the work factory. It's a little less complicated than I thought it might be, and only broke one spec in the process.

Closes #827

### DIFF
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -17,6 +17,8 @@ FactoryBot.define do
       end
 
       after(:build, :stub) do |work_version, evaluator|
+        next if work_version.creators.any? # Skip if there are already creators specified
+
         work_version.creators = build_list(:authorship, evaluator.creator_count) do |creator, index|
           creator.resource = work_version
           creator.position = index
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
This looks good to me.

PR #1180, opened 2022-01-20
Limiting the dois returned from AllDois for WorkVersions

### LINKED ISSUE #1164: DOI api endoint should only include Work IDs
The API endpoint for looking up resource IDs by their DOI appears to provide IDs of work versions; it should only provide IDs for Works (or Collections). This is a problem for the integration with RMDB since it means the imported links to ScholarSphere will be for versions, not works.

ajkiessl:
After fiddling around with Solr + Blacklight and realizing I needed to reindex Solr, and after chatting with Seth, it looks like all I had to do was remove one line from `WorkVersion`.  By removing `DoiSchema` from `WorkVersion`'s `#document_builder`, `WorkVersion`s will no longer be a part of the query `DoiSearch` uses to return UUIDs matched to DOIs.  Only `Work`s and `Collection`s are included now.

Closes #1164

### DIFF
--- a/app/models/concerns/all_dois.rb
+++ b/app/models/concerns/all_dois.rb
@@ -35,7 +35,7 @@ module AllDois
   # strings in an array.
   def all_dois
     # WorkVersions should only be included here if they are the
-    # latest_published_version? and their associated Work does not have a DOI
+    # latest published version and their associated Work does not have a DOI
     return [] if forbidden_work_version?
 
     fields_with_dois
@@ -53,7 +53,7 @@ module AllDois
     def forbidden_work_version?
       return false unless self.class == WorkVersion
 
-      !self.latest_published_version? || (self.latest_published_version? &&
-              self.work.fields_with_dois.collect { |field| self.work.send(field).present? }.include?(true))
+      !latest_published_version? || (latest_published_version? &&
+              work.fields_with_dois.map { |field| work.send(field).present? }.include?(true))
     end
 end
--- a/app/models/concerns/all_dois.rb
+++ b/app/models/concerns/all_dois.rb
@@ -34,10 +34,6 @@ module AllDois
   # automatically flattened. All valid dois found are returned as formatted
   # strings in an array.
   def all_dois
-    # DOIs for WorkVersions should only be returned if the WorkVersion is the
-    # latest published version and its associated Work does not have a DOI
-    return [] if forbidden_work_version?
-
     fields_with_dois
       .flat_map { |field_name| send(field_name) }
       .lazy
@@ -47,10 +43,4 @@ module AllDois
       .uniq
       .to_a
   end
-
-  private
-
-    def forbidden_work_version?
-      true if self.class == WorkVersion && null_doi?
-    end
 end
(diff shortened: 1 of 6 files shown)


ajkiessl:
After thinking about this a little more, I need to find a way to identify any most recently published WorkVersions that have a DOI but their Work does not.  I'm going to have to make some changes, so this isn't quite ready for review.

ajkiessl:
This is ready to go.

nicktiberi (dismissed):
This looks good to me, @ajkiessl!

@rschenk - would you mind taking a quick look at this as well?

spec/models/concerns/all_dois_spec.rb
```diff
@@ -41,10 +41,50 @@ def other
...
+        expect(instance.all_dois).to match_array(['doi:10.18113/3ln3-2by1',
+                                                  'doi:10.1515/pol-2020-2011',
+                                                  'doi:10.1007/s10570-013-0029-x'])
+      end
+    end
+
+    context 'when resource is a WorkVersion' do
+      subject(:instance) { work.versions.last }
+
+      let!(:work) { create :work, versions_count: 2, has_draft: false }
```
rschenk, line 55: Because we are using FactoryBot here, we need to change the line at the top of the file from `require 'spec_helper'` to `require 'rails_helper'` (I can't make a comment on that line because it didn't change)

app/models/concerns/all_dois.rb
```diff
@@ -43,4 +47,10 @@ def all_dois
       .uniq
       .to_a
   end
+
+  private
+
+    def forbidden_work_version?
+      true if self.class == WorkVersion && null_doi?
+    end
```
rschenk: This method adds some class-specific code and creates a two-directional dependency here, which I am not crazy about. But there is another way: Roll back all changes in `all_dois.rb` and instead override `all_dois` inside the `WorkVersion` class.

app/models/work_version.rb
```diff
@@ -265,6 +265,11 @@ def update_doi?
     super && latest_published_version?
   end
 
+  def null_doi?
+    !latest_published_version? || (latest_published_version? &&
+        work.fields_with_dois.map { |field| work.send(field).present? }.include?(true))
+  end
+
```
rschenk: There are two changes we can do here.

First is an easy one, `work.fields_with_dois.map { |field| work.send(field).present? }.include?(true)` basically just re-implements `#all_dois` so we can change this to `work.all_dois.any?`

The second thing we can do is delete this method, but use the logic that you wrote to override `#all_dois` for `WorkVersion` like so:

```ruby
class WorkVersion < ApplicationRecord
  # ...

  def all_dois
    return [] if !latest_published_version? || 
                   (latest_published_version? && work.all_dois.any?)

    super
  end

 #...
end

```

rschenk (changes requested):
I think I figured out a way to do the implementation that avoids the `WorkVersion` specific code in the concern (and in fact requires no changes at all to `all_dois.rb` file). See what you think...

rschenk (approved):
🥳 🎉

PR #1190, opened 2022-01-25
Preview/1151 dismax q escape

### LINKED ISSUE #1151: Basic search fails often: "Sorry, I don't understand your search."
# Bug Report

## Description
When searching all fields, the response is often "Sorry, I don't understand your search." I'm not sure what's going on but the search interface should be much more tolerant to varied user input. It shouldn't break for basic searches (see example).

## To Reproduce
1. from the homepage search: "Miniature magnetic rod interfacial stress rheometer for general-purpose microscopes"
2. see: "Sorry, I don't understand your search."

## Expected behavior
basic search should almost never "break" this way.

whereismyjetpack:
Uses `CGI.escape` to escape special characters in the user query portion of the solr query. resolves #1151


nicktiberi (approved):
Looks great! 🚀

PR #1194, opened 2022-01-26
Collection Form: Search for Works Asynchronously

### LINKED ISSUE #1150: Admins not able to add any work to a collection
# Bug Report

## Description
With admin privileges, a user should be able to add any work to any collection, but this isn't the case. Specifically, not all works are findable in the "Add works to collection" form. The drop-down list of works is quite long, but it doesn't include everything. I'm not sure how it is scoped. 

## To Reproduce
1. Enable admin privileges
2. Find a collection to update ([for example](https://scholarsphere-qa.dsrd.libraries.psu.edu/resources/e52091de-ccdd-40b8-8a8b-26a8e4c1430a))
3. On the "add works to collection" form, search for a work that exists (for example: "[Building a Community of Curatorial Practice at Penn State: A Case Study](https://scholarsphere-qa.dsrd.libraries.psu.edu/resources/67744923-b5f4-4beb-8c61-588b9c09daa3)")
4. A work called "Preview of Building a Community of Curatorial Practice at Penn State" is found, but that's not the intended work.

## Expected behavior

With admin privileges, a user should be able to add any work to any collection. *More generally, a user can add a work to a collection if they have edit permissions for both the work and the collection.*

nicktiberi:
Closes #1150.

Instead of searching for all works that a user has access to immediately, we now hit a JSON endpoint with an AJAX request that performs the Solr query. This limits the amount of works that are returned to a maximum of 50.

For admins, this allows them to view a subset of the works they have access to, instead of returning a gigantic dropdown list of ~10k works.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -107,6 +107,7 @@ RSpec/VerifiedDoubles:
     - 'spec/components/resource_edit_button_spec.rb'
     - 'spec/components/work_versions/version_navigation_component_spec.rb'
     - 'spec/components/work_histories/work_history_component_spec.rb'
+    - 'spec/controllers/dashboard/work_search_controller_spec.rb'
 
 Style/ClassVars:
   Exclude:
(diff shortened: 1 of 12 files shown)


banukutlu (approved):
lgtm 👍🏻

PR #1197, opened 2022-01-26
update markdown pages to refer to new documentation site

srerickson:
- replace `help.md` with notice about new documentation site (in case someone lands on this page for some reason).
- add link to documentation site in `about.md`


config/locales/en.yml
```diff
@@ -520,7 +520,6 @@ en:
     admin_name: 'Administrator'
     heading:
       about: 'About'
-      help: 'Documentation & Support'
```
nicktiberi, line 523: @srerickson - Removing this translation value causes the app to raise `translation missing: en.navbar.heading.help` when we visit `/help`. 

Is the intent to keep the help page so that any external links to it still work, and to just change the `Help` link in the header to nav to `https://docs.scholarsphere.psu.edu/`? If we want to keep the Help page, we will need to re-add this.

whereismyjetpack: I'd almost suggest, in `routes.rb` just redirecting to docs.scholarsphere.psu.edu on gets to `/help`?

srerickson: Thanks - I like the redirect idea. Then I can delete `help.md` as well

nicktiberi (approved):

PR #1200, opened 2022-01-26
Fix bug where "Publish" button was incorrectly being hidden if errors on the form

### LINKED ISSUE #1195: Attempting to publish a work with errors encounters a dead end
# Bug Report

## Description
If you are on the last page of the Create New Work form and you attempt to publish a work with errors, it will correctly re-render the form with errors, but the primary action button in the footer changes from "Publish" to "Finish", which alters the action that the backend takes and actually prevents you from fixing the errors and publishing the work.

This bug was caused by d261f471 [here](https://github.com/psu-libraries/scholarsphere/commit/d261f47151ca0febab9509fd1dad1d6649b5dc2e#diff-0462626eb67ec7c14551069786a92f04b9a189be055093f58c5f5ecfcda8de2b)

## To Reproduce
1. Create a new work via the form
2. Add no files
3. Click the Publish button on the last step
4. Behold, the primary action button now says "Finish", and clicking it saves your changes, keeps the work in draft, and takes you back to the dashboard.

## Expected behavior
The button remains "Publish" and fixing the errors on the form publishes the work. This will become more of an issue once #1184 is done.

rschenk:
This closes #1195 which describes this bug in detail. See the writeup there.

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -57,6 +57,22 @@ module Dashboard
           resource.aasm_state = initial_state
         end
 
+        # This controller can be used in two ways, either by a user to publish
+        # a new WorkVersion (most common scenario), or by an admin to edit an
+        # already published version. When we render the form (and re-render the
+        # form with errors) we need to know which action the user is
+        # attempting, so we can render the proper buttons.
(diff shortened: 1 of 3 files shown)


nicktiberi (approved):
Tested locally with and without this fix and it looks good! 👍

PR #1205, opened 2022-01-28
adds opt out of monthly email reports checkbox to profile

### LINKED ISSUE #1142: Ability to opt out of monthly email reports
# Feature Request

Users should be able to opt out of receiving monthly emails about article downloads

On the profile page, there should be a checkbox for controlling this. Something like this:

__Email Settings__:
- [ ] Receive monthly report on downloads and views

banukutlu:
Closes #1142

### DIFF
--- a/app/controllers/dashboard/profiles_controller.rb
+++ b/app/controllers/dashboard/profiles_controller.rb
@@ -30,7 +30,8 @@ module Dashboard
             :orcid,
             user_attributes: [
               :id,
-              :admin_enabled
+              :admin_enabled,
+              :opt_out_stats_email
             ]
           )
       end
--- a/app/views/dashboard/profiles/_user_fields.html.erb
+++ b/app/views/dashboard/profiles/_user_fields.html.erb
@@ -1,0 +1,11 @@
+<%= form.fields_for :user do |user_fields| %>
+
+  <div class="form-group">
+    <div class="form-check">
+      <%= user_fields.check_box :opt_out_stats_email, class: 'form-check-input' %>
+
+      <%= user_fields.label :opt_out_stats_email, class: 'form-check-label' %>
+    </div>
+  </div>
+
+<% end %>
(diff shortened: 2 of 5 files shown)


nicktiberi:
I tested this locally and it works, but I am a little confused on the wording + the name of the field.

The name of the field is `opt_out_stats_email`, which to me, implies that if that value is true (the checkbox is checked), the user is opting out and does not want to receive the monthly email.

However, the text is `Receive monthly report on downloads and views`, which implies that if the checkbox is checked, they are actually opting in, and want to receive the monthly email.

So maybe this should actually be inverted, and that checking the box should set `opt_out_stats_email` to `false`, and unchecking it should set it to `true`. Or maybe to be less confusing, we should change the name of the field to `receive_stats_email` to better align with the help text.

We also may need to invert the check in `user.rake` depending on how we proceed with this: https://github.com/psu-libraries/scholarsphere/blob/develop/lib/tasks/user.rake#L8

What do you think?

banukutlu:
> I tested this locally and it works, but I am a little confused on the wording + the name of the field.
> 
> The name of the field is `opt_out_stats_email`, which to me, implies that if that value is true (the checkbox is checked), the user is opting out and does not want to receive the monthly email.
> 
> However, the text is `Receive monthly report on downloads and views`, which implies that if the checkbox is checked, they are actually opting in, and want to receive the monthly email.
> 
> So maybe this should actually be inverted, and that checking the box should set `opt_out_stats_email` to `false`, and unchecking it should set it to `true`. Or maybe to be less confusing, we should change the name of the field to `receive_stats_email` to better align with the help text.
> 
> We also may need to invert the check in `user.rake` depending on how we proceed with this: https://github.com/psu-libraries/scholarsphere/blob/develop/lib/tasks/user.rake#L8
> 
> What do you think?

Oh you are right, I will invert the checkbox value. Changing the field name would need a migration and it is used in couple place, I think it is easier to just invert the value.

banukutlu:
@nicktiberi I ended up just changing the language of the checkbox rather than reversing the logic, seemed cleaner this way.

nicktiberi (approved):
😎

PR #1215, opened 2022-02-01
Prevent "Penn State Only" WorkVersions from being published with an open access license

### LINKED ISSUE #1184: Prevent submissions that are "Penn State Only" with an open access license.
# Feature Request
Depositors sometimes select "Penn State Only" access and apply a license that would allow the Libraries to share the work publicly (i.e., any license other than "rights reserved"). In these case, the access setting and the license conflict with one another and we should do a better job to notify users about this conflict so they can change either the license or the access setting.

If a user selects "Penn State Only" access and an OA license (anything other than "In Copyright (Rights Reserved)") they get a warning:

> The access and license settings conflict with each other. Please set access to "Public" (preferred), or set the license to "In Copyright (Rights Reserved)." 

This validation should not prevent updates by administrators to existing works where the access and license settings are in conflict. 

We use URLs to identify licenses in `config/authorities/licenses.yml`. We might use a setting like the following to configure the validation above:

```yml
psu_access_required_license: [https://rightsstatements.org/page/InC/1.0/]
```

rschenk:
This is conceptually pretty simple: on the Publish tab, if you choose open access then you can pick any license from the dropdown; if you choose Penn-State only access then you can only choose a subset of those licenses.

If that's all there was to it, this would be trivial. Unfortunately, because we have legacy data in the system that may be in violation of this new rule, and because Admins still need the ability to edit those legacy works, we need to make this validation conditional upon the current logged in user.

We don't have an existing pattern for doing this in scholarsphere, this is actually the first instance of a validation that's different between a plebeian and an admin. The idea that I had was to use Rails validation contexts ([Rails docs](https://guides.rubyonrails.org/active_record_validations.html#on) and [a blog](https://boringrails.com/tips/activerecord-validation-context)). 

On the one hand they work. And they work pretty well in this case. My beef with them is that they can only be additive, in other words you can only use them to add more validations and you cannot use them to remove existing validations. I would have rather had admins been the special case, but because the validations need to be _less restrictive_ for admins, they must instead be the default case, and regular users must be the special case. I don't care for that, but I also can't figure out a better way.

Closes #1184

### DIFF
--- a/app/controllers/dashboard/form/base_controller.rb
+++ b/app/controllers/dashboard/form/base_controller.rb
@@ -16,9 +16,9 @@ module Dashboard
           end
         end
 
-        def process_response(on_error:)
+        def process_response(on_error:, validation_context: nil)
           respond_to do |format|
-            if save_resource
+            if save_resource(validation_context: validation_context)
               format.html do
                 redirect_upon_success
               end
@@ -78,10 +78,10 @@ module Dashboard
           resource_klass.model_name.param_key
         end
 
-        def save_resource
+        def save_resource(validation_context: nil)
           @resource.indexing_source = nil # uses the default source
           @resource.update_doi = (publish? || finish? || save_and_exit?)
-          @resource.save
+          @resource.save(context: validation_context)
         end
     end
   end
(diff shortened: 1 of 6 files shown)


ajkiessl (approved):
We have a similar "problem" in ETDA, where the admins can avoid certain validations that regular users can't.  We do the same thing and trigger the more restrictive validation when the regular users are submitting, so this might just be the best way to do this.  Otherwise, everything looks good.

PR #1216, opened 2022-02-02
Prevent works from being added to a collection multiple times

### LINKED ISSUE #1203: Prevent users from adding works to a collection that are already in the collection
# Bug Report

## Description
When creating a collection, a work can be added to that collection multiple times.

## To Reproduce
1. Create a new collection, or modify an existing collection.
2. Add a work to the collection.
3. Add the same work to the collection again.
4. Save the collection.
5. The work is now in the collection twice.

## Expected behavior
A work which is already in the collection should not be allowed to be added to the collection again.

nicktiberi:
Closes #1203.

### DIFF
--- a/app/javascript/controllers/search-works_controller.js
+++ b/app/javascript/controllers/search-works_controller.js
@@ -13,8 +13,13 @@ export default class extends Controller {
         url: endpoint,
         dataType: 'json',
         processResults: (data) => {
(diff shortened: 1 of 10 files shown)


nicktiberi:
This is ready for review.

rschenk (approved):

PR #1218, opened 2022-02-02
makes ORCid readonly.

### LINKED ISSUE #518: Save OrcID in an actor record
Presently in ScholarSphere OrcID is a user editable field. 

We populate a User record with metadata from AzureAD, and an Actor record with a combination of the ScholarSphere User record, and Penn State's `search-service`

AzureAD, and `search-service` do not present a members OrcID. We can populate a users OrcID via `directory-service` 

When saving a user record we can 'refresh' or add OrcID by calling the `directory-service` API. 

WDLL

- [ ] get an API token with read access to directory-service. 

- [ ] write a service that calls `directory-service` to refresh an Actor's metadata. we can just grab `orc-id` as well.

whereismyjetpack:
- updates psu_identity to 0.4.0
- on login, an actor record gets it's ORCiD from `directory-service`


app/views/form_fields/_text.html.erb
```diff
@@ -6,6 +6,8 @@
                       placeholder: true,
                       required: local_assigns[:required].presence,
                       data: local_assigns[:data],
+                      disabled: local_assigns[:disbled].presence,
```
nicktiberi: Typo: `disbled` -> `disabled`

app/models/user.rb
```diff
@@ -68,6 +68,11 @@ def self.from_omniauth(auth)
       actor.email = actor.email.presence || auth.info.email
       actor.given_name = actor.given_name.presence || auth.info.given_name
       actor.surname = actor.surname.presence || auth.info.surname || auth.info.family_name
+      begin
+        actor.orcid = PsuIdentity::DirectoryService::Client.new.userid(auth.uid).orc_id
+      rescue StandardError => e
+        logger.error(e)
+      end
```
nicktiberi: Putting this try/catch in its own private method might avoid having to ignore the cyclomatic complexity issue in `user.rb` that Rubocop is complaining about.

config/locales/en.yml
```diff
@@ -202,7 +202,7 @@ en:
         version_name: For datasets or software using semantic versioning (e.g., x.y.z)
         rights: The license you select determines how your work can be used by others. [Read More](https://docs.scholarsphere.psu.edu/guides/licenses/)
       actor:
-        orcid: "This is a 16-digit identifier in the format ####-####-####-####"
+        orcid: "To link your ORCiD please visit [This link](https://guides.libraries.psu.edu/orcid)"
```
nicktiberi: For accessibility, it might be better to make this whole text a link and have it say something like "Click here to link your ORCiD" — though @srerickson might have other preferences.

app/models/actor.rb
```diff
@@ -61,11 +61,6 @@ class Actor < ApplicationRecord
             uniqueness: {
               case_sensitive: false,
               allow_nil: true
-            },
-            format: {
-              with: /\A\d{15,15}[0-9X]{1,1}\z/,
-              message: 'must be valid',
-              allow_nil: true
```
nicktiberi: Are we now just going to assume all ORCIDs in the system are valid? I guess even in the case where we are adding an actor in the create/edit work form, we are going out to the ORCID API to validate that ORCID, so we could safely assume it is in the correct format?

I'm guessing I'm just wondering if we should keep the format validation/tests as a sanity check.

app/models/user.rb
rschenk: I agree with Nick, let's move this to its own method

app/views/dashboard/profiles/edit.html.erb
```diff
@@ -19,7 +19,7 @@
         <%= render 'form_fields/text',
                    form: form,
                    attribute: :orcid,
-                   data: { controller: 'inputmask' } %>
+                   readonly: true %>
```
rschenk, line 23: In addition to this, we should also remove orcid from the list of permitted attributes in the controller that handles this form. We can't just rely on the front end to enforce this rule, we should also have something over on the server side too.

rschenk (changes requested):

app/models/actor.rb
rschenk: @nicktiberi I can go either way on this one. My thoughts on validation, is that they are something for user-submitted data, and produce an error message for users to fix their issues. Since the orcid is now coming from PSU Identity, I think we can make the case that this validation can go away. BUT, I would like to see our code doing some type of checks and/or formatting of what comes back from the PSU Identity service, to make sure that all of our orcids are stored in a canonical way. It's possible that how PSU Identity stores orcids and how we want to store them are different

app/models/user.rb
rschenk: I would also feel more comfortable if we did some amount of parsing or formatting on the orcid that PSU Identity returns back. It's possible that PSU Identity could return us back either a malformed orcid, or an orcid in a different format than the canonical way that we store them. Because we have logic that looks up orcids in the db, we need to ensure that they are all stored in the same format as the lookup services are using.

whereismyjetpack: It seems like the safest thing to do would be to Run the id though `OrcidId(orcid).to_s` and store that value, then `.uri` for display?

rschenk: Yup, that's how I would do it. We (currently) want the value that's stored in the database to be just the numbers without any hyphens or the complete uri and running it through our `OrcidId` class is a great way to do that.

I know we talked about whether to store the full uri or not in the database. To me that's an entirely separate issue and would require processing all the current orcids in the database, as well as any services or apis that might search for orcids. I don't think we want to go anywhere near that right now. (Tangent: I don't think we should do it at all, I think we should only store the digits)

config/locales/en.yml
srerickson: Don't have a strong preference - I'm okay with making the whole line of text a link.

app/models/user.rb
srerickson: reminder that orcids are not just digits.

rschenk: "X" is a digit in Roman numerals

app/models/user.rb
```diff
@@ -108,4 +104,18 @@ def guest?
...
+  private
+
+  def self.directory_orcid(uid)
+    begin
+      linked_orcid = PsuIdentity::DirectoryService::Client.new.userid(uid).orc_id
+    rescue StandardError => e
+      logger.error(e)
+    end
+
+    OrcidId.new(linked_orcid).to_s
```
rschenk, line 110: This line should go inside the `begin...` stanza, otherwise you are trying to parse a nil value, which might work but shouldn't happen

rschenk (changes requested):

app/models/user.rb
whereismyjetpack: _doh_

.rubocop.yml
```diff
@@ -27,6 +27,7 @@ Metrics/BlockLength:
 
 Metrics/CyclomaticComplexity:
   Exclude:
+    - 'app/models/user.rb'
```
nicktiberi, line 30: With the change you in `user.rb`, pretty sure this can go away now.

whereismyjetpack: Yeah, i'm still whack-a-moling this thing :(

whereismyjetpack: Turns out, user.rb is still complex, it's only over by one, i'm tempted to leave it?

whereismyjetpack:
I _think_ given the tests go green, this is ready to go. This took far too long, and i'm sorry for that. 

I added actor.orcid.presence || to the setting of the orcid to match the other values, and in the unlikely scenario directory-service is down, or a key revoked, we won't override their database value with a nil.

.rubocop.yml
nicktiberi: Ah — I guess that single `||` in the ORCID assignment pushed Rubocop over the edge 🤯 

I'm fine with leaving this.

rschenk (approved):

nicktiberi:
Thanks @whereismyjetpack for this great work! 🎉

PR #1224, opened 2022-02-07
New service to merge a collection and its works into a single work

### LINKED ISSUE #1185: automation for merging works in a collection
# Feature Request

An unfortunate consequence of ScholarSphere's previous implementation of Collections is that there are many collections consisting of single-file works with filenames for titles. Here are some examples:

- [Soil Properties and Class 100m Grids United States](https://scholarsphere.psu.edu/resources/ea4b6c45-9eba-4b89-aba6-ff7246880fb1)
- [Callahan-Flintoft, C., & Wyble, B. (2017)](https://scholarsphere.psu.edu/resources/6bc12ca3-4d56-4e96-9558-dcfdb49e0cbe)
- [Leeman et al. 2014 JGR Data](https://scholarsphere.psu.edu/resources/8531040d-c55c-4f7e-a7c3-6bb57f623c08)

These collections shouldn't be collections, and their works shouldn't be works (They make my eyes bleed). Ultimately, I want to merge the works in these collections into a single work where possible. Doing this by hand would be too labor-intensive and error-prone, so I'd like to automate it as much as possible. What I have in mind is a rake task or the like to trigger a "merge collection" automation (case-by-case).

The automation would take a collection as its primary input; it would determine if the works in the collection can safely be merged, and it would perform the merge, delete the existing works, and add the new work to the collection. Details below:

To be safely merged, works in a collection MUST:
- all have a single file with a filename that maps in a predictable way to the title of the work [(for example)](https://scholarsphere.psu.edu/resources/62f10ddb-4b88-410e-9198-ae721c6f0044).
- all have unique titles in the context of the collection.
- all have the same access settings, license, work type, and publication date.

To be safely merged, works in a collection SHOULD:
- All have the same description
- All have the same creators and keywords (order may differ because of a bug in SS3)
- All have only one version

(The "SHOULD" checks can be overridden with a "force" parameter)

Works are merged by creating a new work with the following properties:
- Same title and description as the collection
- Same work type, access settings, license, publication date, etc. as shared by all the existing works.
- Files from each of the existing works, named for the existing work title (not the filename of the file in the existing work).

Once the work is created, remove the existing works and add the new work to the collection. The result should be a collection with a single work.

We will want to test this extensively in QA before using it in production.

nicktiberi:
Closes #1185.

This service will be used via rake tasks to merge collections with many single-file works to be merged into a single work with many files.

Rake task examples:

```
# merge a collection, but don't delete anything
bundle exec rake resource:merge_collection[<uuid of collection>]

# merge a collection and delete original collection and works
bundle exec rake resource:merge_and_delete_collection[<uuid of collection>]
```


nicktiberi:
After looking at the Code Climate issues a bit more, I think I agree with Ryan that they are not worth fixing. So at this point, this PR is ready for review.

Since Ryan did most of the work on this PR, @banukutlu, would you mind taking a look at it?

app/services/merge_collection.rb
```diff
@@ -0,0 +1,221 @@
...
+      end
+
+      if work_metadata_diff.any?
+        diff = work_metadata_diff.inspect
+        errors << "Work-#{canonical_work.id} has different work metadata than Work-#{w.id}\n#{diff}"
+      end
+
+      if !w.representative_version.published?
+        errors << "Work-#{canonical_work.id} is not published"
+      end
```
banukutlu: @nicktiberi wondering if it makes sense to move these to `validate_work(w)` method and also move line 70, 71 in that method too?

app/services/merge_collection.rb
```diff
@@ -0,0 +1,221 @@
...
+        .map(&:file_resources)
+        .flatten
+
+      work = Work.build_with_empty_version(
+        work_type: canonical_work.work_type,
+        embargoed_until: canonical_work.embargoed_until,
+        doi: canonical_work.doi,
+        depositor: canonical_work.depositor,
+        proxy_depositor: canonical_work.proxy_depositor,
+        deposited_at: deposited_at
```
banukutlu: @nicktiberi  is merging mostly gonna be for older collections? if for newer as well, maybe need to add the new column `notify_editors` to get the value from canonical work as well, that PR just merged. 🤔

banukutlu (changes requested):

app/services/merge_collection.rb
nicktiberi: Good call - this will be mostly for older works, but we should be thorough and validate/merge all fields that are part of the canonical work.

app/services/merge_collection.rb
nicktiberi: I like that idea - this might even make Code Climate a bit happier 🙂

app/services/merge_collection.rb
```diff
@@ -0,0 +1,228 @@
...
+      end
+
+      if delete_collection?
+        collection.destroy!
+
+        collection.works.each do |work|
+          work.versions.each do |version|
+            DestroyWorkVersion.call(version, force: true)
+          end
+        end
```
banukutlu, line 48: @nicktiberi this potentially could move to its own small method as well, but totally up to you, I think the PR looks good overall.

banukutlu (approved):
lgtm 👍🏻  just added one comment.

rschenk (approved):
I like these suggestions that @banukutlu made, the code is definitely easier to follow now

PR #1225, opened 2022-02-09
Adding a draft work to a collection no longer 500s

### LINKED ISSUE #1211: Adding a draft work to a collection fails silently
# Bug Report

## Description
When adding a work to a collection, draft works are included in the "Add works to collection" search results, however selecting them errors silently (there is a `500 server error` reported in the browser console). We should either support draft works in collections (preferred) or notify the user that the work cannot be added because it is a draft.

## To Reproduce
1. go to a collection's update works page
2. search for a work that is a draft and click to add it to the collection

## Expected behavior
- draft work is added or user notified of problem

rschenk:
Very simple one, will update some other functionality around drafts in collections in future issues.

Closes #1211

### DIFF
--- a/app/views/dashboard/form/members/_work_membership_fields.html.erb
+++ b/app/views/dashboard/form/members/_work_membership_fields.html.erb
@@ -6,7 +6,7 @@
      {}
    ) %>
 <%- index = form.try(:index) || 0 %>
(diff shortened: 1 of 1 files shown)


nicktiberi (approved):
🆒

Issue #1114, opened 2021-08-04
Thumbnails for works and collections


awead:
# Feature Request

- [x] Works and Collections can have thumbnails with same size and placement as current placeholder graphics.
- [x] UI on work settings page to upload a thumbnail image, with recommended dimensions.
- [x] Uploaded thumbnails are automatically scaled/cropped to necessary sizes (of current placeholder graphics).
- [x] Thumbnail appears on search results (list of resources view)
- [x] Thumbnail appears on resource display page (show resource view). See note below.

Nice, but not required:
- [x]  Thumbnail upload UI has an option to auto-generate a thumbnail from the content. (i.e., fist page of PDF).

# Placement of thumbnail on work/collection display page:

Thumbnails are not part of the existing resource display pages, so we will need to do some basic design work. For a first pass, let's include a "thumbnail" section in the metadata block on the right-side, below the title. We may want to explore other options.

ajkiessl:
@srerickson I added this checkbox to the settings page for Works and Collections.  It allows users to toggle the thumbnail display 'on' or 'off'.  If it is toggled 'off', then one of the icons that we've been displaying will show in its place.

![Screen Shot 2022-02-08 at 3 50 18 PM](https://user-images.githubusercontent.com/32677188/153074263-9b1eee91-392b-4156-915f-53a86eaabe46.png)

It is defaulted to 'off'.  I figured we could roll this out at the end of this sprint with it defaulted to 'off', and if it seems like thumbnails are generating properly, then maybe default to 'on' later.

I'm going to get to work on the feature giving users the ability to upload their own thumbnail next, but that'll be in a separate branch/PR and probably won't be released until the end of the next sprint.

srerickson:
@ajkiessl, this looks great. What are your thoughts on automatically generating thumbnails for all new submissions and changing the checkbox to "Use auto-generated thumbnail ..." (default to unchecked). Could we do that this sprint? The thing is, people will want to see the thumbnail before using it.

PR #1227, opened 2022-02-10
Preview/thumbnails

ajkiessl:
This is a big one, so I added everyone as reviewers.   Not everyone needs to review, though.  In this PR:

1.) Adds job using Shrine to generate thumbnail derivatives from uploaded files.  Right now we can process thumbnails for image types (like `.png`s and `.jpeg`s), `.pdf`s, and `docx`s.

2.) Users can toggle the display of an auto-generated thumbnail 'on' or 'off' from the Work and Collection settings pages.  It is defaulted to 'off' whenever a Work or Collection is created. The auto-generated thumbnail is displayed in the settings page so the user can see what it looks like before toggling on.  The 'Thumbnail' section of the settings page will not render if the Work or Collection does not have any thumbnails.

3.) Thumbnails (if one exists and auto-generate is toggled on) will display in place of the placeholder icons in the resource cards that show in the catalog, dashboard, featured page, and collections.  If no thumbnail exists or auto-generate is toggled off, the placeholder icons will show.

There are some system packages required to get the derivative creation to work if you are running the app locally.  The tests that require these packages are skipped locally with `skip: !ci_build?`, so you don't need to install this stuff to run the tests and develop.  If you do want to get the derivative creation working, the dockerfile documents what should be installed if you want this to work locally:
https://github.com/psu-libraries/scholarsphere/blob/preview/thumbnails/Dockerfile#L6-L10
The circle ci config also has this info.  You'll also need to configure `ImageMagick` to handle pdfs and docxs.

This PR completes checkbox 1, 3, 4, and 6 in #1114.  I'll finish up the rest in a separate branch/PR in the next sprint.  Also, we can work on handling more file types in the next sprint.

@whereismyjetpack Feel free to add to this.


app/jobs/shrine/promotion_job.rb
```diff
@@ -6,7 +6,17 @@ class Shrine::PromotionJob < ApplicationJob
   def perform(record:, name:, file_data:)
     attacher = Shrine::Attacher.retrieve(model: record, name: name.to_sym, file: file_data)
     attacher.atomic_promote
+    Shrine::ThumbnailJob.perform_later(record) if perform_thumbnail?(record)
   rescue Shrine::AttachmentChanged, ActiveRecord::RecordNotFound
     # attachment has changed or record has been deleted, nothing to do
   end
+
+  def perform_thumbnail?(record)
```
whereismyjetpack: This might be better in the FileResoruce as something like `thumbnailable?` and we could use it in IRB, rake tasks etc. ?

db/schema.rb
```diff
@@ -13,7 +13,6 @@
 ActiveRecord::Schema.define(version: 2022_02_07_221742) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
whereismyjetpack: was this intentional to remove?

ajkiessl: Whoops, not sure why that got removed.  I'll add it back in.

app/jobs/shrine/promotion_job.rb
ajkiessl: Yeah, I'll see if I can move it there.

app/components/thumbnail_component.rb
```diff
@@ -3,26 +3,47 @@
...
+      else
+        'thumbnail-icon col-xxl-6 ft-work__img'
+      end
     else
-      'thumbnail'
+      if thumbnail_url.present?
+        'thumbnail-image'
+      else
+        'thumbnail-icon'
+      end
```
nicktiberi: I think you can avoid this nesting and just do:

```
if featured?
   # do stuff
else if thumbnail_url.present?
  'thumbnail-image'
else
  'thumbnail-icon'
end
```

Then you can remove the rubocop `Style/IfInsideElse` rule.

app/components/thumbnail_component.rb
```diff
@@ -3,26 +3,47 @@
...
   end
 
-  def icon
-    icon_map.fetch(icon_key, 'bar_chart')
+  def thumbnail
+    if thumbnail_url.present?
+      "<img src=#{thumbnail_url}></img>"
+    else
+      "<i class='material-icons material-icons--thumbnail'>#{icon}</i>"
+    end
```
nicktiberi: IMO, I think it might be cleaner to put this logic and the HTML generation inside `thumbnail_component.html.erb`. Then you won't have to use `html_safe` in the template. We do conditional rendering in other existing components like `files_message_component` and `work-version_detail_component`.

app/components/thumbnail_component.rb
nicktiberi: Alternatively you could build up an array of classes based on `thumbnail_url.present?` and `featured?` and join them at the end, which may reduce the cognitive complexity a bit.

app/jobs/shrine/promotion_job.rb
whereismyjetpack: thanks! i wrote this silly code, and you fixed it! 🎉

.rubocop.yml
```diff
@@ -36,6 +36,10 @@ Rails/DynamicFindBy:
   Exclude:
     - 'spec/models/file_resource_spec.rb'
 
+Style/IfInsideElse:
+  Exclude:
+    - 'app/components/thumbnail_component.rb'
+
```
nicktiberi: I think this could go away now with your latest change to the thumbnail component.

app/models/collection.rb
```diff
@@ -146,8 +146,25 @@ def work_type
...
+    auto_generate_thumbnail? ? auto_generated_thumbnail_url : nil
+  end
+
+  def auto_generated_thumbnail_url
+    thumbnail_urls.last
+  end
+
+  def thumbnail_present?
+    thumbnail_urls.present?
+  end
```
nicktiberi, line 161: Could these methods (and the equivalent methods in `work.rb`) go in a `Thumbnailable` module that could be included in both `Collection` and `Work`? Just spitballing here; I'm not sure if this is the Ruby way to do things, but it seems like it could avoid some duplication of these identical methods.

nicktiberi (commented):
I just had a couple more comments, but overall, this looks great to me!

Great work @ajkiessl  and @whereismyjetpack! 🥇

.rubocop.yml
ajkiessl: Fixed

app/models/collection.rb
ajkiessl: I could go either way on this.  On one hand, we have duplication that we could resolve with this abstraction.  On the other hand, some of these features are still developing, and the abstraction might be a bit premature.  This is my inexperienced opinion, though.  @rschenk would you create a module here?

app/components/thumbnail_component.rb
```diff
@@ -3,21 +3,32 @@
...
+      else
+        'thumbnail-icon col-xxl-6 ft-work__img'
+      end
+    elsif thumbnail_url.present?
+      'thumbnail-image'
     else
-      'thumbnail'
+      'thumbnail-icon'
     end
   end
```
rschenk: I like @nicktiberi's suggestion of doing an array here

```ruby
def html_classes
  classes = []
  classes << if thumbnail_url.present? 
               'thumbnail-image' 
             else
               'thumbnail-icon'
             end
  classes << 'col-xxl-6 ft-work__img' if featured?

  classes.join(' ')
end
```

app/models/collection.rb
rschenk: I agree with both of you. To Nick's point, yes, a Thumbnailable module would be the Ruby way to share these implementations. To Alex's point, Adam and I often had a rule of thumb to not go too nuts on refactoring until the third time we were doing something. Alex I like how you used an rspec shared example group to at least enforce the contract between all the thumbnailable things, which I think is good enough for now. Maybe it would warrant a little comment on the various thumbnailable classes pointing out that it's a good place for a future refactor?

app/models/collection.rb
```diff
@@ -146,8 +146,25 @@ def work_type
...
+
   private
 
+    def thumbnail_urls
+      works.map { |work| work&.latest_published_version&.file_resources }
```
rschenk: Fun Ruby method: [flat_map](https://apidock.com/ruby/Enumerable/flat_map) that combines a .map and .flatten

app/models/file_resource.rb
```diff
@@ -62,6 +62,19 @@ def indexing_source
     @indexing_source ||= SolrIndexingJob.public_method(:perform_later)
   end
 
+  def thumbnail_url
+    file_attacher.url(:thumbnail)
+  end
+
+  def thumbnailable?
+    mime_type = file_data['metadata']['mime_type'] || ''
```
rschenk: Might maybe want to use `file_data.dig('metadata', 'mime_type') || ''` here, just in case there is no `'metadata'` key in file_data. Maybe it's guaranteed to be there with Shrine, but can't hurt

rschenk (dismissed):
I think this looks really excellent, Alex. I added a few stylistic ideas here and there but overall it looks great. I'll approve the PR, up to you whether you want to change your code with my suggestions

ajkiessl:
Thanks @rschenk.  I'll try to get these changes in before lunch.  Then, we should be ready to go!

nicktiberi (approved):
🚀

PR #1236, opened 2022-02-15
Updates to merge/delete collection services

nicktiberi:
Re: #1185.

- `delete_collection` now lives in its own separate service/task
- `force_merge_collection` now allows the user to ignore mismatches in:
  - keywords
  - creators
  - descriptions
- merged work's files will be named after the work the file was originally associated with
- merged WorkVersion's keywords + description come from the collection

### DIFF
--- a/app/services/delete_collection.rb
+++ b/app/services/delete_collection.rb
@@ -20,16 +20,13 @@ class DeleteCollection
     ActiveRecord::Base.transaction do
       works = collection.works.map(&:itself)
 
-      collection.works = []
-      collection.save!
+      collection.destroy!
 
       works.each do |work|
-        work.versions.each do |version|
-          DestroyWorkVersion.call(version, force: true)
+        while work.versions.any?
+          DestroyWorkVersion.call(work.versions.last, force: true)
         end
       end
-
-      collection.destroy!
     end
 
     @successful = true
--- a/app/services/delete_collection.rb
+++ b/app/services/delete_collection.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class DeleteCollection
+  attr_reader :uuid
+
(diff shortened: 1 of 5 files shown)


nicktiberi:
This is ready for review aside for the two failing tests. If anyone sees what I'm missing, please let me know. I'm trying to overwrite the keywords and description for a specific work version to induce an error, but it seems at some point my changes are either not taking hold or are being overwritten. Probably just doing something v dumb 🤦

nicktiberi:
Fixed up the specs, so this is ready for review.

rschenk (changes requested):
One thing we forgot to handle was our `LegacyIdentifiers` which is used to maintain old URLs from SSv3 into SSv4. These are stored in a separate table. The collection that we are merging may have some of these, which probably should be updated to point to the newly created Work instead. There is a problem of _when_ to make the update. You can't have two copies of a legacy identifier sitting in the same table pointing to two different things, but the workflow we have now of creating the new work and putting it into draft mode also does not provide us with a good time/place to make the switcheroo

banukutlu (dismissed):
this looks good to me 👍🏻   my only comment is about `merge_collection` that it could be broken into smaller methods but I understand if you prefer to have the whole logic in one place.

nicktiberi:
@rschenk - thanks for bringing up the point about legacy identifiers. We talked about this at standup today and our current plan is now to migrate all the IDs (UUID, DOI, Legacy Identifiers) from the collection to the merged work after the old collection has been deleted. I will open a new ticket for that functionality.

I removed the DOI migrating code form this PR, so I think it's good to go again.

banukutlu (approved):
👍🏻

rschenk (approved):
:+1:

PR #1237, opened 2022-02-15
Preview/md5

### LINKED ISSUE #708: Include File MD5 with Other File Info
It's important for users and curators to be able to confirm that the deposited file is the same as the file they have. The md5 checksum should be accessible through the UI. I think we can just add it along with other per-file info (size, content type, date)

UI Note:
- this could look like github commit hashes, show the abbreviated checksum: `md5: c9f4de...`
- mousing over the checksum shows the full value
- clicking copies to clipboard

### LINKED ISSUE #1207: Checksums in file characterization
# Feature Request
As part of file characterization process, we should be calculating the following checksums for each file

- `md5`: because S3 uses md5 for etags (for files under 5GB), so it's good to have for comparison
- `sha256`: because it's the modern standard

banukutlu:
Closes #708 
Closes #1207

### DIFF
--- a/app/components/file_signature_component.html.erb
+++ b/app/components/file_signature_component.html.erb
@@ -1,4 +1,13 @@
-|
-<span class="meta__signature" data-toggle="tooltip" data-placement="top" title="<%= tooltip %>">
-  <%= type %>: <%= signature_display %>
+| 
+<span data-controller="clipboard">
+  <span class="meta__signature" data-target="clipboard.source" data-signature="<%= value %>">
+    <%= label %>: <%= signature_display %>
+  </span>
+  <button class="btn btn-sm btn-link px-0"
+          data-toggle="tooltip"
+          data-placement="top"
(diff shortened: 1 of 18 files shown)


app/components/file_signature_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<span> | </span>
+<span data-controller="clipboard">
+  <span class="meta__signature" data-target="clipboard.source" data-signature="<%= value %>">
```
nicktiberi: I wonder if we should rename `data-signature` to something more generic, since the clipboard controller could be used in other components and not just this signature component. We could maybe rename it to something like `data-source` (open to other suggestions) and then make the corresponding change in `clipboard_controller.js`.

nicktiberi (approved):
LGTM! 🥇

ajkiessl (approved):
Looks good!

PR #1239, opened 2022-02-16
WIP reports

### LINKED ISSUE #1210: Downloadable Reports
# Feature Request
We should make it easier for admins and users to download reports (`csv` files) about works, downloads, views, etc. This issue describes the reports that should be accessible through the UI.

- Admin reports
  - [x] Monthly Report 
  - [x] All Works
  - [x] All Work Versions
  - [x] All Files
- User Reports
  - [x] Monthly Report 

## Admin reports

These reports should be accessible through a "reports" admin-only page (linked in top-right menu)

### Monthly Reports

These monthly reports are generated each month. Each month's report is individually downloadable from the reports page. We should generate reports for all previous months (as available data permits).  Report includes one work-month per row, with the following columns:

column | description
-- | --
work_id | UUID of the work
month | numeric representation of month (1-12)
year | year (2022)
downloads | aggregated number of downloads for the work in the month
views | aggregated number of views for the work in the month

### All works

Report includes all Works (one per row), with the following columns:

column | description
-- | --
id | work's UUID
depositor | PSU ID
work_type | example: "Artilce"
title | title from most recent version
doi | scholarsphere's doi
deposited_at | Date draft created
deposit_agreed_at | Date of most recent published version
embargoed_until | date
visibility | "Public" or "Penn State Only"
latest_published_version | UUID of published work version (if present)
downloads | aggregated downloads across all versions/files
views | aggregated views across all versions

### All Work Versions
Report includes all Work Versions (one per row), with the following columns:

column | description
-- | --
id | version's UUID
work_id | parent work's UUID
state | published, draft, withdrawn
version_number | number of the version in the work
title | meta field (joined if list)
subtitle | ..
version_name | ..
keyword | ..
rights | ..
description | ..
publisher_statement | ..
resource_type | ..
contributor | ..
publisher | ..
published_date | ..
subject | ..
language | ..
identifier | ..
based_near | ..
related_url | ..
source | ..
~downloads~ | ~aggregated downloads across all files~
views | number of views

### All Files
Report includes all *File Memberships* (join of file resource and work version, one per row), with the following columns:

column | description
-- | --
id | file resource's UUID
version_id | work version's UUID
filename | name of the file
mime_type | MIME type
size | number of bytes
md5 | md5 checksum (once implemented)
sha256 | sha256 checksum  (once   implemented)
downloads | number of downloads

## User Reports

User reports are scoped to the works that users have deposited. They are linked from the user's profile page.

### Monthly Reports

Report includes one work-month per row, with the following columns:

column | description
-- | --
work_id | UUID of the work
title | title of the work
month | numeric representation of month (1-12)
year | year (2022)
downloads | aggregated number of downloads for the work in the month
views | aggregated number of views for the work in the month

rschenk:
This has all the basics for the reports to close #1210 

It has some opportunity for improvement:
`AllWorksReport` and `MonthlyUserWorksReport` currently have some code like this: 
```ruby
      latest_version = work
        .versions
        .max_by(&:version_number)
```
These should be updated to only look at draft or published works and/or ignore withdrawn ones.

Other areas for improvement:
Some of the fields can have multiple values (example: keyword). Right now what happens is these multiple values get encoded like json when being inserted into the CSV row. Seth had mentioned that he would like it double encoded as quotated csv. That's something that should not happen in the Report classes (they know nothing about CSV and should not—what if we wanted to offer reports in json?), it should happen in the controller that encodes the csv. 

The UI needs work, I just did the absolute minimum to demo all of them. Obviously for the User one, we would not have a field for their access id, it would pull `current_user` from the controller.

### DIFF
--- a/app/controllers/dashboard/reports_controller.rb
+++ b/app/controllers/dashboard/reports_controller.rb
@@ -3,8 +3,7 @@
 require 'csv'
 
 class Dashboard::ReportsController < Dashboard::BaseController
-  def show
-  end
+  def show; end
 
   def all_files
     return deny_request unless current_user.admin?
@@ -25,6 +24,7 @@ class Dashboard::ReportsController < Dashboard::BaseController
   end
 
   private
+
     def deny_request
       render json: { message: I18n.t('errors.not_authorized.heading'), code: 401 }, status: 401
     end
@@ -32,14 +32,14 @@ class Dashboard::ReportsController < Dashboard::BaseController
     def generate_csv(report)
       result = CSV.generate(headers: true) do |csv|
         csv << report.headers
-  
+
         report.rows do |row|
           csv << row
         end
       end
 
       date = Time.now.strftime('%Y-%m-%d')
-  
+
       send_data result, filename: "#{report.name}_#{date}.csv"
     end
 end
--- a/app/controllers/dashboard/reports_controller.rb
+++ b/app/controllers/dashboard/reports_controller.rb
@@ -3,27 +3,23 @@
 require 'csv'
 
 class Dashboard::ReportsController < Dashboard::BaseController
-  def all_files
-    return deny_request unless current_user.admin?
+  before_action :authorize_report, except: [:index, :monthly_user_work_versions]
 
+  def index; end
+
+  def all_files
     generate_csv(AllFilesReport.new)
   end
 
   def all_works
-    return deny_request unless current_user.admin?
-
     generate_csv(AllWorksReport.new)
   end
 
   def all_work_versions
-    return deny_request unless current_user.admin?
-
     generate_csv(AllWorkVersionsReport.new)
   end
 
   def monthly_work_versions
-    return deny_request unless current_user.admin?
-
     date = Date.new(
       params[:report_date][:year].to_i,
       params[:report_date][:month].to_i,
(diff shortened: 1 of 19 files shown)


nicktiberi:
@rschenk - I made some changes since you initially opened the PR and I think it's ready to go.

- ignoring withdrawn work versions when checking for latest version
- pull current user's PSU ID to pass into user report
- build more fleshed out UI
- write specs for controller and feature

Also, we talked in standup today and Seth is ok with the array values being represented in the CSV as `["value1", "value2"]` etc. so no need to do any special processing for those values at this time.

app/controllers/dashboard/reports_controller.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class Dashboard::ReportsController < Dashboard::BaseController
+  def all_files
+    return deny_request unless current_user.admin?
```
rschenk: I would move all these into a `before_action`

Have a look at `app/controllers/dashboard/collections_controller.rb` for an example of how this is done. You can use `except: [:monthly_user_work_versions]` on the end of the before action to authorize only the admin ones

rschenk (commented):
Looks good to me! I can't approve it since I am also one of the authors

ajkiessl (approved):
Looks good!

PR #1244, opened 2022-02-22
Track external apps

### LINKED ISSUE #1214: Record the Application that Created a Work Version
# Feature Request

For reporting purposes, we should record the mechanism through which a work (version) was created. For example, if a work was deposited using the Researcher Metadata Database integration, that fact should be represented in the data model for the work (version) itself. (Currently, you have to dig through the `papertrail` logs to get this information).

Here is a suggested solution, but alternative proposals are welcome:

- Add a new field to the `WorkVersion` model called `depositing_application_id` that can be set to the the id of the ExternalApplication that created the work version.
- Update API controllers to set this value when a work version is created.
- Migrate the database to populate this field for existing work version created via external applications.

rschenk:
A little bit of background: we have a table/model called ExternalApps that links to all of our API keys. When our API is hit with a certain key, we can tell which "external app" is performing the action. Previously, if a work was created via the ingest API, either from the ScholarSphere3 migration (or coming in from RMD?), the only record of said external app was buried way down in our auditing package, called PaperTrail. Seth wants the linkage to be a little more explicit, so it's easy to tell which WorkVersions were created by SSv3 migration, RMD, or by active ScholarSphere users.

This PR does a few things: creates an optional FK from the work version to the external app that may have created it. It also updates the service object used by the ingest api to pass the external app through to the new work version that it creates. Finally there is a migration to backport all of the old data to use this new fk.

The key to understanding the migration code is this fact: PaperTrail tracks the user that made the change by the `whodunnit` string. Because actions in SS could either be performed by `User`s or by `ExternalApp`s, we store a `GlobalID` in this column, which looks like `"gid://scholarsphere/User/23"` for a `User` with id 23 or `"gid://scholarsphere/ExternalApp/2"` for an `ExternalApp` with id 2. The migration loads out every PaperTrail version for every WorkVersion, looks for _any_ versions that were performed by an `ExternalApp`, and if it finds any, it stores the first one as the `external_app` on that WorkVersion. The code is actually remarkably small, but pretty dense.

Closes #1214

### DIFF
--- a/app/controllers/api/v1/ingest_controller.rb
+++ b/app/controllers/api/v1/ingest_controller.rb
@@ -52,7 +52,8 @@ module Api::V1
           metadata: metadata_params,
           depositor_access_id: depositor_params,
           content: content_params,
-          permissions: permission_params
+          permissions: permission_params,
+          external_app: api_token.application
         )
       end
 
--- a/app/models/external_app.rb
+++ b/app/models/external_app.rb
@@ -6,6 +6,9 @@ class ExternalApp < ApplicationRecord
            foreign_key: 'application_id',
            inverse_of: 'application'
 
+  has_many :work_versions,
+           dependent: :nullify
+
   validates :name,
             presence: true,
             uniqueness: true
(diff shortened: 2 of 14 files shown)


nicktiberi (approved):
This looks great to me! 🚀

Issue #1230, opened 2022-02-14
Add rel=nofollow to download links


nicktiberi:
# Feature Request

## Describe the solution you'd like
Add `rel="nofollow"` to download links for files to prevent search engines/crawlers/etc from hitting these links and skewing our true download stats.

## What does "done" look like?
All file resource download links will have the `rel="nofollow"` attribute set.

<img width="939" alt="Screen Shot 2022-02-14 at 9 59 51 AM" src="https://user-images.githubusercontent.com/639920/153888455-ff05afea-aa0b-47ea-b5d7-35ed644cf001.png">

srerickson:
@whereismyjetpack, it would be interesting to see how this change affects our metrics. We should try to look into this about a month after it is released.

PR #1245, opened 2022-02-22
Cache download counts per session, per day

### LINKED ISSUE #1222: Improve how we count downloads
# Feature Request

The downloads metrics for ScholarSphere 4.0 are much higher than they were for ScholarSphere 3.0. This may be because we're not doing a good job of filtering "robot" downloads. We might consider taking session information into account like we do for view counts.

rschenk:
As my [comment](https://github.com/psu-libraries/scholarsphere/issues/1222#issuecomment-1047966253) in #1222 states, we are not deduping the download counts in the same way that we are with views. If the same person views the same work twice in one day, it is counted as one single view. However if the same person downloads the same file twice in one day, it is counted as two downloads. This PR dedups the download counts to match the same way we are doing it in the views.

Also while I was investigating this issue, I found a bug in the stats cache code! So this PR fixes that too. 

I don't know if this PR should _close_ the issue linked to above, so I did not do the GitHub auto closing magic. I guess that should be up to @srerickson to decide. Update: closes #1222

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -111,6 +111,7 @@ RSpec/VerifiedDoubles:
     - 'spec/components/work_versions/version_navigation_component_spec.rb'
     - 'spec/components/work_histories/work_history_component_spec.rb'
     - 'spec/controllers/dashboard/work_search_controller_spec.rb'
+    - 'spec/services/session_view_stats_cache_spec.rb'
 
 Style/ClassVars:
   Exclude:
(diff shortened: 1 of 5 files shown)


nicktiberi (approved):
These changes make sense to me. It will be good to get these changes out at the same time as the `rel=nofollow` change so we have a single point in time where we changed how downloads/views are counted, and the re-evaluate in a month or so to see how that affected things.

srerickson:
@rschenk, This sounds good to me too. I also think with this and #1230, #1222 can be closed.

PR #1246, opened 2022-02-22
Service to migrate identifiers from a collection to a work

### LINKED ISSUE #1242: Automation for Migrating Collection Identifiers to a Work
# Feature Request

Related to #1185 - after the collection and its files has been merged into a new work, the new work has been verified by a curator, and the original collection has been deleted, we should migrate the following values from the original collection to the new work:

- DOI
- UUID
- Legacy Identifier(s)

This might simply expand on the existing `delete_collection` rake task so that everything can be done in a single step.

nicktiberi:
Closes #1242.

This service will delete the collection + it works, and then copy the following identifiers over to a specified target work:

- DOI
- UUID
- Legacy Identifiers

### DIFF
--- a/app/services/migrate_collection_ids.rb
+++ b/app/services/migrate_collection_ids.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+class MigrateCollectionIds
+  attr_reader :collection_uuid, :work_uuid
+
+  def self.call(collection_uuid, work_uuid)
+    instance = new(collection_uuid, work_uuid)
+    instance.migrate
+    instance
+  end
+
+  def initialize(collection_uuid, work_uuid)
+    @collection_uuid = collection_uuid
+    @work_uuid = work_uuid
+    @successful = false # updated by #migrate
+  end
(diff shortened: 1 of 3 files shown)


app/services/migrate_collection_ids.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def work
+      @work ||= Work.find_by(uuid: work_uuid)
+    end
+
+    def migrate_legacy_identifiers(legacy_ids)
+      legacy_ids.each do |legacy_id|
+        legacy_id.resource_type = 'Work'
+        legacy_id.resource_id = work.id
```
rschenk: This is a polymorphic relationship here and would be more idiomatic to replace these two lines with `legacy_id.resource = work`. 

Not that it really matters, nor is it worth changing and resubmitting the PR

rschenk (dismissed):
Looks good to me, just one very minor thing I found that's not critical at all

rschenk (approved):

PR #1248, opened 2022-02-22
empty collections should not appear in search results/browse all

### LINKED ISSUE #1226: Empty Collections should not appear in search results/browse all
# Feature Request
ScholarSphere users often create collections before they create the works that will (eventually) go in them. As a result, empty collections are often floating around on the "browse all" page.

WDLL: Collections without any published works should not appear in search results. They are similar to draft works in this respect.

banukutlu:
Closes #1226

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -160,9 +160,9 @@ class Collection < ApplicationRecord
     thumbnail_urls.present?
   end
 
-  def draft?
+  def empty?
     works
-      .select { |work| work.representative_version.aasm_state == 'published' }
+      .select { |work| work.representative_version.published? }
       .empty?
   end
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -160,6 +160,12 @@ class Collection < ApplicationRecord
     thumbnail_urls.present?
   end
 
+  def draft?
+    works
+      .select { |work| work.representative_version.aasm_state == 'published' }
+      .empty?
+  end
+
   private
 
     def thumbnail_urls
@@ -187,7 +193,8 @@ class Collection < ApplicationRecord
         FacetSchema,
         WorkTypeSchema,
         DoiSchema,
-        TitleSchema
+        TitleSchema,
+        CollectionSchema
       )
     end
 
--- a/app/schemas/collection_schema.rb
+++ b/app/schemas/collection_schema.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class CollectionSchema < BaseSchema
+  def document
+    return {} unless resource.respond_to?(:draft?)
+
+    {
+      is_empty_bi: resource.draft?
+    }
+  end
+end
--- a/app/schemas/collection_schema.rb
+++ b/app/schemas/collection_schema.rb
@@ -2,10 +2,10 @@
 
 class CollectionSchema < BaseSchema
   def document
-    return {} unless resource.respond_to?(:draft?)
+    return {} unless resource.respond_to?(:empty?)
 
     {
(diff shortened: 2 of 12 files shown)


app/models/collection.rb
```diff
@@ -160,6 +160,12 @@ def thumbnail_present?
     thumbnail_urls.present?
   end
 
+  def draft?
```
banukutlu: @nicktiberi @rschenk pls let me know if you can think of a better name for this method, it is basically to check if the collection has any published work or not.

app/schemas/collection_schema.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CollectionSchema < BaseSchema
```
banukutlu, line 3: @nicktiberi @rschenk same here, better schema and field name suggestion are more than welcome :) 
also open to suggestions re if this field could be part another schema or not?

app/models/collection.rb
```diff
@@ -160,6 +160,12 @@ def thumbnail_present?
...
+    works
+      .select { |work| work.representative_version.aasm_state == 'published' }
```
nicktiberi: We can use `work.representative_version.published?` here so we don't have to use AASM fields directly. I believe `published?` is an autogenerated method from [AASM](https://github.com/aasm/aasm) (along with `draft?` and `withdrawn?`.

https://github.com/psu-libraries/scholarsphere/blob/main/app/models/work_version.rb#L181-L201

app/models/collection.rb
nicktiberi: Maybe `empty?` just for consistency with the `is_empty_bi` Solr field? And also to distinguish it from the AASM draft state for work versions?

app/schemas/collection_schema.rb
nicktiberi: Caveat that I don't know how these schemas are typically set up/organized, but I like `CollectionSchema` here — it seems like a good place to put anything else that may be specific to collections.

nicktiberi (commented):
Just a few minor suggestions, but overall this looks great to me!

app/schemas/collection_schema.rb
```diff
@@ -0,0 +1,11 @@
...
+  def document
+    return {} unless resource.respond_to?(:draft?)
+
+    {
+      is_empty_bi: resource.draft?
```
nicktiberi: Just a question for my own Solr understanding — when does `is_empty_bi` get populated for all documents in Solr? Does this just magically happen or do we need to run a full reindex or something to populate this field for all collections so the existing collections with no published works don't appear in search results?

app/models/collection.rb
banukutlu: ahh good point, I will use that! thnx!

app/schemas/collection_schema.rb
banukutlu: right, we need to run full reindex, thanks for bringing it up, I forgot to mention that in my comments.

banukutlu: I am open to other suggestions on how to handle it, if you have another idea.

nicktiberi: Thanks for confirming! I think this approach makes sense. I can check on QA to see how long it takes to index all the collections. We can do this post launch and some time during off hours if it takes a long time.

banukutlu: sounds good 👍🏻

app/models/collection.rb
banukutlu: updated in the last commit

nicktiberi (approved):
LGTM!

PR #1252, opened 2022-02-23
Gives users an interface to upload their own thumbnails

ajkiessl:
Leaving this as a draft for now.  I'm going to try to make the thumbnail settings page look a little better.  @srerickson Is also going to take a look at it and provide some feedback.  If anyone else has some recommendations let me know.  It currently looks like this:

Screen Shot 2022-02-23 at 4 18 06 PM

The preview instance is here: https://scholarsphere-upload-thumb.dsrd.libraries.psu.edu

Otherwise, all the backend functionality is working and tested.  If anyone has time and wants to get a head start reviewing the code, go ahead.


whereismyjetpack:
I don't have a lot to add, but I fear/feel that the Thumbnail section needs some sort of logic. i.e if a thumbnail is uploaded, the upload box disappears for example, that might make the section flow a bit?

ajkiessl:
@whereismyjetpack I agree, that could reduce some of the clutter.

srerickson:
Something like this is what I have in mind for work settings config:

Untitled drawing

The little circles under each are radio buttons or mutually exclusive check boxes. The middle option is disabled unless a thumbnail has been uploaded.

ajkiessl:
@srerickson Ok, thanks!  I'll see what I can do.

app/views/dashboard/shared/_thumbnail_form_fields.html.erb
```diff
@@ -0,0 +1,65 @@
...
+      <div>
+        <%= field.radio_button :thumbnail_selection, ThumbnailSelections::DEFAULT_ICON.to_sym %>
+        <%= field.label :thumbnail_selection,
+                        'Use Default Icon',
+                        value: ThumbnailSelections::DEFAULT_ICON.to_sym %>
+      </div>
+    </div>
+    <div class="col-sm-4">
+      <div class="row d-flex justify-content-center">
+        <% if thumbnail_form.resource.thumbnail_upload.present? %>
```
ajkiessl, line 13: There's a lot of logic in this file like this control flow.  I tried refactoring it into a view component, but it just made things more complicated and less testable imo.  Perhaps a presenter class would be better, or maybe it's fine as it is.

nicktiberi:
I just tried uploading a 336kb PNG that was larger than 200x200px, and got an nginx 413 request entity too large page. Is the upload limit for preview branches just really small compared to what it would be in production? I uploaded a 145kb PNG that was larger than 200x200 and it worked fine.

db/migrate/20220224191232_change_works_and_collections_auto_generate_thumbnail_to_thumbnail_selection.rb
```diff
@@ -0,0 +1,23 @@
+class ChangeWorksAndCollectionsAutoGenerateThumbnailToThumbnailSelection < ActiveRecord::Migration[6.1]
+  def up
+    add_column :works, :thumbnail_selection, :string, default: ThumbnailSelections::DEFAULT_ICON
```
rschenk: It may actually be better to use the string values here instead of the `ThumbnailSelections` constants. The reason being that, if in the future the `ThumbnailSelections` class is refactored or moved, this migration would no longer be executable. Even though it's way less DRY, it would be more future-proof to use `"default_icon"` instead of `ThumbnailSelection::DEFAULT_ICON`

app/forms/thumbnail_form.rb
```diff
@@ -10,16 +10,34 @@ def initialize(resource:, params:)
     super(params)
   end
 
-  def auto_generate_thumbnail
-    @auto_generate_thumbnail ||= resource.auto_generate_thumbnail
+  def thumbnail_upload
+    return nil if @thumbnail_upload.blank?
+
+    @thumbnail_upload
+  end
```
rschenk: `@thumbnail_upload.presence`

app/forms/thumbnail_form.rb
```diff
@@ -10,16 +10,34 @@ def initialize(resource:, params:)
...
-    resource.auto_generate_thumbnail = @auto_generate_thumbnail
+    resource.thumbnail_selection = thumbnail_selection
+    if thumbnail_upload.present?
+      resource.thumbnail_upload.destroy! if resource.thumbnail_upload.present?
+
+      tu = ThumbnailUpload.new resource: resource
+      tu.attributes = { file_resource_attributes: { file: thumbnail_upload } }
+      tu.file_resource.save
+      tu.save
+    end
```
rschenk: I think this should probably get wrapped in a transaction?

app/models/concerns/thumbnail_selections.rb
```diff
@@ -0,0 +1,25 @@
...
+              presence: true
+  end
+
+  def auto_generated_thumbnail?
+    thumbnail_selection == AUTO_GENERATED
+  end
+
+  def default_thumbnail?
+    thumbnail_selection == DEFAULT_ICON
+  end
```
rschenk, line 24: FYI you could do all of this functionality with an [ActiveRecord enum](https://medium.com/@moulayjam/ruby-on-rails-how-to-create-perfect-enum-easily-d8678e52488e). I personally only use them with a hash and string keys. You can use the `_prefix` or `_suffix` options to make those predicate methods like you have them.

I don't necessarily think you need to refactor this to use enums for it to be acceptable (I think it's fine as is) but it's good to know all the tools in your toolbox.

rschenk (dismissed):
I think this looks really good. I did offer a few ideas here and there, but none are show stoppers, so I'll let you decide which ones you'd like to change, or not.

app/views/resources/_thumbnail_display.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="row mb-3 no-gutters d-inline-block">
```
nicktiberi: Maybe this could be more generically renamed as `_heading_display.html.erb` or something like that, since it contains more than the thumbnail (title and description), and sometimes does not even contain the thumbnail.

ajkiessl:
@nicktiberi I'm not sure.  I just uploaded a 684KB image, though, and it worked.

nicktiberi (commented):
Overall this looks great! Aside from Ryan's comments, I just have one minor suggestion for renaming a partial, but not a dealbreaker.

ajkiessl:
Thanks @nicktiberi and @rschenk !  I'll work on the suggested changes.

nicktiberi (approved):
Great work, @ajkiessl! This is going to be an awesome feature. 🚀

PR #1253, opened 2022-02-23
fix for notify users checkbox

### LINKED ISSUE #1250: Editor notification emails sent even when not requested
# Bug Report

User report:

>  There is a box under the 'add editors' section of the "update setting" page in each work which one needs to check in order to send a notification to the added editor.  Although I'm not checking that box, I'm hearing from my colleagues that they are receiving an email notification every time I add them to a work as an editor.  Is that page working as it's should or do I misunderstand the function of that check box?

banukutlu:
Closes #1250

I was able to recreate this bug with a feature spec I just added. The checkbox values was "1" and "0" which was failing the check, changed it to be true, false.

### DIFF
--- a/app/forms/editors_form.rb
+++ b/app/forms/editors_form.rb
@@ -47,7 +47,7 @@ class EditorsForm
     resource.notify_editors = notify_editors
     resource.save!
 
-    send_emails(new_users: new_users) if notify_editors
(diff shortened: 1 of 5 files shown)


nicktiberi (approved):
🚀 Thanks, Banu!

PR #1256, opened 2022-02-24
Allow adding draft works to collections

### LINKED ISSUE #1231: Ability to add draft works to a collection
# Feature Request

Users should be able to add drafts work to collections through the collection "works" form. Currently, only published works can be added.

- [x] Draft works that a user has edit access to should appear in the "Add works to collection" search box.
- [x] Adding a draft work causes the work to appear in the list of the works on the "works" form. The work should be labeled as a draft.
- [x] Hint text on the page should read: "Draft works will not appear on the collection's public page until they are published."

Note this issue *does not* require the draft works to be displayed on the collection's public display page.

banukutlu:
Closes #1231

### DIFF
--- a/app/indexers/work_indexer.rb
+++ b/app/indexers/work_indexer.rb
@@ -13,6 +13,10 @@ class WorkIndexer
       IndexingService.add_document(version.to_solr, commit: false)
     end
 
+    work.collections.map do |collection|
+      IndexingService.add_document(collection.to_solr, commit: false)
+    end
+
     IndexingService.add_document(work.to_solr, commit: commit)
     Blacklight.default_index.connection.commit if commit
   end
--- a/app/models/concerns/catalog_search_behavior.rb
+++ b/app/models/concerns/catalog_search_behavior.rb
@@ -19,7 +19,7 @@ module CatalogSearchBehavior
 
   def exclude_empty_collections(solr_parameters)
     solr_parameters[:fq] ||= []
-    solr_parameters[:fq] << '-is_empty_bi:true'
+    solr_parameters[:fq] << '-is_empty_bsi:true'
   end
 
   def limit_to_public_resources(solr_parameters)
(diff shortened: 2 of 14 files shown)


app/views/dashboard/form/members/_search.html.erb
```diff
@@ -3,7 +3,6 @@
     <div class="form-wrapper form-wrapper--wide">
       <div class="keyline keyline--left mb-3">
         <h4><%= t('dashboard.form.members.search_heading') %></h4>
-      </div>
```
nicktiberi: Typo maybe? Looks like now there is a missing closing `</div>` tag.

nicktiberi (commented):
This looks great!

I did some local testing and noticed that if you have a collection with 1 draft work in it, and then publish the draft, the collection does not get reindexed, so the collection still won't appear in the main search results. I wonder if we need to just add an action to reindex any collections that the work is a part of as part of publishing a work version?

app/views/dashboard/form/members/_search.html.erb
banukutlu: yes typo, mistakenly pushed :) thanks for catching it.

banukutlu: this is fixed now.

banukutlu:
> This looks great!
> 
> I did some local testing and noticed that if you have a collection with 1 draft work in it, and then publish the draft, the collection does not get reindexed, so the collection still won't appear in the main search results. I wonder if we need to just add an action to reindex any collections that the work is a part of as part of publishing a work version?

that makes sense, I will look into it. thanks for testing this.

banukutlu:
@nicktiberi this is ready for review again :)

nicktiberi:
Awesome! I think this is good to go. Thanks, @banukutlu! 🙂  🚀

nicktiberi (approved):

PR #1262, opened 2022-02-28
Increase # of works requested from Solr as collection size grows

### LINKED ISSUE #1249: Problem adding works to a collection
# Bug Report
>  I'm in the process of uploading individual research posters into ScholarSphere and then adding each work into the Stuckeman School Research Open House collection. I've added 50 works into the collection, but I'm unable to add any additional works.  Are collections capped at 50 works?  

# Steps from user report:

1- Create and publish a work
2- Go to my dashboard and scroll until I find the entry for the collection
3- Click on Update Collection
4- Click on Works
5- Scroll to bottom of list of works to the 'Add works to the collection' box
6- Click on the down arrow in the box to reveal the title of the work that I just published
7- Scholarsphere shows the list of works I published that are not yet in a collection
-- although the work is published (I double checked), it does not show up in the box.  Instead, I receive this:  "No results found"

nicktiberi:
Closes #1249.

To prevent users from seeing 'No results found' when searching for works to add to a collection, we now request 50 + the number of works that are currently in the collection. This ensures that if there are any eligible works to be added to the collection, we will show them in the dropdown without the user needing to type anything into the search box.

### DIFF
--- a/app/controllers/dashboard/work_search_controller.rb
+++ b/app/controllers/dashboard/work_search_controller.rb
@@ -3,8 +3,9 @@
 class Dashboard::WorkSearchController < Dashboard::BaseController
   def index
     query = "#{params[:q]}*"
+    max_documents = params[:max_documents].present? ? params[:max_documents].to_i : 50
 
-    (member_works, _deprecated_document_list) = search_service(query).search_results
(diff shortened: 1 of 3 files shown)


banukutlu (approved):
lgtm 👍🏻

PR #1264, opened 2022-02-28
Remove 'collection' work type & set all works of this type to 'project'

### LINKED ISSUE #1234: It shouldn't be possible to create a work that is a collection
# Feature Request

Collections and works are mutually exclusive categories, however "Collection" is also a work type option. I think this is for historical reasons. We should remove the Collection work type and set existing works using this type to "Project".

nicktiberi:
Closes #1234.


ajkiessl (approved):
lgtm!

db/migrate/20220228193340_change_collection_works_to_projects.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeCollectionWorksToProjects < ActiveRecord::Migration[6.1]
+  def change
+    execute <<~SQL
+      UPDATE works SET work_type = 'project' WHERE work_type = 'collection'
+    SQL
```
ajkiessl, line 5: Just out of curiosity, is it better practice to use SQL in migrations that alter data?  I think I've done this before, but I've used the rails active record stuff instead.

nicktiberi: You're probably right that it's better to use ActiveRecord over raw SQL when possible. I'm happy to redo this if you'd prefer.

ajkiessl: I don't think one is necessarily better than the other, just was curious if there was some general preference when it comes to migrations.  The SQL way would probably be more performant.  I think it's fine the way it is.

ajkiessl: Actually, now that I think about it a bit more, the SQL would probably be preferable just in case you were to change the name of a model (or remove a model).  Say you were to change the `works` table's name and its associated model, and then run all the migrations.  The SQL would still work.  You would get an error if you used the ActiveRecord model in the migration.  You'd probably want to avoid changing the names of models, though.

ajkiessl: Deleting a table and model would probably be more common than changing their names, but it would still cause an error if using ActiveRecord.  Sorry, I'm just ranting now.  This all looks good.

PR #1265, opened 2022-03-01
1255 lifecycle timestamps

### LINKED ISSUE #1255: Improved work/version lifecycle timestamps
# Feature Request

With our existing timestamps, there's no way to sort submissions by when the current version was "published". (This refers to the [`published` state](https://github.com/psu-libraries/scholarsphere/blob/3f45a4b594c287230528a077a3b6aa1c077e8816/app/models/work_version.rb#L185) of a work version, not the `published_date` metadata field. It's confusing - I'm sorry). 

In addition to `created_at` and `updated_at`, we currently include the following "life cycle" timestamps in the data models for works and versions:
- Work's `deposited_at`: This timestamp is badly named. It is like `created_at` but in a separate field so that we can migrate content to new versions of ScholarSphere and not worry about preserving the old `created_at` values. For older deposits (from ScholarSphere 3.0), this is simply the datetime when the deposit was created/published. For works created with ScholarSphere 4.0, `deposited_at` should be the same as `created_at`.
- Work's `deposit_agreed_at`:  The intention of this timestamp is to refer to the time when a user *first* agreed to the deposit agreement indicated in `deposit_agreement_version`. If we update the deposit agreement and user publishes a new version of a work (they agree to the new version of the agreement) then this timestamp is updated. For reference, the code is [here](https://github.com/psu-libraries/scholarsphere/blob/b2a84e22ecdb8eff524b52fc2a5095b5282809f8/app/models/work_version.rb#L185).

No lifecycle timestamps are stored at the Work Version level which means we don't track when a version was published. To address this, we should:

- [x] Confirm that we can extract last state change datetimes for all work versions from Papertrail.
- [x] create `{state}_at` timestamps on the work version that are updated when the `aasm_state` field is set/updated. If a version is published, the `published_at` field is updated; if withdrawn, the `withdrawn_at` field is updated; etc.
- [x] Populate this field using information from Papertrail. (see step 1)

rschenk:
This closes #1255 and the description over there is pretty extensive, so check that out.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    aasm (5.1.0)
+    aasm (5.2.0)
       concurrent-ruby (~> 1.0)
     actioncable (6.1.4.4)
       actionpack (= 6.1.4.4)
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -178,7 +178,7 @@ class WorkVersion < ApplicationRecord
     raise ArgumentError, 'cannot delete published versions' if prevent_destroy
   end
 
-  aasm do
+  aasm timestamps: true do
     state :draft, intial: true
     state :published, :withdrawn, :removed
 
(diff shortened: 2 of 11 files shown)


nicktiberi (commented):
LGTM! I tested this locally and it works great. Thanks, Ryan! 🥇

nicktiberi (approved):

PR #1269, opened 2022-03-04
background jobs database locking

whereismyjetpack:
In qa/prod when you upload a thumbnailable? file, the signature job, and the thumbnail job kick off at the same time, and the thumbnail job thomps (this is the technical term) on the file resource. we run both jobs with db table locking so that the thumbnail job will be in queue until the signature job is complete.

### DIFF
--- a/app/jobs/shrine/promotion_job.rb
+++ b/app/jobs/shrine/promotion_job.rb
@@ -6,8 +6,8 @@ class Shrine::PromotionJob < ApplicationJob
   def perform(record:, name:, file_data:)
     attacher = Shrine::Attacher.retrieve(model: record, name: name.to_sym, file: file_data)
     attacher.atomic_promote
-    Shrine::ThumbnailJob.perform_later(file_resource_id: record.id) if record.thumbnailable?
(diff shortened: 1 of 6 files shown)


ajkiessl (approved):
This looks good.  I don't think this would cause any serious issues.  I'm assuming locking the record will prevent other transactions from happening (outside of jobs) on that record that perhaps in very rare cases may cause some latency (kinda just spitballing here cuz I'm not too sure).  I don't think that's really something to be too concerned about, but maybe just something to consider.

rschenk (commented):
I agree with @ajkiessl. After doing a little digging, it appears that the type of lock that ActiveRecord uses will prevent updates but not reads, so that seems good.

I guess the other alternative is to eliminate all the race conditions by combining all the file derivatives into a single, bigger job?

nicktiberi:
Thanks guys. This is only locking the `FileResource` record, which is rarely written to (except by these types of jobs), so I think it's fine as well.

Dann and I talked about possibly combining all the file derivative generation into a single, larger job but kind of preferred the flexibility of keeping them separate. This may be a good thing to revisit later.

PR #1270, opened 2022-03-07
Migrate view statistics with collection ids

### LINKED ISSUE #1267: update migration_collection_ids to transfer view #s
# Feature Request

The `migration_collection_ids` service should transfer view/download statistics from the collection to the work. See #1185, #1242

rschenk:
Originally, we were migrating the view statistics inside the `MergeCollection` service. However, now we have split that process into two different actions, one to merge a collection into a work, and another to actually flip the switch and remove the old collection. It makes sense to move the view statistics over during that latter process so we don't miss any view stats that occurred between the first step and the second.

I also went a little nuts and used ActiveRecord's `insert_all` feature to do a batch insert of all the newly migrated view statistics.

Closes #1267

### DIFF
--- a/app/services/merge_collection.rb
+++ b/app/services/merge_collection.rb
@@ -205,7 +205,6 @@ class MergeCollection
       version.metadata = canonical_work_version.metadata
       version.creators = canonical_work_version.creators.map(&:dup)
       version.file_resources = all_file_resources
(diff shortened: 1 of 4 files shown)


nicktiberi (approved):
Looks great! 🚀

PR #1279, opened 2022-03-09
Fixes the carriage return when appending the Publisher's Statement to description

ajkiessl:
I also added a test that checks the rendered html has a separate paragraph for the Publisher's Statement.


banukutlu (approved):
lgtm 👍🏻

PR #1283, opened 2022-03-14
Set thumbnail_selection to 'auto_generated' for Works and Collections that have an autogenerated thumbnail upon creation

### LINKED ISSUE #1281: Automatically set thumbnail_selection after Work or Collection creation
If a Work or Collection has an `auto_generated_thumbnail_url` on creation, then set the `thumbnail_selection` to "auto_generated".  Otherwise, the `thumbnail_selection` will default to "default_icon".

It might also make sense to create a script that converts all `thumbnail_selections` for existing Works and Collection to "auto_generated" if an `auto_generated_thumbnail_url` is found.  It should be noted, however, that this script could overwrite the `thumbnail_selection` for a record where the user actually preferred to use the default icon over the auto generated thumbnail.  Early on, this may be very few (or zero) so it might not be a big deal.

ajkiessl:
This was a little more complicated than I expected.  Since Works and Collections aren't created on a single form, I had to find a way to update the `thumbnail_selection`s at the right moment.  Also, generating thumbnail URLs in an asynchronous job added some more nuance to Works.  I also had to take into account that we don't want this happening every time someone edits a Collection or adds another version to a Work.  I used different patterns for Collections and Works.  I didn't want to stuff anything in the controllers, so I used the `after_save` hook in CollectionWorkMemberships to handle Collections.  However, I needed to put some code in the PublishController to handle Works properly.

I also realized that someone could delete a Work from a Collection or a Version from a Work containing their auto-generated thumbnail _after_ setting their `thumbnail_selection` to use the auto-generated thumbnail.  I added some logic in the `ThumbnailComponent` to display the generic icon in these instances.

closes #1281

### DIFF
--- a/app/components/thumbnail_component.rb
+++ b/app/components/thumbnail_component.rb
@@ -11,7 +11,7 @@ class ThumbnailComponent < ApplicationComponent
   end
 
   def display_thumbnail?
-    !resource.default_thumbnail?
+    !resource.default_thumbnail? && thumbnail_url.present?
   end
 
   def thumbnail_url
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -25,6 +25,8 @@ module Dashboard
         # save again--this time using the published validations. That way the
         # appropriate error messages will appear on the form when it's re-rendered
         if publish?
+          # WorkVersion#set_thumbnail_selection may be unreliable if the Shrine::ThumbnailJob is delayed
+          @resource.set_thumbnail_selection
           @resource.indexing_source = Proc.new { nil }
           @resource.save
           @resource.publish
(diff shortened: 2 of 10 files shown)


banukutlu (dismissed):
@ajkiessl this looks good to me 👍🏻

To your point regarding having to add logic to PublishController to handle Works properly, I also ended up doing same exact thing at the publish? method for to call minting DOI service on publishing a version ([here](https://github.com/psu-libraries/scholarsphere/pull/1285/files#diff-8b37d85cf8c961b3867615062d098402e6171ec88d0abe1be5440fb3787cf642R28)). I was wondering if there was another way to call that service. But what you have made sense to me for your case.

ajkiessl:
Adding @nicktiberi to reviewers.  Probably makes sense to wait to merge until he gets back.

ajkiessl:
I forgot to add a rake task to update the existing Works and Collections.  Just added this now.

nicktiberi (approved):
LGTM! Thanks @ajkiessl! 🥇

PR #1288, opened 2022-03-18
visibility fields for initial drafts

### LINKED ISSUE #1261: Work settings when creating first versions.
# Feature Request
On the "review and publish" page of the work submission form, you can set Penn State or Public access. This is a "work setting" that really only makes sense on the first version of a work (we disabled changing access from Public to Penn state once a version is published - see #1204). In addition, it would make sense to allow users to configure other "work settings" when depositing the first version of a work. After deposit, these settings can be managed through the "work settings" page.

- [x] The Penn State/Public access field should only appear for v1 draft works.
- [ ] Add embargo field (same UI as Work Settings Page) below the Access field in "Work Details". This  field should only appear for v1 draft works. Hint text: "Files will be restricted until after the embargo date. When the embargo lifts, files may be accessed based on the access settings above."
- [ ] Add checkbox: "Create DOI on Completion" that, when checked, causes a DOI to be be minted for the work *when the first version is published*.  This field should only appear for v1 draft works. Hint text: "Do not create a DOI if another publisher or repository has already assigned one. [Read more](https://docs.scholarsphere.psu.edu/guides/deposit-overview/#assigning-a-doi)"

banukutlu:
Part of #1261

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -300,6 +300,11 @@ class WorkVersion < ApplicationRecord
     end
   end
 
+  def initial_draft?
+    version_number == 1 &&
+      (draft? || temporarily_published_draft?)
+  end
+
   delegate :deposited_at,
            :depositor,
            :embargoed?,
@@ -332,5 +337,9 @@ class WorkVersion < ApplicationRecord
         TitleSchema,
         MemberFilesSchema
       )
+    end
+
+    def temporarily_published_draft?
+      aasm.from_state == :draft && aasm.to_state == :published
     end
 end
--- a/app/views/dashboard/form/publish/_visibility_field.html.erb
+++ b/app/views/dashboard/form/publish/_visibility_field.html.erb
@@ -1,6 +1,10 @@
-<div class="form-group">
-  <p class="label font-weight-bold mt-3"><%= WorkVersion.human_attribute_name(:visibility) %></p>
+<div class="keyline keyline--left">
+  <h2 id="<%= WorkVersion.human_attribute_name(:visibility).parameterize %>" class="h4">
+    <%= WorkVersion.human_attribute_name(:visibility) %>
+  </h2>
+</div>
 
+<div class="form-group">
   <% Permissions::Visibility.options_for_select_box.each do |lbl, val| %>
     <div class="form-check form-check-inline">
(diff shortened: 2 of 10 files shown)


app/models/work_version.rb
```diff
@@ -293,6 +293,15 @@ def all_dois
     super
   end
 
+  def thumbnail_url
+    auto_generate_thumbnail? ? file_resources&.map { |fr| fr&.thumbnail_url }&.last : nil
+  end
```
nicktiberi: Did this get added in by mistake? I'm not sure why this is part of the PR.

app/views/dashboard/form/publish/_visibility_field.html.erb
```diff
@@ -1,6 +1,10 @@
-<div class="form-group">
-  <p class="label font-weight-bold mt-3"><%= WorkVersion.human_attribute_name(:visibility) %></p>
+<div class="keyline keyline--left">
+  <h2 id="<%= WorkVersion.human_attribute_name(:visibility).parameterize %>" class="h4">
+    <%= WorkVersion.human_attribute_name(:visibility) %>
+  </h2>
+</div>
```
nicktiberi, line 6: Maybe I missed this while I was out; are we not adding the ability to mint at DOI in this form for this release? That's fine if so. Just want to make sure it was intentionally omitted.

nicktiberi (commented):
Overall this looks great! I tested it out and it works as expected. Just have 2 minor questions.

app/views/dashboard/form/publish/_visibility_field.html.erb
banukutlu: right, we are omitting the minting DOI part, here is the [thread](https://github.com/psu-libraries/scholarsphere/issues/1261#issuecomment-1070984401) about that decision.

app/models/work_version.rb
banukutlu: yes, probably missed it when rebasing, I will clean it up. thanks for catching it.

banukutlu: ok should be cleaned up now.

nicktiberi (approved):
LGTM - thanks, @banukutlu! 🎉

PR #1292, opened 2022-03-24
download, proccess, delete

whereismyjetpack:
Previously, io object was in memory -- for large files this was causing disk issues, this change downloads them to a tmp space, the we process on that file


nicktiberi (approved):
👍

PR #1308, opened 2022-05-11
cgi escape solr password

whereismyjetpack:
allows scholarsphere to use an password with special characters. 
rsolr update includes handling sending auth properly


banukutlu (approved):
lgtm 👍🏻

PR #1328, opened 2022-07-07
xit on flaky spec

whereismyjetpack:
this spec fails in CI often, and the MergeCollections controller isn't used at the moment.


nicktiberi (approved):
LGTM 😁

PR #1339, opened 2022-08-08
updates to ruby 3

whereismyjetpack:
- Updates to ruby 3.1.2
- Updates CI to use new circleci images. We now have to install browsers separately
- Introduces `.trivyignore` presently what is in there is webpack related. we'll have to find a pathway off webpack, it seems.

Right now, all rubocop related changes are in `.rubocop_todo.yml` and will be addressed in another PR. the prior PR with both was far too much to properly review. 

Addresses #1333


ajkiessl (approved):
Looks good!

PR #1343, opened 2022-08-25
Doi api endpoint to return Work uuids for WorkVersions with dois

ajkiessl:
DoiSchema removed from WorkVersion.  If a latest published work version has dois, they will be lumped with the associated work in the all dois solr query.


banukutlu (approved):
lgtm 👍🏻

PR #1366, opened 2022-12-12
preview/pid1

whereismyjetpack:
- make sure rails is pid1
- refactor how the app starts up
- don't use entrypoint


bin/startup
```diff
@@ -12,4 +16,4 @@ fi
 echo "starting rails"
 rm -f tmp/pids/server.pid
 
-bundle exec rails s -b '0.0.0.0'
+exec bundle exec rails s -b '0.0.0.0'
```
ajkiessl, line 19: `_exec_ bundle exec`?  Seems like a lot of "execs".  Doesn't this run in the container?

ajkiessl (approved):
Ok lgtm.  I read up on the linux `exec` command.  Makes sense.

whereismyjetpack:
@ajkiessl  -- an interesting read on why we'd use `exec` instead of a sub process https://about.gitlab.com/blog/2022/05/17/how-we-removed-all-502-errors-by-caring-about-pid-1-in-kubernetes/

PR #1392, opened 2023-03-21
Updates d3, and rails

whereismyjetpack:


ajkiessl (approved):
LGTM.  I think the record out on the preview branch doesn't have enough views to do this, but hovering the mouse over the view statistics doesn't let you see the number of views by date.  Given how this works in production, though, it looks like a certain threshold of number of views (or change in number of views) is needed to trigger this.  I'm curious to see how this works in QA.

Edit: Or the threshold might just be a number of days, and the record in the preview branch might not have been out there long enough.

whereismyjetpack:
If you'd like we can deploy this code to qa pretty easily and see what it does. I think there are some things that could be fixed in the code, like if you only have one days' worth of views the code does throw errors (verified in production , too)

ajkiessl:
Ok, I was thinking it would deploy to QA when merged to main and we could check it out then.

Issue #1399, opened 2023-04-18
Multiple "Actor" records for individual PSU members 


ajkiessl:
# Bug Report

## Description

There are many instances where a Penn State user has two Actor records.  One that has their Webaccess ID, but no ORCiD, and another with their ORCiD, but no Webaccess ID.  This seems to happen when the user does not have an actual "User" record in the system.  This causes issues when that user finally does decide to log in.  The log in will try to create a User record, then link an Actor record via Webaccess ID, and then populate the ORCiD.  It will throw an error here since there is already an Actor record with this ORCiD in the database.

We'll need to build a script to merge all the Actor records, their works, access IDs, and ORCiDs.  We'll need to identify all the places that duplicate Actors are being created and fix them.  One location this is happening is in the Work creation workflow.  When selecting contributors, if you search by access ID, it just populates access ID.  If you search by ORCiD, it just populates ORCiD.

## To Reproduce

Find a PSU user with two Actor records, and attempt to log in.

## Expected behavior
Users should only have a single Actor record.

## Solution

- [x] #1400 
- [ ] Identify and fix locations where multiple Actors are being created.  Take a look at the adding co-authors feature when creating a work.
- [ ] Investigate why some users have their psu email as their `psu_id` in their `Actor` record

ajkiessl:
Moving this to the backlog since we encountered this issue again and should be bumped up to high priority.

PR #1400, opened 2023-04-20
Rake task to deduplicate Actor records

ajkiessl:
This task operates under the assumption that these duplicates are only ever a set of 2: one with an ORCiD, and one with a PSU ID.

I've ran this several times locally and it fixes 25 duplicate groups.  I'd like to run this once on QA when the PR is merged before running it on production.

### DIFF
--- a/lib/tasks/actor_deduplication.rake
+++ b/lib/tasks/actor_deduplication.rake
@@ -1,8 +1,10 @@
-desc "Merge duplicate Actor records"
+# frozen_string_literal: true
+
+desc 'Merge duplicate Actor records'
 task actor_deduplication: :environment do |_task|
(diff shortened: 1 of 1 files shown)


whereismyjetpack:
Do we need to make a ticket for fixing the bug so this can't occur in the future?

ajkiessl:
I just threw everything in here: https://github.com/psu-libraries/scholarsphere/issues/1399 .  That's one of the tasks in that ticket.  I figured we could handle it in the next development cycle, since it seems to be a rare occurrence.

whereismyjetpack:
Great! I think that works for me.

whereismyjetpack (approved):

PR #1408, opened 2023-05-23
Security Vulnerabilities

ajkiessl:
@whereismyjetpack Tagging you as an FYI.  Imagemagick needed to be updated.  I tried uploading an image to kick off the thumbnail stuff in the preview branch but it didn't work.  I believe it's because minio isn't hooked up, though.  So I'll try this out again in QA when it gets merged.


whereismyjetpack:
This *should* work on previews, I'll give it a go later on

whereismyjetpack:
@ajkiessl new buckets are getting minted with different default ACL options, even though i think we are setting them in the aws operator. this will move to crossplane hopefully soon and i'll see if that fixes it. i modified the acls manually and verified that the background jobs are working -- this should work fine in qa and prod as those buckets already exist

whereismyjetpack (approved):

PR #1466, opened 2024-01-30
Split work type selection out into a separate step in deposit workflow

EricDurante:
A couple of decisions that I made here that weren't explicitly discussed:

cc @bdezray 

1. Instead of pulling just the work type selection out into its own step in the deposit workflow, I pulled out the whole "required metadata" section of the original form which includes the title as well. I did this because it was the cleaner/easier thing to do, and because I think that it makes sense - every type of work needs to have a title.
2. The link for editing an existing work took you to the form for editing all of the work details/metadata since that used to be the first step in the deposit workflow. Now the first step is title and work type editing, and all of the rest of the metadata is still in the original form (which is now the second step in the workflow). I left the edit link so that it still takes you to the form with the bulk of the metadata in the second step instead of updating it to take you to the new form in the first step. My reasoning is that if someone is editing a work, I assume that it's a lot less likely that they would be editing the title or the work type and much more likely that they'd be editing some other bit of metadata. If that assumption is wrong, or if that's just now how we want it to work, then I can change that.

### DIFF
--- a/app/components/form_tabs_component.rb
+++ b/app/components/form_tabs_component.rb
@@ -30,6 +30,13 @@ class FormTabsComponent < ApplicationComponent
     def work_version_tabs
       [
         OpenStruct.new(
+          label: I18n.t('dashboard.form.tabs.work_version_type'),
+          url: links_enabled && dashboard_form_work_version_type_path(resource),
+          controller: 'work_version_type',
+          active: false,
+          classes: []
+        ),
+        OpenStruct.new(
           label: I18n.t('dashboard.form.tabs.work_version_details'),
           url: links_enabled && dashboard_form_work_version_details_path(resource),
           controller: 'work_version_details',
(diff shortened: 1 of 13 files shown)


ajkiessl (approved):
Looks good and sounds good to me

EricDurante:
Thanks, @ajkiessl. What's the protocol for merging in ScholarSphere? Does anyone need to do QA/acceptance on this before it goes into main?

ajkiessl:
@EricDurante Nope, you can go ahead and merge.  ScholarSphere will work like RMD.  Just one approval is enough for you to merge.  I generally let the author of the PR do the merge in case they want to make any last minute changes.  Like RMD, it will deploy to QA: https://scholarsphere-qa.dsrd.libraries.psu.edu/ when merged to main.  It can be tested out by @bdezray there if needed.

PR #1468, opened 2024-01-30
Assign curator to a work

### LINKED ISSUE #1443: Administrator can assign a curator to a work

Smullz622:
Fixes #1443

### DIFF
--- a/app/controllers/dashboard/works_controller.rb
+++ b/app/controllers/dashboard/works_controller.rb
@@ -40,8 +40,8 @@ module Dashboard
 
         @embargo_form = EmbargoForm.new(work: @undecorated_work, params: embargo_params)
         @editors_form = EditorsForm.new(resource: @undecorated_work, user: current_user, params: editors_params)
-        @depositor_form = DepositorForm.new(resource: @undecorated_work, params: depositor_params)
-        @curator_form = CuratorForm.new(resource: @undecorated_work, params: curator_params)
+        @depositor_form = DepositorForm.new(resource: @undecorated_work, params: depositor_params) if current_user.admin?
+        @curator_form = CuratorForm.new(resource: @undecorated_work, params: curator_params) if current_user.admin?
         @withdraw_versions_form = WithdrawVersionsForm.new(work: @undecorated_work, params: withdraw_versions_params)
         @thumbnail_form = ThumbnailForm.new(resource: @undecorated_work, params: thumbnail_params)
       end
--- a/app/controllers/dashboard/works_controller.rb
+++ b/app/controllers/dashboard/works_controller.rb
@@ -41,6 +41,7 @@ module Dashboard
         @embargo_form = EmbargoForm.new(work: @undecorated_work, params: embargo_params)
         @editors_form = EditorsForm.new(resource: @undecorated_work, user: current_user, params: editors_params)
         @depositor_form = DepositorForm.new(resource: @undecorated_work, params: depositor_params)
+        @curator_form = CuratorForm.new(resource: @undecorated_work, params: curator_params)
         @withdraw_versions_form = WithdrawVersionsForm.new(work: @undecorated_work, params: withdraw_versions_params)
         @thumbnail_form = ThumbnailForm.new(resource: @undecorated_work, params: thumbnail_params)
       end
@@ -52,6 +53,8 @@ module Dashboard
           @editors_form
         elsif params[:depositor_form].present?
           @depositor_form
+        elsif params[:curator_form].present?
+          @curator_form
         elsif params[:withdraw_versions_form].present?
           @withdraw_versions_form
         elsif params[:thumbnail_form].present?
@@ -94,6 +97,14 @@ module Dashboard
           .fetch(:depositor_form, {})
           .permit(
             :psu_id
+          )
+      end
+
+      def curator_params
+        params
+          .fetch(:curator_form, {})
+          .permit(
+            :access_id
           )
       end
 
--- a/app/forms/curator_form.rb
+++ b/app/forms/curator_form.rb
@@ -1,42 +1,41 @@
 # frozen_string_literal: true
 
 class CuratorForm
-    include ActiveModel::Model
-  
-    attr_reader :resource
-  
-    def initialize(resource:, params:)
-      @resource = resource
-      super(params)
-    end
-  
-    def access_id
-      @access_id ||= current_curator_access_id
-    end
-  
-    def access_id=(access_id)
-      @access_id = access_id
-    end
-  
-    def save
-      add_curator
-      return false if errors.present?
-  
-      resource.save
-    end
-  
-    private
-  
-      def add_curator
-        user = User.find_by(access_id: @access_id)
-        Curatorship.create(user: user, work: resource)
-        unless user.present?
-          errors.add(:access_id, :not_found, access_id: @access_id)
-        end
-      end
+  include ActiveModel::Model
 
-      def current_curator_access_id
-        resource.curatorships.order(created_at: :desc)&.first&.access_id
-      end
+  attr_reader :resource
+
+  def initialize(resource:, params:)
+    @resource = resource
+    super(params)
   end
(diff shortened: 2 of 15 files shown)


app/views/dashboard/works/edit.html.erb
```diff
@@ -124,6 +124,28 @@
       </div>
 
       <% if current_user.admin? && policy(@work.latest_version).destroy? %>
```
EricDurante: Is this logic correct for deciding whether to hide or show the curator form/info?

I don't know, but I ask because it looks kind of odd. It looks like a condition for showing the curator bits is that the latest version of the work is able to be deleted (or something like that), and I'm not sure what those two things would have to do with each other.

app/controllers/dashboard/works_controller.rb
EricDurante, line 1: I would feel better if the controller enforced the rules for who is allowed to change what instead of us depending entirely on the UI (i.e. only hiding the curator form for non-admin users). Fortunately, it _looks_ like that's probably pretty easy to do here. Maybe something like:

Line 44:
```ruby
@curator_form = CuratorForm.new(resource: @undecorated_work, params: curator_params) if current_user.admin?
```
and/or

Line 56:
```ruby
elsif params[:curator_form].present? && current_user.admin?
  @curator_form
```
This change would probably result in the app throwing an error if a non-admin user submitted a request with only curator params, but that's probably also the case currently if any user submits a request with no params (so I wouldn't worry about it).

It's arguable whether this is really necessary in this particular case, but I'm of the mindset that it's better to be in the habit of coding defensively (or at the very least, thinking defensively). Since I can think of one (extremely unlikely but possible) scenario where the UI-only enforcement could be defeated without even hand-crafting an HTTP request, I'd say that it's worth doing the enforcement here where it can't possibly be circumvented.

app/forms/curator_form.rb
```diff
@@ -0,0 +1,41 @@
...
+      user = User.find_by(access_id: @access_id)
+      Curatorship.create(user: user, work: resource)
+      if user.blank?
+        errors.add(:access_id, :not_found, access_id: @access_id)
+      end
+    end
+
+    def current_curator_access_id
+      resource.curatorships.order(created_at: :desc)&.first&.access_id
+    end
```
EricDurante: I imagine that at some point we're going to need to add most of this same logic to the `Work` model (i.e. `Work#curator` to get the user who is the current curator for a given instance of a work), and that seems like a more natural place for it to live regardless. We don't need to implement it there now as long as we're mindful to not duplicate it if/when we do.

EricDurante (commented):
Looks good, but I had a few questions/comments/suggestions. Happy to discuss if needed.

app/views/dashboard/works/edit.html.erb
Smullz622: Initially I put curator in that block because that's where the depositor form is & they have the same access/permissions requirements.

Digging deeper, I don't think either curator or depositor should be there.  It looks like the full if condition [was first added](https://github.com/psu-libraries/scholarsphere/issues/867) to encompass the 'Danger Zone' so checking admin & able to delete made sense. Then it was requested that [the depositor form be added](https://github.com/psu-libraries/scholarsphere/issues/1031) to that zone & it was added within the same condition block, but not within the danger section.

So I think curator, depositor, and danger should be within `if current_user.admin?` but `if policy(@work.latest_version).destroy?` should just be around Lines 157-179 (danger section about deleting). Does that sound right?

app/controllers/dashboard/works_controller.rb
Smullz622: Makes sense to me! I assume I should also add that to depositor form while I'm at it since that also relies entirely on UI as of now?

app/views/dashboard/works/edit.html.erb
EricDurante: That certainly sounds right to me, but I'm new here 😄

app/controllers/dashboard/works_controller.rb
EricDurante: I would say so. Good catch!

app/views/dashboard/works/edit.html.erb
ajkiessl: The logic in the WorkVersionPolicy for destroy? might return false even for admins if the version is not the latest or the version is not persisted:
```
  def destroy?
    return false if record.new_record?
    return true if user.admin? && record == record.work.latest_version

    !record.published? && editable?
  end
```

Perhaps there was a reason to prevent editing the depositors in these cases?  I wouldn't change the logic for the depositors, but we might want the curator logic outside of that?

db/schema.rb
```diff
@@ -10,10 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_03_01_182323) do
+ActiveRecord::Schema.define(version: 2024_01_26_170854) do
 
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
ajkiessl, line 16: Not sure if having this extension removed will cause issues.  Probably best to keep it in.

db/schema.rb
```diff
@@ -212,7 +220,8 @@
   end
 
   create_table "versions", force: :cascade do |t|
-    t.string "item_type", null: false
+    t.string "item_type"
+    t.string "{:null=>false}"
```
ajkiessl: Not sure what's going on here.  

`t.string "item_type", null: false`

Seems normal and right to me.  Never seen these get split like that before.

ajkiessl (changes requested):
There's a couple things with the db schema I noticed.  The other things look good 👍

app/views/dashboard/works/edit.html.erb
ajkiessl: Although, I guess we're always passing the latest version in here, so that probably doesn't matter.  The persistence part might.

EricDurante: > Perhaps there was a reason to prevent editing the depositors in these cases?

Yeah, good question. Is there a test for this logic somewhere? If so, then I think we can assume that it was intentional. If not, 🤷‍♂️... who would know?

db/schema.rb
Smullz622: Good catch on both! Both of these db/schema changes keep happening to me automatically & accidentally were committed with the actual changes.

app/views/dashboard/works/edit.html.erb
EricDurante: The original [request](https://github.com/psu-libraries/scholarsphere/issues/1031) for that feature sounds like Seth was describing things in terms of how the UI looks and not necessarily in terms of how the business logic should work. It sounds like the important thing is that only admins can change depositors (which is also true for the "danger zone" actions). There might have been a miscommunication there. On the other hand, there doesn't seem to be a problem with the way that it has been working up to now.

If the way that it's working now is intentional and correct, then there's probably a better name than "destroy?" for that policy..

ajkiessl: I think we should just leave things as they are and wait to see if any issues pop up.  I agree, though, that the policy method's name (the way we are using it here) doesn't quite describe what's going on.  I don't see any tests that test for any of this logic other than admin vs non-admin.

app/models/work.rb
```diff
@@ -217,6 +217,10 @@ def latest_published_version_dois
     latest_published_version.all_dois
   end
 
+  def current_curator_access_id
+    curatorships.order(created_at: :desc)&.first&.access_id
+  end
+
   private
```
EricDurante, line 224: @Smullz622 Could you please add a test for this method since it's public?

EricDurante (changes requested):

ajkiessl (approved):
lgtm

app/models/work.rb
Smullz622: Ahh missed that one. Added!

EricDurante (approved):

PR #1480, opened 2024-02-07
1467 scholarly works pathway

### LINKED ISSUE #1467: Customized forms for the Scholarly Works Pathway
Fields for this pathway are listed below. In our current set up the creators are added in their own form and the embargo date occurs during the work settings. It may be useful to consider moving the embargo setting somewhere in the deposit form. Additionally, we would like to prevent ScholarSphere DOI minting for records that have a publisher identifier (DOI).
Work Title (Required)
Work Type (Required)
Description (Required)
Publication Date (Required)
Subtitle (Optional)
Publisher's Statement (Optional)
Keyword(s) (Optional) -> Does RMD pull keywords?
Publisher (Optional)
Publisher Identifier (DOI) (Optional)
Related URLs (Optional)
Subject (Optional)
Language (Optional)
Creator (Required)
Depositor (Required)
Embargo Date (Optional)
License (Required)

EricDurante:
Resolves #1467 

@rschenk This isn't finished yet, but I wouldn't mind having your eyeballs on what I'm doing here. For some context, [this part](https://github.com/psu-libraries/scholarsphere/pull/1466) is already done.

One thing to note is that we decided to add some of the validations that formerly only applied to published works to the earlier steps in the deposit process. For example, description and published date are now required to save the "work details"  form. This duplicates the conditional validations in the `WorkVersion` model. I considered removing validations from the model and allowing the forms to handle them entirely, but then I realized that the validations may still need to be implemented at the model level also since there are other ways (apart from the web forms) to create work versions that may also need to be validated (like via the API?). Does that sound right to you?

Otherwise, how does this approach feel to you?


rschenk:
> there are other ways (apart from the web forms) to create work versions that may also need to be validated (like via the API?). Does that sound right to you?

Yep, I think this is spot on. I don't know how the API works (that was something developed by Adam and Seth without me) but I know it was used to port all content from SS3 to SS4 so clearly it _is_ creating work-versions outside the context of the deposit form

app/models/work.rb
```diff
@@ -212,6 +226,16 @@ def latest_published_version_dois
...
+  def deposit_pathway
+    if Types.scholarly_works.include?(work_type)
+      :scholarly_works
+    elsif Types.data_and_code.include?(work_type)
+      :data_and_code
+    elsif Types.general.include?(work_type)
+      :general
+    end
+  end
+
```
rschenk: I really like where this is going, the only thing I see is that perhaps the `Work` model should not have any knowledge of the deposit pathway. Perhaps this method should live elsewhere?

EricDurante: Good call. I didn't like this either.

I think I know how I want to refactor it.

EricDurante: https://github.com/psu-libraries/scholarsphere/pull/1480/commits/8fbf1612ade5cc77480517020ddff0762ab9f95a

EricDurante:
This is ready for a final review, I think.

app/models/work_version.rb
```diff
@@ -310,6 +314,8 @@ def initial_draft?
            :auto_generated_thumbnail?,
            :thumbnail_url, to: :work
 
+  delegate :deposit_pathway, to: :work, prefix: false
```
ajkiessl: Should this be removed?  If I'm understanding correctly, this logic is now held in the WorkDepositPathway class.

spec/models/work_version_spec.rb
```diff
@@ -246,6 +246,7 @@
   it { is_expected.to delegate_method(:proxy_depositor).to(:work) }
   it { is_expected.to delegate_method(:work_type).to(:work) }
   it { is_expected.to delegate_method(:thumbnail_url).to(:work) }
+  it { is_expected.to delegate_method(:deposit_pathway).to(:work) }
```
ajkiessl: Same here.  Can this be removed?

ajkiessl (changes requested):
Looks great so far, Eric.  I haven't gotten through all the tests yet, but I wanted to drop these change requests in the meantime.

app/models/work_version.rb
EricDurante: Yep - I missed this. Thanks!

app/forms/work_deposit_pathway.rb
```diff
@@ -0,0 +1,151 @@
...
+          ).freeze
+        end
+
+        form_fields.each { |attr_name| attribute attr_name }
+
+        validates :version_name,
+                  allow_blank: true,
+                  format: {
+                    with: /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/, # rubocop:disable Layout/LineLength
+                    message: I18n.t('activerecord.errors.work_version.attributes.version_name.format'),
```
ajkiessl: The path to this translation is missing `.models`: `I18n.t('activerecord.errors.models.work_version.attributes.version_name.format')`

Also, I'm curious why this validation is here.  It's basically the same as the model level validation except for `allow_blank: true`.

app/models/work_version.rb
```diff
@@ -123,7 +123,7 @@ def label(id)
             allow_nil: true,
             format: {
               with: /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/, # rubocop:disable Layout/LineLength
-              message: 'Version names must be in semantic version format, ex. 1.0.0',
+              message: I18n.t('format'),
```
ajkiessl: This translation didn't work either.  `I18n.t('activerecord.errors.models.work_version.attributes.version_name.format')` worked for me.

ajkiessl (changes requested):
Pretty much everything else looks good.  I just noticed two things below.

app/forms/work_deposit_pathway.rb
EricDurante: @ajkiessl Thanks for not just reviewing the code but also doing thorough QA - very much appreciated!

You're right. I just needed the form to reflect the underlying model validation here since the rules are identical, but there are better ways to do that than what I did (duplicating the same validation logic).

app/models/work_version.rb
EricDurante: Yep - I did this wrong. Good catch!

ajkiessl (approved):
LGTM thanks!

Issue #1458, opened 2024-01-21
Replace Go script


bdezray:
Replace the Go Script which uses a GitHub Action to add new tasks to an Airtable database each day accounting for the issues described in #1306

bdezray:
This solution should also enable curators to remove closed records from AirTable when curation is complete. The current system will import records which are not currently present in the AirTable. So if we remove a record when curation is completed, the next time the GitHub Action runs that record will be reimported into AirTable as if it was never curated before.

PR #1482, opened 2024-02-09
add request for dataset curation

### LINKED ISSUE #1463: Dataset Pathway: User initiated data curation request prior to publication
# Feature Request
Provide users with an option to request curation prior to publication while still in draft state. This could be displayed on the final review page as an option. Selection of this option would need to trigger an entry in AirTable.

Smullz622:
Fixes #1463 

Using Airrecord gem here to handle the AirTable client request. Hopefully this will also set the foundation for #1458 

Request curation button will appear under the following conditions:
- on the 'Review & Publish' tab of the form
- work is a "dataset" or "software or program code" type
- curation has not already been requested

When curation has already been requested, request curation and publish buttons are both disabled.


app/controllers/dashboard/form/base_controller.rb
```diff
@@ -73,11 +78,40 @@ def cancel_path
...
+          current_controller == 'dashboard/form/publish' && current_action == 'edit'
+        end
+
+        def data?
+          @resource.work.work_type == 'dataset' || @resource.work.work_type == 'software_or_program_code'
+        end
+
+        def curation_eligible?
+          !@resource.is_a?(Collection) && !@resource.draft_curation_requested && data?
+        end
```
Smullz622, line 110: FYI `data?` is just a placeholder until the work for `deposit_pathway` is merged

app/controllers/dashboard/form/base_controller.rb
```diff
@@ -16,6 +16,7 @@ def resource_klass
           end
         end
 
+        # where drafts are being saved
```
EricDurante: Guessing that you probably didn't mean to commit this comment?

app/controllers/dashboard/form/base_controller.rb
```diff
@@ -73,11 +78,40 @@ def cancel_path
...
           resource_klass.model_name.param_key
         end
 
+        def in_publish_edit_action?
+          current_controller = params[:controller]
+          current_action = params[:action]
+
+          current_controller == 'dashboard/form/publish' && current_action == 'edit'
+        end
+
```
EricDurante: I have some ideas for what we can do with the two methods below once [1467-scholarly-works-pathway](https://github.com/psu-libraries/scholarsphere/tree/1467-scholarly-works-pathway) is merged into this work.

db/schema.rb
EricDurante, line 1: It looks like there are some schema changes here that weren't the product of a migration in this branch. We should probably remove those and make sure that they get committed alongside their corresponding migrations in the appropriate branch instead.

It's definitely a pain to keep all of this straight when working on two (or more) different features at the same time in two different branches with two different sets of database migrations. Unfortunately, the cleanest way to do it is probably to remember to roll back the new migrations to your dev & test databases before you switch branches to work on other database changes.

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -24,12 +24,19 @@ def update
...
           @resource.set_thumbnail_selection
           @resource.indexing_source = Proc.new { nil }
           @resource.save
           @resource.publish
+        elsif request_curation?
+          begin
+            AirtableExporter.call(@resource.id)
+            @resource.draft_curation_requested = true
+          rescue Airrecord::Error
+            flash[:error] = t('dashboard.form.publish.curation.error')
```
EricDurante, line 55: What happens if Airrecord errors out but then there's also a validation error when attempting to save the work? Will the user see the both the validation error message and the message letting them know that the curation request failed? Either way, I wonder how much it matters in terms of UX. It might be worth it to actually try this scenario and see what happens and how it feels as a user.

EricDurante (changes requested):
I have a handful of thoughts about things that I'd like to see here. Some of them could maybe come later, but they're all worth thinking about now.

First, I've gathered from various team conversations that AirTables are expensive and that there unfortunately isn't a practical way for us to provision real AirTable sandboxes for our test, development, QA, and staging environments - is that right? If so, that's too bad since it means that we can't have an automated, end-to-end test that exercises a real, live AirTable integration. However, if we can't have the real thing, then I think that we should have one test that gets as close as we possibly can to a true end-to-end exercise. All that really means is that I'd like to see a feature test where we click on the button to request curation and then expect a mocked-out `Submission` class to receive the correct call with the correct data. I know that we have a test for the `AirtableExporter` class that does this (which is good), but I'd also like to see one test that verifies that all of the right stuff happens when we actually click the button in the UI. An ideal test would read back and verify data that was actually saved to an AirTable sandbox. If we can't have that, then the next best thing is to verify that our interface to AirTable (`Submission`) receives the correct call when the button is clicked. This one I think we should definitely do now.


Second, it's awesome that you've started to break ground here on what will likely become a full and direct integration with AirTable. Knowing that this is somewhere on the horizon, and knowing that maybe even further down the line we'd replace AirTable with something else altogether (like a similar SaaS option, or even functionality that's built directly into ScholarSphere), I'd like to start out with a design that's as flexible as possible to facilitate those future changes. The most obvious improvement that I see that we could make right now is in the AirTable error handling. Ideally, for our design to remain flexible, there should be exactly one part of ScholarSphere that knows about AirTable, and rest of ScholarSphere should be entirely oblivious to the existence of AirTable specifically (and when I say "AirTable" here, in practice I mean the 3rd-party gem that serves as our interface to AirTable).

Right now, we already have two parts of ScholarSphere that know about AirTable/Airrecord - the `AirtableExporter` class knows how to create new AirTable records with Airrecord, and the `PublishController` class knows how to handle Airrecord errors. It would be better if `AirtableExporter` knew how to handle Airrecord errors instead and hid that knowledge from `PublishController`. Right now, in concrete terms, that might only mean having `AirtableExporter` rescue `Airrecord::Error` and raise its own custom error that would then be handled by `PublishController` in turn. That's so trivial that it might sound silly, but it makes a huge and important difference in the design that I really can't overstate. To see why, it's helpful to think of `AirtableExporter` as something more abstract than its current name suggests.

If we think about how to describe what the ScholarSphere app is really trying to accomplish here, it's tempting to speak in terms of the current implementation details and say, "well, ScholarSphere needs to send data to AirTable" - hence "Airtable Exporter". However, that's describing the current means rather than the end. What ScholarSphere _really_ wants to do is to add a new task to a list of curation tasks that need to be completed. So let's suppose for a minute that we change the name of `AirtableExporter` to `CurationTaskExporter`. Again, it's a trivial-seeming name change, but, at least for me, it immediately starts to change how I think about the whole design of the integration.

Suppose that in a few months a new SaaS startup launches a product called WebSheetz (no connection to the gasoline and breakfast sandwich company... or is there?). For our purposes, WebSheetz does pretty much the same thing as AirTable, but it costs half as much (and they have better coffee), so we decide that we want to switch. WebSheetz of course has a REST API, and the developers have kindly provided a ruby wrapper for it, but its API is naturally different than Airrecord's. Suddenly, if the details of how to add curation tasks to a list by interacting with AirTable/Airrecord are woven throughout the ScholarSphere codebase, then we have a big problem. Switching from AirTable to WebSheetz is going to require a massive change to a lot of code. On the other hand, if we've  completely hidden all of those details behind one abstraction, and the rest of ScholarSphere only interacts with that abstraction, then the job becomes much easier. If we succeed in completely abstracting the storage of curation tasks so that the rest of ScholarSphere knows nothing about _how_ curation tasks are stored, then `CurationTaskExporter` is the only code that we'll need to change in order to switch from AirTable to WebSheetz. That includes handling problems that arise from `CurationTaskExporter`'s internal workings. `Airrecord::Error` is part of Airrecord's API, so the rest of ScholarSphere shouldn't be exposed to it if we want to minimize this coupling as much as possible.

If all other parts of ScholarSphere that end up dealing with curation tasks are able to say to `CurationTaskExporter`, "Here's some task data - do whatever it is that you do to store it", and if `CurationTaskExporter` raises `CurationTaskExporter::Error` (for example) when something goes wrong with its internal workings that it can't handle, then we're free to completely change how `CurationTaskExporter` does its thing without fear of breaking and needing to change any other part of the app (like `PublishController`, for example).

To take this further, suppose that we get to the point where we no longer want to pay for a SaaS solution (or maybe none of the available SaaS solutions do everything that we want), and we decide to bring all of this functionality into ScholarSphere itself. Maybe there's a `curation_tasks` table in the ScholarSphere database now and a `CurationTask` ActiveRecord. We could still leave the rest of ScholarSphere using the `CurationTaskExporter` interface to store tasks (although now the name isn't quite as accurate), and again, we wouldn't need to change anything except for the implementation of `CurationTaskExporter`. If we're looking this far ahead, then we might be able to lay the groundwork for an even cleaner transition in the future by giving ScholarSphere an interface for storing curation tasks that looks more like an ActiveRecord. For example, instead of `CurationTaskExporter.call(work_version_id)`, maybe we have `CurationTask.create_for_work_version(work_version_id)`. Behind that interface we could still have code that sends data to AirTable or to WebSheetz, but now that class could cleanly be made to inherit from ActiveRecord and store data in the ScholarSphere database also. Awesome!

We stand to possibly benefit from a design that's this flexible even if we never move off of AirTable at all. The other day at stand-up, I think I remember you expressing some very justified trepidation about introducing a dependency on Airrecord since the gem didn't seem to be undergoing a lot of active maintenance. That kind of thing is _always_ a danger, and this kind of design is what shields us from that danger as much as is possible. Our abstraction for storing curation tasks uses the Airrecord gem right now, but just like switching to a totally different SaaS product or even to ActiveRecord, the abstraction should allow us to painlessly switch to a different third-party AirTable gem (if one comes along) or to our own home-grown AirTable API client if necessary. Similarly, the abstraction insulates the rest of ScholarSphere from breaking API changes when Airrecord goes from v1 to v2 and beyond.

Sorry to have gone on so long about this one, but I did so because it's really important. This way of thinking about the problem is part of what makes the difference between software that survives to the end of its usefulness and software that slowly dies from becoming unmaintainable. I think that we should make some of these little tweaks now so that we're starting down the best possible path that we can given what we know right now.


Finally (and this one is much less important than the first two), it would be nice if this feature "worked" in development. That is to say, since we don't have a development environment for AirTable, it would be nice if the app at least did _something_ other than error out for lack of an AirTable API key when we click the "Request Curation & Save" button in development. That "something" could just be pretending that we sent the data somewhere and that it worked. Even better than that would be to maybe write the data to a log or to a tempfile for inspection. I have an idea for how we could do this that would be pretty easy if we implemented the kind of clean interface for storing curation tasks that I described above. I'd consider this to be really low priority, though, and definitely not something that we need to do right now. Just an idea for later if there's time.


I'd be happy to meet if you'd like to discuss any or all of this.

whereismyjetpack:
Apologies for not being as involved in Scholarsphere development as usual. Eric's points are insightful, and I recall our previous discussion about potentially removing Airtable altogether and integrating the curation workflow directly into Scholarsphere. While I appreciate the effort to integrate Airtable to address some of the issues with unmaintained scripts, I don't believe (at least from our last discussion) that it represents our desired final state.

Additionally, it's important to note that LST's involvement with the Airtable setup is minimal; it remains a black box to us, which could be perceived as a potential risk.

app/controllers/dashboard/form/publish_controller.rb
Smullz622: Just tested this out, and it works the same way as the save & exit button so users wouldn't see the validation error & Airrecord error at the same time. Validation error messages are contained entirely within the form, and users can save & exit/request curation even if they have an invalid entry. I assume this behavior is to allow data to persist so users can come back later & see where they went wrong. The Airrecord validation error displays after the redirect back to the work page. So it is possible a user could get a message that their work saved successfully as well as a message that the curation request failed. In this case, they are prompted to request curation again or contact us if the problem persists. I was debating how to handle this scenario & am open to suggestions.

ajkiessl: If I input a junk version_name and submit it shows both the Airtable error and validation error.  Would it make sense to only proceed with the Airtable export if the record is valid + saves?

ajkiessl:
Outside of my comment above and what Eric said, I don't have anything else to add.  I like how this is shaping up.  Looking good.

app/controllers/dashboard/form/publish_controller.rb
EricDurante: Good question, @ajkiessl. I'm not sure what the right answer is. Thinking through it - the work/version records would have already been created at this point with at least some of the metadata entered. A validation error at this step would indicate that something is missing or incorrect that would prevent the work from being published. My assumption is that we'd want the depositor to give their best shot at submitting complete/publishable files and metadata _before_ the work is sent off for curation. With that being the case, I think we'd want to do what you suggested and only proceed with the Airtable export if the record is valid and saves successfully. However, it would need to be clear to the depositor that their curation request didn't succeed due to the validation errors and that they would need to retry after fixing the errors.

EricDurante:
Here's one thing that I hadn't thought about until now -

In discussing this feature, I think that we decided that it'll be up to the curators to ultimately publish works for which the depositor has requested curation. What I'm unsure of is whether or not there will actually be a way for the curators to do that without some additional change.

Is there already some means by which a curator/admin could publish a work that has `draft_curation_requested` set? Or do they have a way to unset `draft_curation_requested`? Does what I'm asking make sense?

EricDurante:
I guess that if the curation activity results in a new WorkVersion, then the curator will be able to publish that new version, but if it doesn't, then I'm not sure that there's currently a way for them to get the current version published.

Smullz622:
@EricDurante oh yeah, good catch! That will definitely be a problem. I can add some logic so the publish button always renders for admins so we can still prevent users from publishing after the request, but admins will be able to publish after curation.

config/locales/en.yml
```diff
@@ -399,6 +399,7 @@ en:
           collection: Save and Exit
           work_version: Save as Draft & Exit
           admin_save: Save and Exit
+        request_curation: Request Curation & Save
```
EricDurante: This is of pretty low importance (and let me know if you disagree), but I'd like to switch the words around on this button so that it's "Save & Request Curation" or even "Save as Draft & Request Curation" (more verbose, but also more descriptive). That wording better describes the order in which things actually happen, it may help to make it obvious to the user that the "request curation" part doesn't happen if there's a problem with the saving part (like a validation error), and it's consistent with the wording on the other buttons (i.e. "Save as Draft & Exit").

app/controllers/dashboard/form/base_controller.rb
```diff
@@ -73,11 +77,38 @@ def cancel_path
...
+          current_controller = params[:controller]
+          current_action = params[:action]
+
+          current_controller == 'dashboard/form/publish' && ['edit', 'update'].include?(current_action)
+        end
+
+        def curation_eligible?
+          @resource.is_a?(WorkVersion) &&
+            !@resource.draft_curation_requested &&
+            Work::Types.data_and_code.include?(@resource.work.work_type)
```
EricDurante, line 109: Just wanted to note that I have this same code in my current feature branch. After it's all merged together, I'd like to refactor this, and I think I know roughly how I'd like to do it. Essentially I think I'd like the `WorkDepositPathway` object to be responsible for this logic about whether or not curation can be requested rather than having it here in the controller. That'll require some other changes though - like making an `WorkDepositPathway` instance available everywhere within the controllers that are responsible for the deposit workflow (which makes sense, I think).

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -26,12 +26,33 @@ def update
...
+        elsif request_curation?
+          # We want validation errors to block curation requests and keep users on the edit page
+          # so WorkVersion needs to temporarily act like it's being published. It's returned to it's
+          # initial state before being saved.
+          begin
+            @resource.save
+            initial_state = @resource.aasm_state
+            @resource.publish
+            if @resource.valid?
+              CurationTaskExporter.call(@resource.id)
```
EricDurante, line 45: We would want this action to be idempotent with regard to creating Airtable records. So one more thing that I would try out is to make sure that you can't create duplicate records in Airtable by double-clicking on the "Request Curation & Save" button (to fire off a second request before the first one returns). I suspect that you might be able to. If so, I think there are a couple of things that we could do to mitigate that.

EricDurante (commented):
I think your changes look good. One more thing to check (sorry that I didn't catch it the first time), and one really minor suggestion that's just my opinion (so feel free to ignore).

app/controllers/dashboard/form/publish_controller.rb
Smullz622: I believe Airtable actually does handle this. It doesn't allow duplicate IDs (in this case the scholarsphere uuid).

config/locales/en.yml
Smullz622: Save as Draft is definitely more descriptive & I think it would be a good change. I think it might be good to leave Request Curation first though. It's super minor, but I think having it first will help catch users attention more & make sure someone doesn't click it without reading the whole thing thinking they're just saving it.

app/controllers/dashboard/form/publish_controller.rb
EricDurante: Will attempting to create a duplicate ID raise an error in Airrecord? If so, then we may want to bubble that error up, rescue it, and show a message to the user so that a double-click on the button doesn't just result in a 500 error page.

config/locales/en.yml
EricDurante: Good point - sounds good to me!

app/controllers/dashboard/form/publish_controller.rb
Smullz622: Nevermind, I take it back. When I tried double clicking, it only sent one record so I thought it confirmed that Airtable doesn't allow duplicates. However, I just tested sending a record twice in the console and both went through.

EricDurante: Ok, in that case we should probably try to do something to ensure idempotence here. One thing that would probably work would be to do
```ruby
@resource.update_column(:draft_curation_requested, true)
```
right after
```ruby
elsif request_curation?
```
and then first thing in both the `rescue` block and in the `else` condition:
```ruby
@resource.update_column(:draft_curation_requested, false)
```

EricDurante: Oh, for that ☝️ to work, we'd also need to make the `elsif` condition
```ruby
elsif request_curation? && !@resouce.draft_curation_requested
```
which we should do anyway.

EricDurante: I fixed a mistake in my last comment ☝️

EricDurante (approved):

PR #1483, opened 2024-02-12
set cap on embargo length

### LINKED ISSUE #1474: Cap embargo length to 48 months

Smullz622:
Fixes #1474

### DIFF
--- a/app/forms/embargo_form.rb
+++ b/app/forms/embargo_form.rb
@@ -50,6 +50,10 @@ class EmbargoForm
 
       begin
         Date.parse(embargoed_until.to_s)
+        unless embargoed_until < (DateTime.now + 4.years)
+          errors.add(:embargoed_until, :max)
+          nil
+        end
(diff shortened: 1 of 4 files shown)


EricDurante (dismissed):
Looks good!

ajkiessl (commented):
Looks good.  The only thing I noticed is that the "Remove Embargo" button shows up after validations fail.  This doesn't seem to break anything but it might be a little confusing.

Smullz622:
> Looks good. The only thing I noticed is that the "Remove Embargo" button shows up after validations fail. This doesn't seem to break anything but it might be a little confusing.

I think it'll make sense overall since it's still functioning correctly. When the "Remove Embargo" shows up, the invalid date is still in the field. If you click "Remove Embargo", it clears that value, or if you navigate away then come back, the invalid value & "Remove Embargo"button are both gone.

EricDurante (approved):

PR #1484, opened 2024-02-13
Only render matomo code in production environment

ajkiessl:
Matomo was blowing up our test suite in CI with a `Net::ReadTimeout` error.  The matomo code does not need rendered in any environment outside of production, so I wrapped the code in some logic to reflect that.

### DIFF
--- a/spec/features/matomo_spec.rb
+++ b/spec/features/matomo_spec.rb
@@ -9,8 +9,8 @@ RSpec.describe 'Matomo Traking Code' do
   end
 
   after do
+    ENV.delete("MATOMO_TEST")
     Capybara.ignore_hidden_elements = true
-    ENV.reject! { |k, _v| k == 'MATOMO_TEST' }
   end
 
   it 'has the default configuration' do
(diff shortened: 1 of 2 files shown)


whereismyjetpack (dismissed):

spec/features/matomo_spec.rb
```diff
@@ -3,9 +3,15 @@
...
-  before { Capybara.ignore_hidden_elements = false }
+  before do
+    ENV['MATOMO_TEST'] = 'true'
+    Capybara.ignore_hidden_elements = false
+  end
 
-  after { Capybara.ignore_hidden_elements = true }
+  after do
+    Capybara.ignore_hidden_elements = true
+    ENV.reject! { |k, _v| k == 'MATOMO_TEST' }
```
whereismyjetpack: Since this is a test, I don't think it matters -- but I wonder if something like `ENV.delete('MATOMO_TEST')` would be more sturdy

ajkiessl: Ah yeah that's cleaner.  I'll change it.

Smullz622 (approved):
lgtm!

Issue #1465, opened 2024-01-25
Depositor initiated contact to curators


ajkiessl:
Currently, depositors can initiate contact generally through the [Contact Form](https://scholarsphere.psu.edu/incidents/new). However, this does not allow a depositor to seamlessly communicate with a curator about a particular deposit and have that information linked to that deposit. This potentially could be accomplished by including an Ask widget that links to the ScholarSphere queue. Need to explore how that info gets stored.

PR #1490, opened 2024-02-19
Create Curator to Depositor Ticket

### LINKED ISSUE #1464: Curator initiated contact to depositor
Ideally, this process would enable autogenerated information from ScholarSphere to be entered into the "Create a Ticket Initiated by You (System)" in LibAnswers. Some information can be pulled directly from the ScholarSphere deposit into the "Create a Ticket Initiated by you (System)" form, while others need to be supplemented with other information:
| LibAnswers Field | ScholarSphere Field |
|------------------|---------------------|
| Send to (name) | Depositor Name (directly from ScholarSphere) |
| Send to (email) | Depositor Email (directly from ScholarSphere) |
| Subject | Insert "Work Title" into fillable email subject line (Supplemented with info directly from ScholarSphere) |
| CC Email to | Creator and Editor Emails (directly from ScholarSphere) |
| Email Message | Message from curator which needs to be generated somewhere either in ScholarSphere or LibAnswers |

ajkiessl:
fixes #1464 

This doesn't do exactly what we wanted because the LibAnswers API does not create tickets the way the "Create a Ticket Initiated by You (System)" form does in LibAnswers.  The "Create a Ticket Initiated by You (System)" form creates a ticket and sends an email to the patron at the same time without any prior contact with the patron.  The API creates a ticket as if the patron was sending a message to the admins.  The workaround for this is to use the API to create the ticket with just the patron's name and email and a canned subject.  Then the admin can add a message in the LibAnswers UI and anything else they would like before submitting and sending an email to the patron.

There is a button in the work settings page that initiates this:
Screenshot 2024-02-19 at 2 54 31 PM
With a confirmation dialog:
Screenshot 2024-02-19 at 2 55 08 PM
Confirming will take the admin to the ticket in LibAnswers where they can complete the rest of the information.  When an admin does this and submits the ticket in LibAnswers an email will then go out to the patron.  The email will look a bit awkward:
Screenshot 2024-02-19 at 2 58 08 PM
Where the Original Question is the subject from when the ticket was created via the API.  The response above is the message the admin completes in LibAnswers.


app/controllers/admin/create_curation_ticket_controller.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Admin
+  class CreateCurationTicketController < ApplicationController
```
EricDurante: Just wondering about the naming here. Can we foresee any other actions that ScholarSphere might perform on LibAnswers tickets? Would it make more sense for this to be `CurationTicketsController` with a `#create` action?

app/services/libanswers_api_service.rb
```diff
@@ -0,0 +1,67 @@
...
+    def send_to_name
+      work.depositor.display_name
+    end
+
+    def send_to_email
+      work.depositor.email
+    end
+
+    def subject
+      "ScholarSphere Deposit Curation: #{work.versions.last.title}"
```
EricDurante: Refactoring suggestion:  replace `work.versions.last` with `work.latest_version`.

app/services/libanswers_api_service.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  private
+
+    def send_to_name
+      work.depositor.display_name
+    end
+
+    def send_to_email
+      work.depositor.email
```
EricDurante: Refactoring suggestion:  I think it would be better to have the work delegate to its depositor to access `email` and `display_name`.

spec/features/dashboard/work_settings_spec.rb
```diff
@@ -258,4 +258,44 @@
...
+    context 'when admin user' do
+      let(:user) { create :user, :admin }
+
+      describe 'clicking the contact depositor button' do
+        context 'when no error is raised' do
+          it 'creates a ticket in libanswers and directs to that ticket', :vcr do
+            click_button 'Contact Depositor via LibAnswers >>'
+          rescue ActionController::RoutingError
+            # This is a bit unconventional.  Since clicking the button will redirect to an external site,
+            # a routing error will be raised in the test env.  Rescue it and check the correct redirect location
```
EricDurante, line 282: Yeah, I've done exactly this sort of thing before.

spec/features/dashboard/work_settings_spec.rb
EricDurante, line 1: Just wondering - is it possible to have one test that exercises the live integration with LibAnswers? Like, could we create a separate ticket queue in LibAnswers that's just for testing or some such thing? Normally I'd suggest trying to do something like that if possible, but since this is an admin-facing feature with relatively few users, any accidental breakage would be low-risk.

So I'd say that if it's possible/easy to do, then it's maybe worth it. Otherwise, never mind.

EricDurante (changes requested):
Looks good overall. Had a few questions and suggestions for potential changes.

app/controllers/admin/create_curation_ticket_controller.rb
ajkiessl: Hmm yeah potentially.  The #submit action was leftover from when I had this as a form, so I think I will change it to the way you say. While there isn't anything in the works, I could maybe see listing tickets for a work, or maybe giving admins the capability to update or delete tickets straight from ScholarSphere. Other than that, the depositor contacting curators feature  ( #1465 ) will use the LibAnswers API.  However, that one won't be within the Admin module.  I guess we could move the controller out of the Admin module and have a before action do some kind of authentication on the action the creates tickets from the curator to depositor perspective?

spec/features/dashboard/work_settings_spec.rb
ajkiessl: If there's a way to create a test queue that doesn't notify admins when a ticket is created, then I would say yes, there's definitely a way to do this.  I can check with Briana to see if this is possible.

ajkiessl: Actually, the max number of queues is 10 and we already have 10

ajkiessl: If the test is kept out of the regularly running test suite, and only ran manually when needed, I don't think there would be a problem creating a test that does the full integration into the ScholarSphere queue.  We'd want to make sure the contents make it clear it is a test ticket and can be deleted, though.

EricDurante: Hmm. I can't say that I like the idea of automated tests doing things to a production system, and definitely not without some kind of clear separation of permissions and/or data. It's maybe not so bad as long as it's only doing non-destructive things, but it may not stay that way forever.

I'd say let's shelve this idea for now until there's a better way to do it.

EricDurante (approved):
Changes look good!

PR #1493, opened 2024-02-20
Adds citation for data and code

### LINKED ISSUE #1472: Display citation for datasets & code on work page
# Feature Request
We would like to have the citation display on the work page for datasets & code. This will require pulling metadata for title, creators, and DOI and adding ScholarSphere as the publisher. Here are some examples at other repositories:

[Dryad](https://datadryad.org/stash/dataset/doi:10.5061/dryad.sd7cd06)
[Zenodo](https://zenodo.org/records/10606425)
[Figshare](https://figshare.com/articles/dataset/Shortening_and_asthenospheric-flow_driven_Miocene_vertical-axis_rotations_of_the_NE_Qaidam_Basin_northern_Tibetan_Plateau/25144715)

Smullz622:
Fixes #1472

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -310,7 +310,25 @@ class WorkVersion < ApplicationRecord
            :auto_generated_thumbnail?,
            :thumbnail_url, to: :work
 
+  def citation_display
+    return unless work_type == "dataset" || work_type == "code"
+    "#{creators_citation_display}(#{published_date}). #{title} [Data set]. Scholarsphere. #{doi_url}"
+  end
+
   private
+    def creators_citation_display
+      formatted_creators = ''
+      creators.each_with_index do |creator, index|
+        formatted_creators += "#{creator.surname}, #{creator.given_name}"
+        formatted_creators += index == creators.length - 1 ? " " : "; "
+      end
+      formatted_creators
+    end
+
+    def doi_url
+      return unless doi.present?
+      "https://doi.org/#{doi}"
+    end
 
     def perform_update_index
       indexing_source.call(self)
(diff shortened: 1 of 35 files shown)


app/models/work_version.rb
EricDurante, line 1: A lot of this logic feels like it would be much more at home in the `CitationComponent` object since it has more to do with data presentation than data access. I haven't used ViewComponent much, but it seem like this is exactly the kind of thing that it was created for. I'd suggest playing around with refactoring this along those lines if it feels like a natural use of ViewComponent. If it doesn't, then maybe `Citation` should be its own object that gets passed to the component. I'd maybe experiment with making line 320 part of the template for the component and implementing these methods in the component or somewhere other than the model.

EricDurante (commented):
I have a suggestion for refactoring if you have time.

ajkiessl:
Could we move the copy button to the end of the citation and just have the tooltip say "Copy to clipboard"?  Like this:
Screenshot 2024-02-22 at 11 00 28 AM
It looks a bit clunky as it is.

app/models/work_version.rb
```diff
@@ -314,8 +314,31 @@ def form_partial
...
+      formatted_creators
+    end
+
+    def doi_url
+      doi.blank? ? '' : " https://doi.org/#{doi}"
+    end
+
+    def year_published
+      Date.edtf(published_date).try(:year)
+    end
```
ajkiessl: Can this stuff be removed since it is in the ViewComponent?

ajkiessl (changes requested):
Just one thing.  Everything else looks good to me.

app/models/work_version.rb
Smullz622: Yes it can. Thanks for catching that!

EricDurante (approved):
Looks good! That refactoring was definitely an improvement.

ajkiessl (approved):
Looks good!

PR #1502, opened 2024-03-05
citation doi hotfix

Smullz622:
doi was not displaying because it was pulling from work version instead of work

### DIFF
--- a/app/components/citation_component.rb
+++ b/app/components/citation_component.rb
@@ -29,7 +29,7 @@ class CitationComponent < ApplicationComponent
     end
 
     def doi_url
-      work_version.doi.blank? ? '' : " https://doi.org/#{work_version.doi}"
+      work_version.work.doi.blank? ? '' : " https://doi.org/#{work_version.work.doi}"
     end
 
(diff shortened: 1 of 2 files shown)


whereismyjetpack (approved):

whereismyjetpack:
I'm ok merging this with CI failures. my previous PR should fix this up for main

ajkiessl:
@Smullz622 Would there ever be an instance where the `identifier` field in WorkVersion has the DOI, but the minted DOI in Work is not present?

Smullz622:
@ajkiessl I think overall yes, but not for datasets (which is what this is confined to). My understanding is that they will only have a minted doi since they are only published through scholarsphere

whereismyjetpack:
If the database isn't requiring a field, we should probably prefer safe navigation _just in case_

ajkiessl:
I think `#blank?` handles nils

ajkiessl (approved):

whereismyjetpack:
i think you're right. nevermind!

PR #1503, opened 2024-03-06
add embargo to work version form

### LINKED ISSUE #1486: Move embargo date from work settings to main create/edit form
Currently a work cannot be embargoed until after it is created/published since the embargo date field does not appear in the series of forms used to create the work. It would make more sense for a depositor to be able to set an embargo date as the work is being created.

Open questions:
- Do we know why it was designed this way in the first place? Was it done purposefully for a reason that we haven't considered?
- After we make the change so that a work can be created with an embargo date, should the workflow for changing/removing the embargo date remain as is (in the work settings)? Or should this instead be done in the metadata/file editing workflow also?

Smullz622:
Fixes #1486 

Couple notes on this:
- Since the embargo form couldn't be used here, the validations are moved to the model level. Also, 'not a date' validations were no longer needed since the field was switched to a date picker instead of a string field. The date picker handles those validations automatically.
- Embargo section only shows up on v1 draft for work version and can be updated in work settings afterwards. Thought process was that we want to give users the ability to add embargo before publishing but don't want to give users the impression that they'll have to create a new work version in order to update embargo.
- This created a rails weird edge case where an invalid embargo date was rendering the 'maximum embargo is four years' error twice. There's a hacky work around for now that removes duplicate max embargo errors.


app/forms/embargo_form.rb
```diff
@@ -8,54 +8,41 @@ class EmbargoForm
...
                              nil
                            elsif embargoed_until.present?
-                             Time.zone.parse(embargoed_until).beginning_of_day
+                             parsed_embargo_date
                            end
+
+    unless work.valid?
+      work.errors.each do |attr, message|
+        errors.add(attr, message)
+      end
```
EricDurante, line 36: I haven't had a chance to fully review this yet, but one quick note here:

This code works, but it turns out that this way of iterating over errors is deprecated. The modern way to do this is
```ruby
work.errors.each do |error|
  errors.add(error.attribute, error.message)
end
```

app/models/work_version.rb
```diff
@@ -351,4 +352,15 @@ def document_builder
     def temporarily_published_draft?
       aasm.from_state == :draft && aasm.to_state == :published
     end
+
+    def remove_duplicate_errors
+      over_limit_errors = []
+      errors.errors.each_with_index do |error, index|
+        over_limit_errors << index if error.type == :max
```
ajkiessl: I'm concerned that this will affect other errors with the type `:max`.  Perhaps we can also target the `embargoed_until` field directly?  

I wonder why the WorkVersion is validating `embargoed_until` for Works.  Is that normal Rails behavior?

spec/controllers/dashboard/works_controller_spec.rb
```diff
@@ -14,7 +14,7 @@
   let(:invalid_attributes) {
     {
       'embargo_form' => {
-        'embargoed_until' => 'not a valid date'
+        'embargoed_until' => '2030-02-12'
```
ajkiessl: Could we do something like `Date.today + 5.years` here so it doesn't break in the future.  Assuming this is supposed to throw a validation error because of it being forward more than 4 years.

app/models/work_version.rb
ajkiessl: Also, we'll want a test for this.

ajkiessl (changes requested):
A couple questions/change requests below.

app/models/work_version.rb
Smullz622: Looks like we can, I'll add that & a test. As far as Eric & I could tell, it's a weird edge case. I think Eric found one stack overflow reporting the same type of problem but there wasn't a solution for it.

EricDurante: @ajkiessl 
> I wonder why the WorkVersion is validating `embargoed_until` for Works. Is that normal Rails behavior?

`WorkVersion` is validating `embargoed_until` for the associated `Work` because `WorkVersion`  is set up to accept nested attributes for the `belongs_to :work` association. That part is normal. What @Smullz622 and I observed is that when the work version's work has an invalid embargo date, validating the work version once causes a nested error to be set on the nested `work.embargo_date` attribute of the work version (that's expected). However, validating the work version more than once additionally causes another, identical nested error to be set on `work.versions.work.embargoed_until` on the work version (that's unexpected... sort of). I dug into this a little bit more myself, and I'm convinced that it's a bug (or at least an unexpected behavior) in ActiveRecord. It seems to be a consequence of using `accepts_nested_attributes_for` on both sides of an association (as we do with `Work` and `WorkVersion`) and running validations on the model more than once (which we must do in the course of publishing a work). To be fair, we're hitting a pretty weird edge case here (two-way nested attributes plus validating multiple times) that the authors of ActiveRecord maybe didn't anticipate.

I don't see any way to set up the associations or attribute nesting that would prevent this behavior from happening. So it seems that all we can do (apart from rewriting everything to not use `accepts_nested_attributes_for`) is to work around the problem by finding and removing duplicate errors after they've been set but before they're shown to the user.

EricDurante: @Smullz622 I think that the code for removing the duplicate errors definitely deserves a comment to explain why it's necessary and to acknowledge that it's a less-than-ideal workaround for what seems to be a bug in the framework.

EricDurante: For bonus points, you could open an issue in the Rails repo to describe what we saw and report it as a bug with our expected behavior being that nested errors would not be duplicated recursively from either side of the attribute nesting after repeated validation of one of the models (thankfully this stops after one recursion instead of going on forever...). That way we'd either get confirmation that it's a bug (and maybe it would get fixed), or we'll learn that we're not using something as intended or that we don't understand how something is supposed to work.

Definitely don't feel that you need to do this, but if you're interested in dipping your toes into contributing to open-source projects, then submitting an issue like this would be a small opportunity to get involved.

EricDurante: ☝️Caveat:  Since we're not on the latest version of Rails, we'd first want to confirm that this hasn't already been found and fixed.

Smullz622: @EricDurante I'll look into the Rails repo, see if it's been addressed, & create an issue if it hasn't. I do love bonus points!

EricDurante (approved):
Looks good!

ajkiessl (approved):

PR #1504, opened 2024-03-08
validating README files for the data and code deposit pathway

### LINKED ISSUE #1491: Require a README file for deposits via the Data & Code Pathway
If a work is being deposited via the Data & Code Pathway (that is, if it has one of the Data & Code work types), then the depositor should be required to upload a README file in order for the work to be published. This could be implemented as an additional validation during the final step of the deposit workflow.

All that we can really do is ensure that one of the uploaded files has a name matching 'README' and that it has a size greater than 0 bytes.

EricDurante:
resolves https://github.com/psu-libraries/scholarsphere/issues/1491

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -13,8 +13,9 @@ module Dashboard
       end
 
       def update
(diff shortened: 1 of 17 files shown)


ajkiessl (approved):
This all lgtm 👍

PR #1505, opened 2024-03-08
fix draft_curation_request bug

Smullz622:
During the go script work, I realized users could still publish after requesting draft curation. By setting `draft_curation_requested` to false in the ensure block, it could never update to true & buttons for publish & request curation were always rendered. If I remember correctly, the initial idea for having it there was for error handling, but I think rescuing the CurationError does an adequate job of handling that scenario.

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -53,7 +53,6 @@ module Dashboard
             logger.error(e)
             flash[:error] = t('dashboard.form.publish.curation.error')
           ensure
-            @resource.update_column(:draft_curation_requested, false)
             @resource.aasm_state = initial_state
           end
         end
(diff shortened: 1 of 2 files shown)


ajkiessl (changes requested):
It might be a good idea to test that the boolean gets set properly in here: https://github.com/psu-libraries/scholarsphere/blob/main/spec/features/dashboard/work_version_form_spec.rb#L1148-L1158 .  There's a lot of logic in this controller, so we should try to cover as much of it as possible somewhere.

EricDurante:
> It might be a good idea to test that the boolean gets set properly in here: https://github.com/psu-libraries/scholarsphere/blob/main/spec/features/dashboard/work_version_form_spec.rb#L1148-L1158 . There's a lot of logic in this controller, so we should try to cover as much of it as possible somewhere.

+1

ajkiessl (approved):
lgtm

PR #1506, opened 2024-03-11
Replace go script

### LINKED ISSUE #1458: Replace Go script

Smullz622:
Fixes #1458 

Couple things to note:
- While Airrecord does have a default `find` method, it takes the actual Airtable record id (not the value for the ID column) so I'm using `.all` with a filter to get the necessary uuids
- We can’t assume the latest version is what should be in Airtable, particularly if it's a draft. The latest version might be a draft with a request to curate (likely already in airtable & should be there) or a draft that is a result of a curation in progress in which case the latest published version (what is being curated) should be in Airtable instead of the draft version.
- The original 60 day window in the go script was because the only timestamp from the catalog API is for when the first draft of a work was created, not published, so 60 days was giving a reasonable margin to capture future versions. By making this an internal process, we have access to actual published timestamp & can shorten that range to 2 days (runs daily but allows for overlap)
- Originally we were considering removing stale curation tasks from Airtable, but after discussing with Admins, are holding off on that for now.  We don't want to remove their notes on existing curations when a new version is added. Instead, we're putting an 'Updated Version' label on what we’re adding if it has a version already in the table. Admins will be responsible for removing things as they see fit. It is possible we could pull notes from existing curation tasks, merge them into the entry we are adding, and remove the stale version as a future feature. 
- At this time, the error coming from Airrecord doesn’t have any helpful methods (status, code, etc) so all we can do is string match the message.


app/services/curation_task_client.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'airrecord'
+
+class CurationTaskClient
+  class CurationError < RuntimeError; end
+
+  def self.send_curation(work_version_id, requested: false)
+    Airrecord.api_key = ENV['AIRTABLE_API_TOKEN']
```
EricDurante: Do you suppose that we can move this into an initializer or something so that it just gets set once when the app boots up?

app/services/curation_task_client.rb
```diff
@@ -0,0 +1,68 @@
...
+      raise CurationError.new(e)
+    end
+  end
+
+  def self.find(uuid)
+    Airrecord.api_key = ENV['AIRTABLE_API_TOKEN']
+
+    begin
+      Submission.all(filter: "{ID} = '#{uuid}'")
+    rescue Airrecord::Error => e
```
EricDurante, line 30: You're probably already planning on doing this, but just wanted to note - I'm not sure what kinds of errors we can expect from Airrecord here and elsewhere, but it may become useful to expand `CurationTaskClient`'s error API so that it raises different classes of errors for different kinds of problems (with Airrecord, network connectivity, etc.) that may be handled differently by the caller. I think that we'll want to make an effort to anticipate and handle what we can before we release this.

app/services/curation_sync_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class CurationSyncService
+  def self.sync(work_id)
```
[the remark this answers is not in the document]
EricDurante: CodeClimate has a point here (that's not always a given). `CurationSyncService.sync` is a big method. I think we could start refactoring some of its logic into private methods. In order to do that, you'll probably need to refactor this into a proper class with state and make `sync` an instance method. So, for example, a `CurationSyncService` object could be initialized with a work to be synced, and then we'd call `CurationSyncService#sync` on the instance instead of the class (passing no argument).

Then, for example, we could extract the local variable assignment for `current_version_for_curation` into a private instance method. That would begin to simplify `CurationSyncService#sync` and reduce its responsibilities. We could probably do the same thing for some of the other local assignments.

It's reasonable for this class to be responsible for knowing what information it needs from a `Work` to do its job, but if we didn't want it to be responsible for also doing ActiveRecord stuff (pulling the work data out of the database and instantiating a `Work`), we could pass a `Work` object to the service's initializer (instead of an ID) and leave the data-finding responsibility to some other part of the code. That would reduce the overall responsibilities of this class and decouple it from ActiveRecord, making it a bit easier to test.

I'm not saying that it'll necessarily be the case here, but if a class ends up with enough private methods after a refactoring like this, and if it's not obvious that most of those private methods actually belong in some other existing part of the code, then it may become obvious that another whole class can be extracted from what was originally a single, large method.

EricDurante (commented):
I think this is looking good so far. I posted a few thoughts on refactoring.

spec/services/curation_sync_service_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+          expect(Rails.logger).to receive(:error).once
+
+          described_class.new(work).sync
+        end
+      end
+
+      context 'when it is a 5XX error' do
+        let(:message) { 'HTTP 503: Service Unavailable' }
+
+        it 'does not retry and logs the error' do
```
ajkiessl: Wording here needs changed since it is retrying

ajkiessl (changes requested):
This all looks goods to me.  I only found a wording issue with one of the tests.  The rest is good

ajkiessl (dismissed):

db/migrate/20240321203236_add_sent_for_curation_to_work_version.rb
```diff
@@ -0,0 +1,5 @@
+class AddSentForCurationToWorkVersion < ActiveRecord::Migration[6.1]
+  def change
+    add_column :work_versions, :sent_for_curation, :datetime
```
ajkiessl: I hate to nitpick, but since this is a datetime we should name it something that ends with `_at`.  So this should be `sent_for_curation_at`.

ajkiessl (changes requested):
Just one thing here I would change.

ajkiessl (approved):
lgtm

PR #1509, opened 2024-03-13
Remove identifier field from work metadata form in data & code pathway

### LINKED ISSUE #1508: Remove `identifier` field from data and code work version forms
DOIs should be minted in ScholarSphere for data and code so remove `identifier` from work submission forms.

EricDurante:
Fixes https://github.com/psu-libraries/scholarsphere/issues/1508

One thing to note here:  I rearranged the order of the optional work version metadata form fields slightly since that was the easiest thing to do. I don't know if that matters. If anyone thinks that the order of the fields should stay the same as it was before, I can do that - it'll just take a little more tweaking.

### DIFF
--- a/app/forms/work_deposit_pathway.rb
+++ b/app/forms/work_deposit_pathway.rb
@@ -60,7 +60,6 @@ class WorkDepositPathway
         subtitle
         keyword
         publisher
-        identifier
         related_url
         subject
         language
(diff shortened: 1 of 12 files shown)


EricDurante:
I reordered the work version metadata form fields per @bdezray's suggestion. I also realized that I broke the translations for the field labels on the work version details form a while ago without realizing it - so I fixed that.

ajkiessl (approved):
lgtm

PR #1510, opened 2024-03-13
Don’t show DOI minting UI for works that have a publisher DOI

### LINKED ISSUE #1487: Prevent DOIs from being minted for works for which a publisher DOI has been supplied by the depositor
We don't want it to be possible to mint a DOI in ScholarSphere for deposited works that already have a DOI from their publisher. If a value that looks like a valid DOI has been supplied as an identifier in the metadata for the most recent version of the work, then the DOI minting functionality (in the work settings) should be disabled for that work. Note:  the Data & Code deposit pathway will not include the identifier metadata field, so this value will not be able to be supplied for these types of works, and thus DOI minting will always be possible for such works.

EricDurante:
Resolves https://github.com/psu-libraries/scholarsphere/issues/1487

### DIFF
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -22,10 +22,8 @@ class WorkPolicy < ApplicationPolicy
     editable? && record.draft_version.blank?
   end
 
-  # @note we are temporarily disabling DOIs on draft works for launch. This
-  # method should probably go away when we re-enable them
   def mint_doi?
-    published? && editable?
+    published? && editable? && !record.has_publisher_doi?
   end
 
   def edit_visibility?
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -10,6 +10,7 @@ class Work < ApplicationRecord
   fields_with_dois :doi, :latest_published_version_dois
 
   delegate :email, :display_name, to: :depositor
+  delegate :has_publisher_doi?, to: :latest_version
 
   belongs_to :depositor,
              class_name: 'Actor',
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -316,6 +316,10 @@ class WorkVersion < ApplicationRecord
     self.class.model_name.param_key
   end
 
+  def has_publisher_doi?
+    !!identifier.find { |id| Doi.new(id).valid? }
+  end
+
   delegate :deposited_at,
            :depositor,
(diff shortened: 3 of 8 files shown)


app/policies/work_policy.rb
```diff
@@ -22,10 +22,8 @@ def create_version?
     editable? && record.draft_version.blank?
   end
 
-  # @note we are temporarily disabling DOIs on draft works for launch. This
-  # method should probably go away when we re-enable them
   def mint_doi?
```
EricDurante, line 25: This authorization logic applies [here](https://github.com/psu-libraries/scholarsphere/blob/fc8f4848bd29e7eae6cb1ec9f352cfeae83ec29c/app/views/dashboard/works/edit.html.erb#L124) and [here](https://github.com/psu-libraries/scholarsphere/blob/fc8f4848bd29e7eae6cb1ec9f352cfeae83ec29c/app/views/dashboard/collections/edit.html.erb#L49) to determine whether or not to show the UI for minting a DOI, but it looks like different authorization logic is being applied [here](https://github.com/psu-libraries/scholarsphere/blob/fc8f4848bd29e7eae6cb1ec9f352cfeae83ec29c/app/controllers/dois_controller.rb#L9) to determine whether or not a DOI is actually allowed to be minted, and I'm not sure why that is. It seems to me that the controller should
```ruby
authorize(resource, :mint_doi?)
```
also, but it's more lax.

Does anybody know why that is? @rschenk ?

EricDurante: I assume that it has something to do with the stale comment that I deleted above...

rschenk: I don't remember why we made this change for only published works to get DOIs, but I do seem to remember it was an afterthought or workaround for something. 

My guess is the controller is more lax because either we forgot, or  because things other than Works can get DOIs and for some insane reason we didn't want to go back and update them all? Frankly I have no idea.

bdezray: Just adding a comment to Ryan's note. Both works and collections can get DOIs

EricDurante: Alright - I'm just going to leave well enough alone here. Since there doesn't seem to be any problem with how the controller is working, I'll keep it the way that it is.

ajkiessl (approved):
lgtm

PR #1511, opened 2024-03-14
Restore admin ability to change visibility for all work types

EricDurante:
I changed this authorization behavior previously without intending to, and I think that we probably still want admins to have this ability regardless of the deposit pathway used by the work. This also moves logic out of the view and into the authorization policy and tightens up the authorization in the controller for changing visibility (which I assume we would want).


ajkiessl (approved):
lgtm

PR #1539, opened 2024-06-26
Update jest to fix vulnerability in braces

### LINKED ISSUE #1538: Scholarsphere security vulnerabilities

ajkiessl:
The remaining dependabot alerts are all downstream of redoc, which is not a production package so it is not a vulnerability (also I don't think it's even being used).  Also, it is not maintained, so there are not upgrades.  The others were fixed by dependabot's own PRs.

Also updates babel-eslint

fixes #1538

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
         user: root
     steps:
       - setup_remote_docker:
-          version: 20.10.6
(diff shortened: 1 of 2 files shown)


jlandiseigsti (approved):

Smullz622 (approved):
lgtm

PR #1561, opened 2024-07-15
discard UnknownFormat

whereismyjetpack:
This is one way to bypass sending events to bugsnag, bots will hit /catalog.rss and throw the UnknownFormat error. 

We could also return something like a 204 in the controller perhaps?


whereismyjetpack:
image

ajkiessl (approved):
Let's go with this solution for now.

PR #1562, opened 2024-07-15
remove CI for previews

whereismyjetpack:
We've migrated to the PR generator for previews in scholarsphere. Now, when you issue a PR with a branch that starts with `preview/*` a preview environment will be created, and deleted when the PR is removed. 

We can issue a PR and put it in `draft` mode to mirror the old behavior


jlandiseigsti (approved):
👍

PR #1584, opened 2024-09-12
DOI minting for data and code pathway

### LINKED ISSUE #1492: DOI minting for the Data & Code Pathway
> DOI minting is on the work settings page, but it might be nice to have the request for DOI minting be during the deposit process or be displayed as recommended just after publication. We also should require DOIs to be minted by ScholarSphere.

Open Questions:
- How exactly should this behave?

Smullz622:
fixes #1492 

When eligible, adds checkbox to publish page of work version form to request doi minting upon publish of the work. Eligible work versions are in the data & code pathway, do not already have a doi on the work, and a doi is not in the process of being minted for the work.

### DIFF
--- a/app/forms/work_deposit_pathway.rb
+++ b/app/forms/work_deposit_pathway.rb
@@ -34,7 +34,7 @@ class WorkDepositPathway
   end
 
   def allows_mint_doi_request?
-    data_and_code? && @resource.work.doi.blank? && !DoiMintingStatus.new(@resource.work).present?
+    data_and_code? && @resource.doi_blank? && DoiMintingStatus.new(@resource.work).blank?
   end
 
   def work?
--- a/app/forms/work_deposit_pathway.rb
+++ b/app/forms/work_deposit_pathway.rb
@@ -33,6 +33,10 @@ class WorkDepositPathway
     data_and_code? && !@resource.draft_curation_requested
   end
 
+  def allows_mint_doi_request?
+    data_and_code? && @resource.work.doi.blank? && !DoiMintingStatus.new(@resource.work).present?
+  end
+
   def work?
     Work::Types.all.include?(work_type)
   end
@@ -117,6 +121,7 @@ class WorkDepositPathway
                :update_doi=,
                :work_type,
                :draft_curation_requested,
+               :mint_doi_requested,
                to: :work_version, prefix: false
 
       private
@@ -258,6 +263,7 @@ class WorkDepositPathway
                  :aasm,
                  :update_column,
                  :draft_curation_requested=,
+                 :mint_doi_requested=,
                  :set_thumbnail_selection,
                  to: :work_version,
                  prefix: false
--- a/app/jobs/doi_minting_job.rb
+++ b/app/jobs/doi_minting_job.rb
@@ -4,6 +4,7 @@ class DoiMintingJob < ApplicationJob
   queue_as :doi
 
   def perform(resource)
+    sleep(2)
     status = DoiMintingStatus.new(resource)
 
     status.minting!
--- a/app/jobs/doi_minting_job.rb
+++ b/app/jobs/doi_minting_job.rb
@@ -4,7 +4,6 @@ class DoiMintingJob < ApplicationJob
   queue_as :doi
 
   def perform(resource)
-    sleep(2)
     status = DoiMintingStatus.new(resource)
 
     status.minting!
--- a/db/migrate/20240910195820_add_mint_doi_requested_to_work_version.rb
+++ b/db/migrate/20240910195820_add_mint_doi_requested_to_work_version.rb
@@ -1,0 +1,5 @@
+class AddMintDoiRequestedToWorkVersion < ActiveRecord::Migration[6.1]
+  def change
+    add_column :work_versions, :mint_doi_requested, :boolean
+  end
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,10 @@ Rails/ApplicationController:
   Exclude:
     - 'app/controllers/api/v1/rest_controller.rb'
 
+Rails/Blank:
+  Exclude:
+    - 'app/forms/work_deposit_pathway.rb' 
+
 Rails/SkipsModelValidations:
   Exclude:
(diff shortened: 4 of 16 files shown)


app/forms/work_deposit_pathway.rb
```diff
@@ -33,6 +33,10 @@ def allows_curation_request?
     data_and_code? && !@resource.draft_curation_requested
   end
 
+  def allows_mint_doi_request?
+    data_and_code? && @resource.work.doi.blank? && !DoiMintingStatus.new(@resource.work).present?
```
EricDurante: I'd suggest two refactorings here.

The first one is trivial and also much less important, but I think that the code would read a little better if we implemented `DoiMintingStatus#blank?` and used that here instead of `!` and `#present?`.

The second and more important thing is that I'd try to get rid of the method chaining here. `WorkDepositPathway` needs to know too much about the resource object that it's holding. It needs to know that the resource has a work and that the work in turn can have a DOI. It would be better if each type of thing that might act as a resource here (maybe it's only `WorkVersion`? I can't remember..) could handle telling the work deposit pathway whether or not it has a DOI directly - i.e. `@resource.doi_blank?`. That reads better too, but it also relieves `WorkDepositPathway` of needing to know about the structure/hierarchy of that other information.

db/migrate/20240910195820_add_mint_doi_requested_to_work_version.rb
EricDurante, line 1: Does this information actually need to be saved to the database? In other words, do we ever need to recall it from the database after it has been saved? Or do we only care about it for the duration of the request where this form field is submitted? Could it simply be added as a transient attribute on `WorkVersion` instead of a persistent database column?

Or maybe it doesn't even need to be an attribute of `WorkVersion` at all. Is it enough for it to just be an attribute of the form object so that we only use it to determine whether or not to queue up the DOI minting and then maybe we don't even pass it along to the `WorkVersion`?

I'm asking as if I actually know the answer, but I don't. I don't immediately see why this would need to be saved, but there might be a reason that I don't know about. I just wanted to make sure that it was something that you considered.

app/jobs/doi_minting_job.rb
```diff
@@ -4,6 +4,7 @@ class DoiMintingJob < ApplicationJob
   queue_as :doi
 
   def perform(resource)
+    sleep(2)
```
EricDurante: What was the difference in behavior that you said that you saw when you omitted this `sleep` call? Did you see it in both your development and test environments? Or was it in just one or the other?

db/migrate/20240910195820_add_mint_doi_requested_to_work_version.rb
Smullz622: So I was debating about that as well. I landed on saving it since users can save & exit the form and come back later to publish. My instinct was that users who previously checked the box might assume it was still checked, submit without rechecking, and incorrectly think they're minting a doi upon publish. I could still go either way though & am open to suggestion.

EricDurante: Ah! I hadn't thought of that, and it totally makes sense. You're way ahead of me. 🤐

app/jobs/doi_minting_job.rb
Smullz622: So test environment is fine either way, but in local dev & the preview branch, without the sleep, v1 work versions do not successfully mint a doi. It will mint one on v2 of the same work successfully though. I had trouble figuring out why this is because it is hitting the DoiMintingJob but not throwing any type of error that I can tell. I only found out the sleep works because I put in a byebug to dig further & it worked that time so I replaced it with a sleep & then everything worked as expected. Also why my suspicion is that it's something around timing with when it sends the record data to data cite.

Smullz622: Though that being said, this is the only spot the sleep made a difference. I also tried one before `if @resource.mint_doi_requested` in `publish_controller` and at the beginning of the call method for MintDoiAsync and neither changed anything

app/forms/work_deposit_pathway.rb
Smullz622: First one I agree though as a note, `#present?` here is not a normal present. It's actually a custom method of `DoiMintingStatus` (threw me off for a long time trying to troubleshoot because `#present?` and `#blank?` were both returning false). But thinking on it, I can also just do a custom `#blank?` on DoiMintingStatus so we can still follow typical conventions.

Second sounds good & I'll get working on that.

app/jobs/doi_minting_job.rb
ajkiessl: I think I was able to get this to work locally without the sleep by disabling ActiveJob in the config for the dev environments:

https://github.com/psu-libraries/scholarsphere/blob/main/config/environments/development.rb#L19-L20

Looks like we override SideKiq in the development environment and just use ActiveJob instead.  The minting still failed for me, but I think it's rejecting my url because it's a localhost url.  But the job queued and ran.  This doesn't explain why it's not working on the preview branch, though.  It should be using SideKiq there.

ajkiessl: Also, I don't know why it works with SideKiq but not ActiveJob.

ajkiessl: Ah I see Eric has a fix for this.  Using SideKiq probably just delayed the job enough to mimic what the sleep was doing.  My comments can be ignored.

app/forms/work_deposit_pathway.rb
```diff
@@ -33,6 +33,10 @@ def allows_curation_request?
     data_and_code? && !@resource.draft_curation_requested
   end
 
+  def allows_mint_doi_request?
+    data_and_code? && @resource.doi_blank? && DoiMintingStatus.new(@resource.work).blank?
```
ajkiessl, line 37: I think we should have some unit tests for this, since there's a handful of conditions.

ajkiessl (changes requested):
This looks good.  I just have one change request.

ajkiessl (dismissed):
Looks good, thanks!

ajkiessl (approved):

PR #1591, opened 2024-09-18
1578 - Adds a unique destroy? policy for file version membership

### LINKED ISSUE #1578: Allow deletion of file for non-latest WorkVersions
We delegate our `Pundit` policies for `WorkVersions` to `FileVersionMemberships` when deleting a file.  Admins are given the ability to delete `WorkVersions` in this policy as long as the `WorkVersion` in question is the most recent version.  This makes sense since we don't want anyone deleting a V1 when there's a V2 without deleting the V2 first.  This doesn't work so well with `FileVersionMemberships`.  Sometimes, admins need to delete old files, even for versions that are not the latest.

I believe the way to fix this problem would be to add a `destroy?` policy to `FileVersionMembershipPolicy`.  Then, rather than authorizing the WorkVersion in `app/controllers/dashboard/file_version_memberships_controller.rb` like this:

```
  def destroy
    authorize(@work_version)
    @file_version.destroy
```

we do it like this:

```
  def destroy
    authorize(@file_version)
    @file_version.destroy
```

jlandiseigsti:
closes #1578

The fileVersionMembershipPolicy used to just delegate to the destroy policy of the WorkVersionPolicy. However, we want a slight change – while we don't want to be able to delete the non-recent work versions, we want to be able to delete files even if they are associated with older versions.

### DIFF
--- a/app/controllers/dashboard/file_version_memberships_controller.rb
+++ b/app/controllers/dashboard/file_version_memberships_controller.rb
@@ -27,7 +27,7 @@ module Dashboard
     end
 
     def destroy
-      authorize(@work_version)
+      authorize(@file_version)
       @file_version.destroy
       respond_to do |format|
         format.html do
--- a/app/controllers/dashboard/file_version_memberships_controller.rb
+++ b/app/controllers/dashboard/file_version_memberships_controller.rb
@@ -32,7 +32,7 @@ module Dashboard
       respond_to do |format|
         format.html do
           redirect_to dashboard_form_files_path(@work_version),
-                      notice: 'Work version was successfully destroyed.'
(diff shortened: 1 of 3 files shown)


ajkiessl (changes requested):
This looks good.  I did notice that the flash message set in the controller for deleting a file is 'Work version was successfully destroyed'.  Since we're making changes in the area, we might as well fix this to say something like 'File was successfully destroyed'.

https://github.com/psu-libraries/scholarsphere/blob/f4419caf5424f8000e519e7bd7c1e7d2ee36e9db/app/controllers/dashboard/file_version_memberships_controller.rb#L30-L38

ajkiessl (approved):
LGTM

PR #1595, opened 2024-09-25
Create instruments work type and pathway

### LINKED ISSUE #1586: Create a New Work Type and Pathway for Instruments
Description: Develop a new work type and pathway specifically for instruments. 

Requirements:
1. Create a New Work Type: Instrument: Add a new work type labeled "Instrument" to the system, accommodating the specific metadata requirements for instruments.

2. Develop a Dedicated Pathway for Instrument Work Type
**Metadata Fields for Instrument Work Type (Mapped to DataCite Schema)**

**_Identifier_**
The DOI that you are going to create. Add as DataCite property Identifier with identifierType=DOI.

**_Name_**
The name by which this instrument is known. Add as DataCite property Title.
**_Owner_**
The organization or individual that manages the instrument. Add as DataCite property Contributor with contributorType=HostingInstitution. Consider also to add an identifier for the owner in the nameIdentifier subproperty of Contributor.
**_Manufacturer_**
The organization or individual that built the instrument. Add as DataCite property Creator. Consider also to add an identifier for the manufacturer in the nameIdentifier subproperty of Creator.
**_Model_**
The name of the model or type of the instrument. As of this writing, the DataCite Schema has no specific property for that. The mapping provided by DataCite suggest to add it as a Description with descriptionType=TechnicalInfo, see note below.
**_Description_**
A textual description of the device and its capabilities. Add as DataCite property Description with descriptionType=Abstract.
**_InstrumentType_**
A classification of the type of the instrument. As of this writing, the DataCite Schema has no specific property for that. The mapping provided by DataCite suggest to add it as a Description with descriptionType=TechnicalInfo, see note below. An alternative might be to add it as keywords providing such a classification in the Subject property. The latter might be particularly useful if the instrument type is using terms from a controlled vocabulary, as Subject allows to link those terms using the subjectScheme, schemeURI, and valueURI subproperties.
**_MeasuredVariable_**
The variables or physical properties that the instrument measures or observes. As of this writing, the DataCite Schema has no specific property for that. The mapping provided by DataCite suggest to add it as a Description with descriptionType=TechnicalInfo, see note below.
**_Date_**
Relevant events pertaining to this instrument instance. Add as DataCite property Date. Use dateType=Available to indicate the date that the instrument is or was in operation. Use a single date if the instrument is still in operation, to indicate a start date. Use a date interval to indicate a start and an end date, if the instrument has already been decommissioned.
**_RelatedIdentifier_**
This can be used to establish links to related resources. The DataCite Schema has a property with the same name, having very similar subproperties and semantics as the PIDINST Schema.
**_AlternateIdentifier_**
To be used if this instrument is also registered elsewhere. Add as DataCite property AlternateIdentifier. Use alternateIdentifierType=SerialNumber for a serial number attributed by the manufacturer. Use alternateIdentifierType=InventoryNumber for an inventory number used by the owner.
**_Publisher_**
Set to ScholarSphere
**_PublicationYear_**
Mandatory in the DataCite Schema.  Set it to the year of issuing the DOI.
**_ResourceType_**

**_FundingReference_**
Optional. Acknowledgement of external funding that supported the purchase or creation of the instrument.

Smullz622:
Fixes #1586


openapi.yml
```diff
@@ -334,6 +334,40 @@ components:
...
+          example: "Classifying Independent Hybridity"
+        alternative_identifier:
+          type: "string"
+          example: "Classifying Independent Hybridity"
+        instrument_resource_type:
+          type: "string"
+          example: "Classifying Independent Hybridity"
+        funding_reference:
+          type: "string"
+          example: "Classifying Independent Hybridity"
```
jlandiseigsti: Do we care about these examples? If so, should they all be the same?

spec/components/citation_component_spec.rb
```diff
@@ -56,5 +57,46 @@
...
+        let(:expected_citation) { 'Grant, Alan (2024). Citation Title [Data set]. Scholarsphere. https://doi.org/10.26207/123' }
+
+        it 'renders the citation' do
+          expect(citation_component.css('div.keyline')).not_to be_empty
+          expect(citation_component.text).to include(expected_citation)
+        end
+      end
+
+      context 'when there are multiple creators' do
+        let(:expected_citation) { 'Grant, Alan; Sattler, Ellie; Malcolm, Ian (2024). Citation Title [Data set]. Scholarsphere. https://doi.org/10.26207/123' }
```
jlandiseigsti, line 89: There's uh, another example. See, here, now I'm sitting by myself, uh, uh, talking to myself. That's-that's chaos theory.

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -30,6 +30,7 @@ def update
           # WorkVersion#set_thumbnail_selection may be unreliable if the Shrine::ThumbnailJob is delayed
           @resource.set_thumbnail_selection
           @resource.indexing_source = Proc.new { nil }
+          @resource.set_publisher_as_scholarsphere if deposit_pathway.instrument?
```
jlandiseigsti: This logic could also potentially go in the model...a `before_save` or even as part of the act of publishing.

jlandiseigsti (commented):
Looks good to me! I'll let someone more familiar with this repo give the official approve but it looks like good test coverage and good organization.

openapi.yml
Smullz622: I believe we do, but I still need to touch base with Briana about what those values would actually be

Smullz622: @ajkiessl do you know if we want to have examples here? Looking further, it looks like some have them & some don't

ajkiessl: I think these will be fine without examples. If anyone is using our API for uploading instruments, they'll probably know what these fields should be, or they'll at least be in close contact with us if they need to know.

app/components/thumbnail_component.rb
```diff
@@ -60,6 +60,7 @@ def icon_map
                                       dataset: 'analytics',
                                       dissertation: 'subject',
                                       image: 'image',
+                                      instrument: '',
```
ajkiessl: You should be able to select anything from here: https://material.io/resources/icons to be the default image for instruments. Might be good to have a default, even though they're required to upload an image. I think anything that looks like a machine or tool would work.

app/controllers/catalog_controller.rb
```diff
@@ -142,6 +142,17 @@ def index
...
+    config.add_show_field 'manufacturer_tesim', label: 'Manufacturer'
+    config.add_show_field 'model_tesim', label: 'Model'
+    config.add_show_field 'instrument_type_tesim', label: 'Instrument type'
+    config.add_show_field 'measured_variable_tesim', label: 'Measured variable'
+    config.add_show_field 'available_date_tesim', label: 'Available date'
+    config.add_show_field 'decommission_date_tesim', label: 'Decommission date'
+    config.add_show_field 'related_identifier_tesim', label: 'Related identifier'
+    config.add_show_field 'alternative_identifier_tesim', label: 'Alternative identifier'
+    config.add_show_field 'instrument_resource_type_tesim', label: 'Instrument type'
+    config.add_show_field 'funding_reference_tesim', label: 'Funding reference'
```
ajkiessl: To be consistent, we should make these title cased.  I actually, don't think these blacklight show fields are used anywhere.  Or they are at least overwritten here: https://github.com/psu-libraries/scholarsphere/blob/main/app/components/work_version_metadata_component.rb .  Ultimately, it looks like it's pulling labels from the en.yml: https://github.com/psu-libraries/scholarsphere/blob/main/config/locales/en.yml#L54-L81

spec/features/dashboard/work_version_form_spec.rb
```diff
@@ -876,6 +990,32 @@
         expect(page).not_to have_field('work_version_work_attributes_visibility_authenticated')
       end
     end
+
+    context 'with a work that uses the instrument works deposit pathway' do
+      let(:work) { create :work, :instrument, versions_count: 1 }
+      let(:work_version) { work.versions.first }
+      let(:user) { work.depositor.user }
+
+      it 'shows only the fields for a scholarly work' do
```
ajkiessl: This should be 'fields for an instrument'.

app/views/dashboard/form/type/_work_version_fields.html.erb
```diff
@@ -4,10 +4,20 @@
...
+                         {},
+                         { class: 'form-control custom-select',
+                           id: 'work_version_work_attributes_work_type',
+                           data: { target: 'work-type.workTypeSelect' },
+                           disabled: local_assigns[:work_type_disabled] } %>
   <% end %>
+
+  <div id="work-title-help-text" data-target="work-type.helpText" style="display: none;">
+    <p><%= t 'dashboard.form.details.instrument_title' %></p>
+  </div>
```
ajkiessl: Adding some bootstrap classes and making the `p` element `small` will make this hint look like the one above:
```
<div class="form-text text-muted" id="work-title-help-text" data-target="work-type.helpText" style="display: none;">
    <small><%= t 'dashboard.form.details.instrument_title' %></small>
  </div>
```

spec/features/dashboard/work_version_form_spec.rb
```diff
@@ -1243,10 +1383,49 @@ def mock_solr_indexing_job
...
+      it 'does not edit the publisher field' do
+        visit dashboard_form_publish_path(work_version)
+        check 'I have read and agree to the deposit agreement.'
+        click_on 'Publish'
+
+        expect(work_version.reload.publisher).to eq []
+      end
+    end
+  end
+
```
ajkiessl: Since this is being done in the model now, could we remove these and just make some unit tests in the `WorkVersion` model tests?  That could save us some time in CI.

spec/models/work_version_spec.rb
```diff
@@ -85,6 +85,10 @@
   end
 
   describe 'states' do
+    subject(:work_version) { build(:work_version, work: work) }
+
+    let(:work) { build(:work) }
```
ajkiessl: I'm curious why this was added?

ajkiessl (changes requested):
This is looking good.  Most of my requests are pretty minor.  There are some questions in there too.

spec/models/work_version_spec.rb
Smullz622: This was one of the weird aasm state things & part of why I was debating about where the logic should be for automatically setting publisher to scholarsphere for instruments. Originally I had it in the publish controller but didn't love that. 

I moved it to the aasm state proc on the publish state (effectively the publish action). So to transition to publish, it has to be able to check work type, but work type is delegated to work. This test was failing because work was nil, and it wasn't able to transition to publish.

ajkiessl: Oh I see.  That makes sense, thanks!

app/models/work_version.rb
```diff
@@ -31,7 +31,18 @@ class WorkVersion < ApplicationRecord
...
                  related_url: [:string, array: true, default: []],
-                 source: [:string, array: true, default: []]
+                 source: [:string, array: true, default: []],
+                 owner: :string,
+                 manufacturer: :string,
+                 model: :string,
+                 instrument_type: :string,
+                 measured_variable: :string,
+                 available_date: :string,
+                 decommission_date: :string,
```
ajkiessl, line 41: Sorry I didn't notice this before.  We should run these dates through the date validation that `published_date` runs through below:

```
validates :published_date,
            presence: true,
            edtf_date: true,
            if: :published?
```

app/controllers/catalog_controller.rb
```diff
@@ -142,6 +142,17 @@ def index
...
+    config.add_show_field 'manufacturer_tesim', label: 'Manufacturer'
+    config.add_show_field 'model_tesim', label: 'Model'
+    config.add_show_field 'instrument_type_tesim', label: 'Instrument Type'
+    config.add_show_field 'measured_variable_tesim', label: 'Measured Variable'
+    config.add_show_field 'available_date_tesim', label: 'Available Date'
+    config.add_show_field 'decommission_date_tesim', label: 'Decommission Date'
+    config.add_show_field 'related_identifier_tesim', label: 'Related Identifier'
+    config.add_show_field 'alternative_identifier_tesim', label: 'Alternative Identifier'
+    config.add_show_field 'instrument_resource_type_tesim', label: 'Resource Type'
+    config.add_show_field 'funding_reference_tesim', label: 'Funding Reference'
```
ajkiessl: The edit and show pages actually pull labels from here: https://github.com/psu-libraries/scholarsphere/blob/main/config/locales/en.yml#L54-L81 .  We should add the fields that need to be title-cased here.

ajkiessl (changes requested):
Just two things below, but everything else looks good!

ajkiessl:
I just noticed this part of the open_api.yml while working on the culminating experiences stuff:

https://github.com/psu-libraries/scholarsphere/blob/70a1606f31a1b08d04f130bfb927e05818754cb3/openapi.yml#L260-L284

We'll want instruments in here too.

app/models/work_version.rb
Smullz622: In the EdtfDate module `#valid?` required the date to be a valid and present to return true so an empty available and/or decommission date would throw an error. This feels redundant because `presence: true` on `published_date` already handles this. I modified it so the module just validates format, allowing for blank values, and let `presence: true` handle that validation as needed.

ajkiessl (dismissed):
LGTM

ajkiessl (approved):
LGTM

PR #1607, opened 2024-10-01
allow admins to remove umg groups from editors

### LINKED ISSUE #956: Managing UMG edit groups
# Feature Request

We should see a list of the current groups that have edit access to a resource. Presently, there's just an multi-select form and you'd only know if a group was there if _you_ the logged in users where a member of it.

A listing of current UMG groups that have edit access, probably above the select box? UI design is up for discussion.

Smullz622:
Fixes #956


ajkiessl (approved):
LGTM Thanks!

PR #1610, opened 2024-10-07
Allow depositors to request accessibility remediation and save as draft

### LINKED ISSUE #1603: Ability for Depositor to Save Work Draft and Request Accessibility Remediation Assistance
From the epic (https://github.com/psu-libraries/scholarsphere/issues/1589):

>Similar to how depositors can currently request curation assistance for datasets prior to publication, we may want depositors of articles and other work types to have the ability to do the same thing in order to request manual review/assistance from a ScholarSphere curator regarding document accessibility prior to publishing their work. If a depositor selected this option, their work would be saved as a draft, and the work would enter a queue for review by a curator.

>Open questions:
>Does this apply to all non-data/code work types?
>Should this always be an option for the applicable work types, or should it only be an option if one or more files doesn't pass automated accessibility checking? (The latter would require us to make the depositor wait for feedback before moving on..)
>Do we want the list of these curation requests to live in AirTable along with all of the other curation tasks? If so, how should they be identified or distinguished from other tasks?

Part of #1589

Smullz622:
Fixes #1603 

Currently, this pr allows accessibility remediation to be requested for any work type as long as it has not already been requested & draft curation has not been requested. Additional logic will need to be added so that the button does not show up if all files have passed the automatic checker, but that is on hold while the checker is being developed. Depending on timing, that addition will go in another ticket.

I updated the language for both request buttons to drop the '& Save as Draft' part. With the current number of buttons, it took up too much space & buttons were starting to overlap. Instead, I added that language to the help text so users still have an indicator that the work is being saved as a draft.

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -35,30 +35,14 @@ module Dashboard
           @resource.save
           @resource.publish
         elsif curator_action_requested?
-          column = request_curation? ? :draft_curation_requested : :accessibility_remediation_requested
-
-          @resource.update_column(column, true)
-          # We want validation errors to block curation requests and keep users on the edit page
-          # so WorkVersion needs to temporarily act like it's being published. It's returned to it's
-          # initial state before being saved.
           begin
-            @resource.save
-            initial_state = @resource.aasm_state
-            @resource.publish
-            if @resource.valid?
-              CurationTaskClient.send_curation(@resource.id, requested: request_curation?, remediation_requested: request_accessibility_remediation?)
-              request_curation? ? @resource.draft_curation_requested = true : @resource.accessibility_remediation_requested = true
-            else
-              @resource.update_column(column, false)
-              render :edit
-              return
-            end
-          rescue CurationTaskClient::CurationError => e
-            @resource.update_column(column, false)
+            DepositorRequestService.new(@resource).request_action(request_curation?)
+          rescue DepositorRequestService::RequestError => e
             logger.error(e)
             flash[:error] = request_curation? ? t('dashboard.form.publish.curation.error') : t('dashboard.form.publish.remediation.error')
-          ensure
-            @resource.aasm_state = initial_state
+          rescue DepositorRequestService::InvalidResourceError
+            render :edit
+            return
           end
         end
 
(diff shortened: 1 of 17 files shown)


app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -34,8 +34,10 @@ def update
           @resource.indexing_source = Proc.new { nil }
           @resource.save
           @resource.publish
-        elsif request_curation? && !@resource.draft_curation_requested
-          @resource.update_column(:draft_curation_requested, true)
+        elsif curator_action_requested?
+          column = request_curation? ? :draft_curation_requested : :accessibility_remediation_requested
+
+          @resource.update_column(column, true)
```
ajkiessl: I think these changes are fine.  If we want this to be a bit cleaner or easier to read, we could pull all of this curation logic out into it's own service class.

ajkiessl (commented):
This is looking good.  I'll wait to approve when the description wording is added.  We should consider what will happen when someone submitting data & code wants curation and remediation.  If accessibility remediation is requested, the work will still be synced to airtable during our daily syncs, right?

app/controllers/dashboard/form/publish_controller.rb
ajkiessl: Or i guess it will already be up in AirTable, so the CurationSyncService will exclude it.

ajkiessl: Ah nevermind. That's only for published works.

ajkiessl:
> This is looking good. I'll wait to approve when the description wording is added. We should consider what will happen when someone submitting data & code wants curation and remediation. If accessibility remediation is requested, the work will still be synced to airtable during our daily syncs, right?

Ah nevermind. That's only for published works.

app/controllers/dashboard/form/publish_controller.rb
Smullz622: My understanding is that curation will include accessibility review (it's currently listed as one of the services in the curation request description), but I agree we need to do something to make that more clear. Might be as simple as updating the description with language emphasizing that curation includes remediation so if you want both, select curation.

I'll play around with moving it to a service class. I tried just pulling it into a private method, but there were some issues with error raising/number of redirects, but I might be able to resolve those if it's a separate service class.

ajkiessl: Ah, I see that now in the description.  Yeah, I think it would be good to emphasize the accessibility part in the curation description a bit more.

ajkiessl (approved):
LGTM!

PR #1612, opened 2024-10-10
add Airtable label when a work version needs accessibility review

### LINKED ISSUE #1604: Ability for Curators to Review/Track Each Deposited Work
From the epic (#1589 )

>An automated check may be the best way to provide quick feedback to depositors about easily detected/correctable accessibility issues. However, we know that we likely won't be able to run an automated accessibility check on every type of file that is uploaded to ScholarSphere (in fact, we may only be able to do so with PDF files). We also know that a passing automated check does not at all guarantee that a file is actually accessible by all WCAG standards. The only way to guarantee that any given file meets the standards is for a human reviewer to carefully inspect it and remediate any issues.

>To ensure that all works deposited in ScholarSphere are actually accessible, we would need each work (regardless of whether it was automatically checked) to enter into a queue for manual review by a ScholarSphere curator. After a curator has inspected each file that's attached to the current work version and worked with the depositor to remediate any issues, they would need a way to manually indicate that the work had been reviewed for accessibility and thus remove it from the queue. If a new version of the work is created in the process of remediating accessibility issues, then the work should not re-enter the queue, but if a new version of the work is created for any other reason, then it should re-enter the queue for another manual accessibility review.

>This queue for manual review would be separate from the queue for works where the depositor actively requested an accessibility review from a curator prior to publication and should exclude works where curation was requested so that the curation of works in these two queues can be prioritized separately. The mechanism for resolving/removing works from both queues could probably be the same.

>Open questions:

>Again, could/should this list be tracked in AirTable?
>Is this actually a reasonable thing to do? In other words, will there be enough curators to handle this work? Or will the to-do list just get hopelessly long?

Part of #1589

Smullz622:
Fixes #1604 

Even if all files pass the auto checker, we want this label added so it gets eyes on it to be sure. The only time this label is not needed is if it isn't the latest published version (only applicable when draft curation is requested) or it has remediation requested meaning it has already undergone the manual accessibility check process.

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -337,6 +337,10 @@ class WorkVersion < ApplicationRecord
     !!identifier.find { |id| Doi.new(id).valid? }
   end
 
+  def needs_accessibility_review
+    latest_published_version? && !accessibility_remediation_requested
+  end
+
   delegate :deposited_at,
            :depositor,
            :embargoed?,
--- a/app/services/curation_task_client.rb
+++ b/app/services/curation_task_client.rb
@@ -12,6 +12,7 @@ class CurationTaskClient
     labels << 'Embargoed' if submission.embargoed?
     labels << 'Updated Version' if updated_version
     labels << 'Accessibility Remediation Requested' if remediation_requested
+    labels << 'Needs Accessibility Review' if submission.needs_accessibility_review
 
     record =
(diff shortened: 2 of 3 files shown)


ajkiessl (approved):
Looks good.  Thanks!

PR #1614, opened 2024-10-11
Hotfixes for Culminating Experiences

ajkiessl:
Adds custom error messages for culminating experiences metadata fields.  Also, fixes params bug in publish controllers


Smullz622 (approved):
lgtm!

PR #1617, opened 2024-10-14
1477 help text buttons

### LINKED ISSUE #1477: Help text for buttons at bottom of work deposit form and publication warning
Add help text that describes the what action will occur by clicking each button when hover over button. It would also be helpful to add a before you publish warning similar to the attached. 
[Mendeley_Data_Before_You_Publish_Screenshot.docx](https://github.com/psu-libraries/scholarsphere/files/14166774/Mendeley_Data_Before_You_Publish_Screenshot.docx)

jlandiseigsti:
closes #1477 

Adds dialogs for all the final steps – publishing, requesting curation, and requesting accessibility mediation.


Smullz622 (commented):
@jlandiseigsti when I test this out locally, cancel & delete work, but save & exit, request curation, and publish do not. The help text is there, but the agreement box isn't popping up which blocks the button action from proceeding. I've seen it work during your demo, so I know it does, but it's probably worth digging into why it's not working on mine (mostly concerned if this is something browser dependent). I'm free to pair on this if you want to look at it together.

jlandiseigsti:
@Smullz622 Hmm, yes, let's dig into this! I wonder what the difference is

Smullz622 (approved):
Looks good! Turns out assets weren't compiling on my end

app/models/work_version.rb
```diff
@@ -117,6 +117,18 @@ def label(id)
             acceptance: true,
             if: :published?
 
+  validates :psu_community_agreement,
+            acceptance: true,
```
ajkiessl, line 121: It took me a while to realize `acceptance: true` rejects falsy things _except_ nil.  That's why we don't need to worry about these with API ingests (like RMD stuff).

ajkiessl (approved):
I like the way this looks.  Thanks!

Issue #1613, opened 2024-10-10
Storage and Presentation of Accessibility Check Results


ajkiessl:
There are several parts to this, so this may need to be an epic of it's own.

**Data model:** A new data model (maybe `AccessibilityCheckResult`) that `belongs_to` a `FileResource` and a `has_one` relationship from `FileResource` to `AccessibilityCheckResult`.  There shouldn't be more than one accessibility check for a `FileResource`.   Alternatively, we could store the results of the accessibility check with FileResource.

The `AccessibilityCheckResult` should store whether the check passed or failed, and the contents of the report.  Since the report is in JSON, we could try storing the report as json-b in the database.  These both should be required fields if using a separate data model.

~~**Parsing the data from the report:** This depends on how #1602 is implemented.  We could pull our file from S3, store it temporarily, then pass it to Adobe to be checked.  Adobe will send us a JSON response with the report.  Another, option is to send a presigned URL to Adobe, they download the file to be checked, and they send us a zip file containing the original file and the JSON report.  How we parse will depend on which implementation is chosen.~~  We'll go with the solution that doesn't involve presigned urls.  Since the parsing can be done in #1602, we don't need to worry about this in this issue.

**Presentation:** The UI on the publish page already lists the uploaded files.  We could add another column in that list that gives the pass/fail outcome of the accessibility check.  There will need to be some polling to check the database for when the `AccessibilityCheckResult` is finally stored.  We'll also want a link to download the report.  The report should be human readable, not JSON.  If any of the files fails the check, there should be a link at the end of the list of files directing the user to a site for help on remediation.  We'll need to get the link for this site.

We'll need a place in the "Work Settings" page that lists the files and a link to download the reports.  We want both edit users and admins to be able to access these.

Smullz622:
The result column added to the table should have "_ out of _ passed" if the checker has completed, 'Processing' if the checker is still running, and 'Needs manual review' for non pdf files that can't run through the checker (or perhaps if there's an error in the checker & it can't complete). This will need to be dynamic (consider hotwire/stimulus) & give real time updates since we are blocking publish until the checker finishes.

PR #1627, opened 2024-10-18
Accessibility Checker

### LINKED ISSUE #1602: Automated Accessibility Checking
From the epic (#1589):

>If feasible, the first step in this process would be an automated check of each document that is uploaded to ScholarSphere during the workflow for depositing a work. This check would use a third-party tool to analyze and evaluate each document in terms of accepted accessibility standards. We've begun to describe this step in https://github.com/psu-libraries/scholarsphere/issues/1580. We would want this automated check to be triggered by each document upload and to run asynchronously in the background. The result of the check should be a pass/fail evaluation of each document along with a list of any detected accessibility issues (and suggested remediations) for each document.

We have decided to use Adobe's PDF Services API to do our accessibility checking.  Refer to https://github.com/psu-libraries/scholarsphere/issues/1580#issuecomment-2359209601 for some links to documentation.  We will need to build our own wrapper that returns a pass/fail analysis and parses out any issues and remediations given to us by the API.  Accessibility checking should kick off in an async Sidekiq job for each uploaded PDF after the file upload step in the submission workflow.

There may be a way to pass signed URLs from ScholarSphere's AWS S3 bucket to Adobe for checking, so let's investigate that before considering other options.  Here's the documentation for passing signed URLs: https://developer.adobe.com/document-services/docs/overview/pdf-services-api/howtos/pdf-external-storage-sol/ .

We'll also need to keep in mind that we only get 500 document checks a month on the free version. (https://github.com/psu-libraries/scholarsphere/issues/1580#issuecomment-2361076606)

ajkiessl:
closes #1602 

Module for pulling a file from S3, sending it up to Adobe through the AdobePdf API, accessibility checking, storing the accessibility report, and deleting the file in Adobe.  There's some over lap with #1613 with storing the report, but it looks like the conflicts are minimal.  I also threw a small accessibility fix in here to make the three required fields for culminating experiences required in the view.

I added an issue (#1638) to decouple the AdobePdf module from ScholarSphere.  It's not a priority, but it might be helpful architecturally, and if we want to use this elsewhere.  I also added an issue (#1637) to use setters for storing the report from Adobe.  Again, not a priority, but it could help us keep track of changes Adobe makes to the report's structure.


app/services/adobe_pdf/base.rb
```diff
@@ -0,0 +1,70 @@
...
+      def client_secret
+        if Rails.env.test?
+          'adobe_client_secret'
+        else
+          ENV.fetch('ADOBE_CLIENT_SECRET', 'adobe_client_secret')
+        end
+      end
+
+      def host
+        'https://pdf-services.adobe.io'
```
jlandiseigsti: I have a slight preference for this value in ENV as well.

app/services/adobe_pdf/base.rb
```diff
@@ -0,0 +1,70 @@
...
+        if Rails.env.test?
+          'adobe_client_id'
+        else
+          ENV.fetch('ADOBE_CLIENT_ID', 'adobe_client_id')
+        end
+      end
+
+      def client_secret
+        if Rails.env.test?
+          'adobe_client_secret'
```
jlandiseigsti, line 34: Something to consider (probably down the line) – we could consider having distinct env files for different environments, so that this sort of logic wouldn't be in the code itself.
For example, https://github.com/direnv/direnv/issues/320#issuecomment-388341274

app/services/adobe_pdf/asset.rb
```diff
@@ -0,0 +1,113 @@
...
+      #         puts "Running additional operations on asset with ID: #{asset_id}"
+      #       end
+      #     end
+      #   end
+      #
+      #   resource = FileResource.new(file_data: { 'id' => '123', 'storage' => 's3' })
+      #   my_asset = MyAsset.new(resource)
+      #   my_asset.perform_operations
+      #
+      # @yield The block to run additional operations on the asset.
```
jlandiseigsti, line 42: Really helpful to include an example like this, thanks!

app/services/adobe_pdf/asset.rb
```diff
@@ -0,0 +1,113 @@
...
+        delete_asset
+      end
+
+      def upload_uri
+        @upload_uri ||= asset_upload_uri_and_asset_id['uploadUri']
+      end
+
+      def asset_id
+        @asset_id ||= asset_upload_uri_and_asset_id['assetID']
+      end
```
jlandiseigsti, line 57: Could this and `upload_uri` potentially get set during initialize as well?

app/services/adobe_pdf/accessibility_checker.rb
```diff
@@ -0,0 +1,87 @@
...
+          else
+            raise AdobePdfApiError, "Failed to get Accessibility Checker status: #{response.status} - #{response.body}"
+          end
+          counter += 1
+          sleep 2
+        end
+        AccessibilityCheckResult.create!(file_resource: resource,
+                                         detailed_report: {
+                                           error: 'Accessibility check failed: Polling time limit exceeded'
+                                         })
```
jlandiseigsti, line 51: I noticed code climate's Complexity warning, but I'm not seeing a lot of obvious places to trim this method down. One very minor thing could be to pull lines 48-51 into some kind of `createTimeoutResult` method.

spec/services/adobe_pdf/accessibility_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+
+require 'rails_helper'
+
+RSpec.describe AdobePdf::AccessibilityChecker, :vcr do
+  let(:checker) { described_class.new(resource) }
+
+  describe '#initialize' do
+    context 'when the file size exceeds the limit' do
+      let(:resource) { instance_double('FileResource', file_data: { 'metadata' => { 'size' => file_size } }) }
+      let(:file_size) { 150_000_000 }
```
jlandiseigsti: We could consider making this `FILE_SIZE_LIMIT * 1.5`, so that if the limit every changed, this would adjust with it.

spec/services/adobe_pdf/accessibility_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+
+      it 'raises a FileSizeExceededError' do
+        expect { checker }.to raise_error(AdobePdf::AccessibilityChecker::FileSizeExceededError, 'File size exceeds the limit of 100Mb')
+      end
+    end
+
+    context 'when the file size does not exceed the limit' do
+      let(:resource) { instance_double('FileResource', file_data: { 'metadata' => { 'size' => file_size } }) }
+      let(:file_size) { 50_000_000 }
```
jlandiseigsti: Similarly, `FILE_SIZE_LIMIT * 0.5`

spec/services/adobe_pdf/accessibility_checker_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+                   'Summary' => {
+                     'Description' => 'The checker found problems which may prevent the document from being fully accessible.',
+                     'Failed' => 18,
+                     'Failed manually' => 0,
+                     'Needs manual check' => 2,
+                     'Passed' => 12,
+                     'Passed manually' => 0,
+                     'Skipped' => 0
+                   }
+                 })
```
jlandiseigsti, line 126: This feels brittle to me, as even slight changes to how Adobe formats their responses would require test rewirites. I'm not sure what the best alternative, but I'd be content know we got a result containing 'Detailed Report'.

app/services/adobe_pdf/base.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module AdobePdf
+  # TODO: This whole module can potentially be decoupled from scholarsphere.
```
jlandiseigsti, line 4: I'm glad we have this on our todo list. Crossing fingers for reusability!

jlandiseigsti (dismissed):
Looks good! I have few potential refactors but didn't see anything that looked broken or hard to maintain. I would cast my notes as: optional.

app/services/adobe_pdf/base.rb
ajkiessl: Ah okay.  I didn't know you could do that with `direnv`.  I wonder if there's a way to switch contexts depending on an ENV variable?  From what I understand in that issue, you still need to manually switch contexts.

app/services/adobe_pdf/base.rb
ajkiessl: Ah yes.  I will change this.

app/services/adobe_pdf/asset.rb
ajkiessl: Yeah, I wasn't sure what would be better.  They both require similar amounts of code, and result in similar memoization.  I think with the way I have it, the value won't be stored until it's needed, which probably makes the code infinitesimally more efficient.

spec/services/adobe_pdf/accessibility_checker_spec.rb
ajkiessl: Good idea.  I'll do this.

spec/services/adobe_pdf/accessibility_checker_spec.rb
ajkiessl: Same.

spec/services/adobe_pdf/accessibility_checker_spec.rb
ajkiessl: The request is mocked with VCR/Webmock here spec/fixtures/vcr_cassettes/AdobePdf_AccessibilityChecker/_call/happy_path/fetches_the_accessibility_report.yml, so it won't break unless that is turned off.  If we do turn that off and run a live test, it would be good to have it break so we can make changes.  I added an issue to further handle the format and storing of the report here #1637 .

app/services/adobe_pdf/accessibility_checker.rb
ajkiessl: Yeah, this one might be hard to clean up.

Smullz622 (dismissed):
I agree with Jesse's suggestions & think it looks great otherwise! Once this is merged, I'll make the adjustments needed for #1628

jlandiseigsti (approved):
LGTM!

PR #1628, opened 2024-10-23
Store & present accessibility check results

### LINKED ISSUE #1613: Storage and Presentation of Accessibility Check Results

Smullz622:
Fixes #1613 

Creates a new model to store the result of the accessibility checker

Adds columns for Accessibility Score and Accessibility Report to the Files table (also adds an abbreviated version of the files table to work settings). This introduces ActionCable to do the live updates as the checker processes. Decided to go that direction due to some roadblocks getting AJAX working, but also it's faster than AJAX as it executes as soon as the accessibility check result is updated instead of polling every X seconds. The tests do not have ActionCable configured at this time & instead do refreshes. Rails 7 introduces Rspec support for ActionCable, and we're about to upgrade so it's being added to a future ticket.


config/cable.yml
```diff
@@ -1,5 +1,8 @@
 development:
-  adapter: async
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
ajkiessl: We should probably pull this from here: https://github.com/psu-libraries/scholarsphere/blob/main/lib/scholarsphere/redis_config.rb .  And do the same in the "production" section below.  We don't store REDIS_URL in our environment variables.  We do store REDIS_HOST, and then the url is pieced together in that RedisConfig class.

app/models/file_resource.rb
```diff
@@ -19,6 +19,8 @@ class FileResource < ApplicationRecord
           required: false,
           dependent: :destroy
 
+  has_one :accessibility_check_result, dependent: :destroy
```
ajkiessl: I put a `required: false` in here, but I'm realizing that it's not needed.  So if you encounter this as a merge conflict, it can be removed.

ajkiessl (changes requested):
This is looking good.  Interesting to see the ActionCable stuff implemented.  Just a few small things on initial review.  

Also, I think we should deploy this as a preview before merging it into main.  So we can work anything out with the deployment before going to QA.

ajkiessl (approved):
Looks good.  Let's see if we can get the preview working.  If not we can just merge to main and check it out in QA.

config/cable.yml
```diff
@@ -1,10 +1,13 @@
...
+  channel_prefix: scholarsphere_development
+  password: <%= ENV.fetch("REDIS_PASSWORD", 'redispassword') %>
 
 test:
   adapter: test
 
 production:
   adapter: redis
-  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
+  url: <%= Scholarsphere::RedisConfig.new.to_hash['url'] %>
```
ajkiessl: I made this change in the deployment branch to get the connection to redis working:

Scholarsphere::RedisConfig.new.to_hash[:url]

ajkiessl (changes requested):
There was just one thing I changed in the deployment branch to get it working.  I mention it below.  Otherwise, looks good!  https://scholarsphere-acc-status.dsrd.libraries.psu.edu/

app/models/accessibility_check_result.rb
```diff
@@ -3,5 +3,101 @@
...
+  after_commit :broadcast_to_file_version_memberships
+
   validates :detailed_report, presence: true
+
+  def score
+    "#{num_passed} out of #{num_total} passed" if detailed_report.key?('Detailed Report')
+  end
+
+  def failures_present?
+    num_passed != num_total
```
ajkiessl: I noticed this bug while testing things out in the preview.  When there is an error stored in `.detailed_report`, this raises an error: https://app.bugsnag.com/penn-state/scholarsphere/errors/6725322fb73b2f0a1437c1ee?filters[event.unhandled]=true&filters[event.since]=30d&filters[app.release_stage]=acc-status .

I think we need to check `if detailed_report.key?('Detailed Report')`.

ajkiessl (changes requested):
One more thing I noticed.

ajkiessl (approved):
lgtm!  Thanks!

PR #1641, opened 2024-10-30
Customizing contributors step of deposit workflow for instruments

EricDurante:
I'd like to do more work on this if I have time, but I'm pushing up what I have currently in case I don't.

In terms of UI behavior, I think that this is correct/complete based on what we've discussed, and it should at least convey the gist of how I think that this should work. At the very least, it needs some additional unit tests for some of the new code as it's currently written. Those are missing because I'm not quite settled on the design of the new form objects that I've introduced.

There were a handful of failing tests in the base branch when I started, and those that were directly related to the new instruments pathway are fixed now, but there are a few tests that use VCR that are still failing for me. I'm not sure why they were failing to begin with since they don't seem related to this work, and I haven't looked into it.

In terms of the overall design, I'm also not yet quite happy with how we're handling data validations in general. Right now we have a mixture of validations on the form objects that validate data in the context of the deposit UI steps, and also validations on the ActiveRecord model that validate data in other contexts such as depositing works via the API. This has the effect of layering validations on top of other validations that are sometimes redundant, and it forces us to incorporate conditional logic in some of the validations. Some validations are common/shared, but they're all context-specific to some extent. In the long run, I think that it might be better to remove most or all of the validation logic from the `WorkVersion` ActiveRecord model and validate input from the API using another, separate form/model object that may share some validation logic with the existing UI forms. However, that's a larger refactoring that doesn't specifically have to do with this new deposit pathway.


ajkiessl:
@EricDurante No problem if you don't have time to finish this.  We'll revisit the instruments stuff in the next ScholarSphere cycle.

EricDurante:
Thanks, @ajkiessl. I likely won't have time to finish, but I'm hoping that I can at least push up a little bit more work on it while it's still fresh in my mind.

ajkiessl:
The 'spec/features/dashboard/work_version_form_spec.rb' instruments and culminating experiences tests had some difficulties merging, but I think it came out still testing what we want it to.

There may be other weird things in here resulting from merge conflicts.

closes #1691

app/forms/work_deposit_pathway.rb
```diff
@@ -363,15 +409,36 @@ def self.form_fields
...
+            contributor
+          }.freeze
+        end
+
+        form_fields.each do |attr_name|
+          delegate attr_name, to: :work_version, prefix: false
+          delegate "#{attr_name}=", to: :work_version, prefix: false
+        end
+
+        def build_creator(*)
```
Smullz622, line 433: I don't totally get what's going on with this method being empty. Could you please explain?

ajkiessl: Since the `@resource` variable in the contributors controller could be a `Collection`, `ContributorsFormBase`, or `Instrument::ContributorForm` and just calls `.build_creator`, it seemed like the best way to handle the polymorphism without adding logic to the controller.  The `Instrument::ContributorForm` doesn't build creators, so when it calls the method, it doesn't do anything.  It does build creators in the other two instances.

ajkiessl: I updated the comment to make this more clear.

ajkiessl: But I think we still want an associated authorship (creator) for these.  Rethinking this.

ajkiessl: Ah never mind the Work will still have the depositor association

Smullz622 (approved):
lgtm!

PR #1654, opened 2024-11-20
Rails 7 upgrade

### LINKED ISSUE #1634: Scholarsphere upgrade to rails 7.2
Part of #1633

Smullz622:
Fixes #1634 

upgrade from Rails 6.1 to Rails 7.2


config/environments/production.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# require "active_support/core_ext/integer/time"
```
jlandiseigsti: Do we need this?

Smullz622: Nope, good catch. I think that was a remnant from troubleshooting that I missed

lib/omniauth/strategies/azure_oauth.rb
```diff
@@ -37,7 +37,7 @@ def callback_url
       end
 
       def raw_info
-        @raw_info ||= JSON.parse(Base64.decode64(access_token['id_token'].split('.')[1]))
+        @raw_info ||= JSON.parse(Base64.decode64(access_token.token.split('.')[1]))
```
ajkiessl: I forget why this was changed, but this seems to prevent the groups from being imported.  `.token` must not contain all the info from the returned token.  Can this be changed back?

ajkiessl (changes requested):
This is looking good.  There's one thing mentioned below I think we're going to have to dig into further:

lib/omniauth/strategies/azure_oauth.rb
Smullz622: That change was made because tests were failing since `access_token['id_token']` was nil. All of the token info was saved in token instead. This will take some figuring out because when trying to put a byebug here, it only hit in the test environment, never in local dev. If you're seeing errors though, that would indicate the nil is a test problem not an overall problem.

Smullz622: Scratch that, I tried a fresh login from a private browser & finally was able to hit the byebug in dev. Looks like token and id_token both return some form of valid data (both let me login in general). Some of the data is the same in both, but there are differences including that id_token data has groups & token does not.

ajkiessl (approved):
LGTM thanks!

PR #1673, opened 2025-02-14
WIP: Ruby and Node Upgrades

ajkiessl:
I was able to get Uppy working with these upgrades.  Turns out the problem with the uploader was not Uppy, but the aws sdk upgrades.  An env variable (`AWS_SDK_CONFIG_OPT_OUT`) now needs to be set to `true` to prevent the aws sdk from trying to use your systems aws config.  Given that the Uppy packages are getting old, I added an issue to address that: https://github.com/psu-libraries/scholarsphere/issues/1688

Other than that, the majority of the changes here are just linter changes.


Smullz622 (approved):
LGTM!

Curious on the controller method reorders, was that a rubocop thing?

ajkiessl:
@Smullz622 Yeah, must've changed with the rubocop updates in niftany.  Looks like there's now a linter that enforces the preferred rails controller action order: https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/ActionOrder .

PR #1685, opened 2025-03-05
Prevent non psu affiliated accounts from logging in

ajkiessl:
Don't let non PSU affiliated accounts sign in to Scholarsphere.  They will still go through omniauth but not log into the app.


jlandiseigsti (approved):
👍

jrpatterson (approved):

PR #1696, opened 2025-04-04
Block publication while accessibility checker is running

### LINKED ISSUE #1625: Prevent publish during deposit workflow if accessibility checks are in progress
The publish button should be disabled until all accessibility checks have completed for uploaded pdfs.  We'll need some help text to explain to users why the button is disabled and how long it might take.

Smullz622:
Fixes #1625 

I moved `allow_publish?` into a concern so that it could be used in the broadcast method as well. From what I read, it's not good practice to call a concern directly which is why I added the service class, but as always, I'm open to suggestion if that's not actually the case.


ajkiessl (approved):
LGTM

PR #1697, opened 2025-04-07
1624 accessibility report to lib answers

### LINKED ISSUE #1624: Add accessibility check report to libanswers export
The libanswers ticket creation export for accessibility review is already done here: #1621 .  Once the accessibility checker is complete (#1602), we can send the report over with it.

jlandiseigsti:
closes #1624

### DIFF
--- a/spec/request/admin/curation_ticket_spec.rb
+++ b/spec/request/admin/curation_ticket_spec.rb
@@ -23,12 +23,37 @@ RSpec.describe 'Create Curation Ticket', type: :request do
 
       before do
         allow(LibanswersApiService).to receive(:new).and_return libanswer_response
-        allow(libanswer_response).to receive(:admin_create_curation_ticket).and_return('Redirect Path')
       end
 
-      specify do
-        post admin_create_curation_ticket_url(work), params: { ticket_type: 'curation' }
-        expect(response).to redirect_to('Redirect Path')
+      context 'when the ticket type is curation' do
+        before do
+          allow(libanswer_response).to receive(:admin_create_curation_ticket).with(work.id.to_s).and_return('Redirect Path')
+          post admin_create_curation_ticket_url(work), params: { ticket_type: 'curation' }
+        end
+
+        it 'redirects to the given path' do
+          expect(response).to redirect_to('Redirect Path')
+        end
+
+        it 'calls the LibanswerApiService #create_curation_ticket' do
+          expect(libanswer_response).to have_received(:admin_create_curation_ticket).with(work.id.to_s)
+        end
+      end
+
+      context 'when the ticket type is other than curation' do
+        before do
+          allow(libanswer_response).to receive(:admin_create_accessibility_ticket).with(work.id.to_s,
+                                                                                        'http://localhost:3000').and_return('Another Redirect Path')
+          post admin_create_curation_ticket_url(work), params: { ticket_type: 'other' }
+        end
+
+        it 'redirects to the returned path' do
+          expect(response).to redirect_to('Another Redirect Path')
+        end
+
+        it 'calls the LibanswerApiService #create_accessibility_ticket' do
+          expect(libanswer_response).to have_received(:admin_create_accessibility_ticket).with(work.id.to_s, request.base_url)
+        end
       end
     end
   end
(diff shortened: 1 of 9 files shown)


spec/request/admin/curation_ticket_spec.rb
```diff
@@ -27,7 +27,7 @@
       end
 
       specify do
-        post admin_create_curation_ticket_url(work)
+        post admin_create_curation_ticket_url(work), params: { ticket_type: 'curation' }
         expect(response).to redirect_to('Redirect Path')
```
ajkiessl: We should test the new logic here when a the ticket type is not 'curation'.  To make it more precise, we should also probably be testing the args passed into `admin_create_curation_ticket` or `admin_create_accessibility_ticket` using `allow(libanswer_response).to receive(...).with(...)` .

spec/services/libanswers_api_service_spec.rb
```diff
@@ -27,33 +27,73 @@
...
+  describe '#admin_create_accessibility_ticket' do
+    let!(:user) { create(:user, access_id: 'test', email: 'test@psu.edu') }
+    let!(:work) { create(:work, depositor: user.actor) }
+    let!(:base_url) { 'www.example.com' }
+
+    context 'when successful response is returned from libanswers /ticket/create endpoint' do
+      it 'returns the libanswers ticket url' do
+        expect(described_class.new.admin_create_accessibility_ticket(work.id, base_url)).to eq 'https://psu.libanswers.com/admin/ticket?qid=15802188'
+      end
+    end
```
ajkiessl, line 65: We should test when there is a report link, too.  I believe this, and the test below, just tests when there isn't a report.

spec/services/libanswers_api_service_spec.rb
```diff
@@ -27,33 +27,73 @@
...
+
+      before do
+        allow(Faraday).to receive(:new).and_return mock_faraday_connection
+        allow(mock_faraday_connection).to receive(:post).and_return dummy_response
+      end
 
-      it 'uses id 2590 for accessibility types' do
-        described_class.new.admin_create_curation_ticket('accessibility', work.id)
+      it 'uses id 5477 for quid and ScholarSphere Deposit Curation for question' do
+        described_class.new.admin_create_accessibility_ticket(work.id, base_url)
```
ajkiessl: This test looks similar to the test above (line 61), but I may not be understanding.  Also, this appears to be doing an accessibility ticket, and is using queue number 2590.

ajkiessl (changes requested):
Just a couple things I noticed with the tests.

spec/services/libanswers_api_service_spec.rb
jlandiseigsti: Yes, there are tests that are very similar between the `create_curation_ticket` and `create_accessibility_ticket` methods, as the methods themselves have a lot in common. I went back and forth on whether they should be separate methods at all – they didn't use to be – but ultimately decided they were beginning to diverge enough that it felt better organized to have to. I do need to change the name of that test so that the queue number and question are correct.

ajkiessl: I was referring to the test on line 61 within the `create_accessibility_ticket` tests.  I think it ultimately does things very similarly, just that test uses VCR and this one uses webmock to mock the request + response.  

I like the idea of separating the curation and accessibility methods.  I think you could handle some of the duplicate code by doing something a little more polymorphic.  Something like:

```
class TicketCreator
  attr_reader :work

  def initialize(work)
    @work = work
  end

  def create!
    raise NotImplementedError
  end

  protected

  def create_connection
    Faraday.new(...) 
  end

  def handle_response(response)
    # logic
  end
end
```
```
class CurationTicketCreator < TicketCreator
  def create!
    subject = "ScholarSphere Deposit Curation: #{work.latest_version.title}"
    conn = create_connection
    response = conn.post(create_ticket_path, "quid=#{SCHOLARSPHERE_QUEUE_ID}&" +
                                             "pquestion=#{subject}&" +
                                             "pname=#{work.display_name}&" +
                                             "pemail=#{work.email}")
    handle_response(response)
  end
end
```

And something similar for the accessibility ticket creator.  Not necessary but could be a good refactoring.

ajkiessl (approved):
LGTM

spec/services/libanswers_api_service_spec.rb
jlandiseigsti: Good idea! I'll do that

PR #1700, opened 2025-04-14
allow admins to edit publisher doi when metadata imported from RMD

### LINKED ISSUE #1674: Enable Admin to edit publisher DOI for scholarly works deposited through ScholarSphere
# Bug Report

## Description
Currently, Admin cannot edit publisher DOI for scholarly works. This was set up because we don't want users to edit the publisher DOI brought in from RMD. But for those deposited in ScholarSphere, there can be user errors that we need to correct as curators.

Smullz622:
Fixes #1674


jlandiseigsti (approved):
LGTM!

ajkiessl (approved):

PR #1702, opened 2025-04-16
remove pathway specific metadata when changing work type to a new pathway

### LINKED ISSUE #1676: Remove culminating experience metadata fields when work type is changed
# Bug Report

## Description
Currently, when changing work type from culminating experience to other, the culminating experience fields remain in the metadata.

Smullz622:
Fixes #1676


app/controllers/dashboard/form/work_version_type_controller.rb
```diff
@@ -23,8 +23,15 @@ def create
...
         @resource = WorkVersion.includes(:work).find(params[:id])
+        original_work_type = @resource.work_type
         authorize(@resource)
         @resource.attributes = work_version_params
+        if original_work_type != @resource.work_type
+          fields_to_reset = WorkDepositPathway.new(@resource).fields_to_reset(original_work_type)
+          fields_to_reset.each do |field|
+            @resource.send(:"#{field}=", nil)
+          end
+        end
```
ajkiessl: I'm wondering if we could pull this logic out and put it in the WorkVersion class?

ajkiessl: Like a `.clear_irrelevent_fields` method

ajkiessl: You might be able to move this entirely out of the controller with ActiveRecord methods/callbacks:

```
# app/models/work_version.rb
class WorkVersion < ApplicationRecord
  before_save :reset_irrelevant_fields_if_work_type_changed

  private

  def reset_irrelevant_fields_if_work_type_changed
    return unless work.will_save_change_to_work_type?

    fields_to_reset = WorkDepositPathway.new(self).fields_to_reset(work.work_type_before_last_save || work.work_type_was)
    fields_to_reset.each { |field| self[field] = nil }
  end
end
```

I _think_ this will work.

app/forms/work_deposit_pathway.rb
```diff
@@ -72,6 +76,21 @@ def scholarly_works?
...
+        DataAndCode::DetailsForm.form_fields.union(DataAndCode::PublishForm.form_fields)
+      elsif Work::Types.grad_culminating_experiences.include?(work_type)
+        GradCulminatingExperiences::DetailsForm.form_fields.union(GradCulminatingExperiences::PublishForm.form_fields)
+        # will need to add instruments once that pathway is merged
+        # elsif Work::Types.instrument.include?(work_type)
+        # Instrument::DetailsForm.form_fields.union(Instrument::PublishForm.form_fields)
+      else
+        General::DetailsForm.form_fields.union(WorkVersionFormBase::COMMON_PUBLISH_FIELDS)
+      end
+    end
```
ajkiessl: Some metaprogramming might reduce some duplication here:

```
def pathway_fields(work_type)
    possible_namespaces = %w[ScholarlyWorks DataAndCode GradCulminatingExperiences]

  namespace = possible_namespaces.find do |ns|
    types_const = Object.const_get("Work::Types::#{ns.underscore}")
    types_const.include?(work_type)
  rescue NameError
    false
  end

  if namespace
    details_form = Object.const_get("#{namespace}::DetailsForm")
    publish_form = Object.const_get("#{namespace}::PublishForm")
    details_form.form_fields.union(publish_form.form_fields)
  else
    General::DetailsForm.form_fields.union(WorkVersionFormBase::COMMON_PUBLISH_FIELDS)
  end
end
```

Smullz622: I had trouble with this exact approach, but I was able to take a similar approach & reduce duplication

app/controllers/dashboard/form/work_version_type_controller.rb
Smullz622: This mostly worked, just had to change how the actual field was set to nil

Smullz622:
Made the suggested changes, also added logic for instruments pathway now that the pathway is merged into main. While doing so, noticed a few inconsistencies with instrument fields & fixed those

ajkiessl (approved):
LGTM

PR #1703, opened 2025-04-17
Delete orphaned FileResources

### LINKED ISSUE #792: Deleting works doesn't delete files
When a work is deleted, the corresponding files aren't deleted. They remain as database records and on s3). How should we do garbage collection for files?

ajkiessl:
closes #792 

Shrine is already set up to delete files in S3 (the file and it's derivatives) when a FileResource is destroyed.  However, since FileResources could have multiple associations with WorkVersions, there was no function to delete the FileResource once orphaned.  This PR fixes that and adds a rake task to cleanup existing orphaned FileResources.

### DIFF
--- a/app/models/accessibility_check_result.rb
+++ b/app/models/accessibility_check_result.rb
@@ -3,8 +3,8 @@
 class AccessibilityCheckResult < ApplicationRecord
   belongs_to :file_resource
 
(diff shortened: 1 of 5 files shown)


Smullz622 (approved):
lgtm!

PR #1708, opened 2025-04-22
Configure session timeout with devise after three days

ajkiessl:
This should force users to reauthenticate with azure after 8 hours so we don't have banned users running months-long sessions.

Note: this only ends the session if there is no activity from the user


jrpatterson (approved):
Let's try this and keep monitoring for account activity.

PR #1709, opened 2025-04-22
Add cascade to foreign key in db for accessibility check results 

ajkiessl:
Oddly, some files were not deleting in production because of a PG error:

`PG::ForeignKeyViolation (ERROR:  update or delete on table "file_resources" violates foreign key constraint "fk_rails_9139814472" on table "accessibility_check_results"
DETAIL:  Key (id)=(18755) is still referenced from table "accessibility_check_results".`

So, I added the cascade to the foreign key in the db with a migration.  Hopefully this works.  I can only test this in QA though.


jlandiseigsti (approved):
👍

PR #1710, opened 2025-04-23
Fix tests

EricDurante:
I was experiencing a lot of test failures locally due to Selenium errors while using Chrome and chromedriver, and I’m not experiencing the same failures while using Firefox and geckodriver.


EricDurante:
Awesome - CI is passing again. I think this is ready to merge.

ajkiessl (approved):
LGTM.  Thanks for fixing this!

PR #1717, opened 2025-05-06
Allow users to input alt text when uploading files

### LINKED ISSUE #1622: Explore options for image alt text
The best approach would be to use Uppy to display a method for applying alt text when an image file is uploaded. So, first step would be to determine if this is something that Uppy can do.

ajkiessl:
closes #1622 

- Requires users to input alt text when uploading an image through uppy
- Gives admins a form field to edit alt text from the Work Settings page
- Adds aria-label describing images on download links on the show page for a work

Some other things addressed that aren't exactly part of the issue:

- I modified the solr start script in the docker compose yml to use solr's `start` command.  This seemed to fixed the persistence issues for about a week, but then I had to switch branches and restart my computer and it stopped working again.  I'm not sure if it's because I switched branches and restarted solr (w/o the `start` command), or because my computer restarted and the persistence doesn't work.  Regardless, I don't think it hurts to use solr's `start` command here.  In fact it's probably a better way of doing it.

- I upgraded uppy to v3.  Not much changed from v1 to v3, but it we should still try to get it to v4 some time since v3 is still 2 or 3 years old.

This does not give API users the ability to upload alt text.  I added an issue here for that: #1747 – if that's something we want to do.


EricDurante (commented):
There's some stuff here that I'm not very familiar with, but this generally looks good to me.

A question that didn't occur to me earlier - there's no way for a depositor to edit/change the alt text for an image after they've submitted it and uploaded the image file, correct? If that's right, I'm a little concerned about people needing to fix typos or just deciding that they want change their wording, etc. prior to publishing. I don't think that they necessarily need the ability to edit existing alt text; if they want to change it, they could just remove and re-upload the image file and enter new text. However, it also doesn't look like there's a way for the depositor to see the alt text that was saved prior to publication. Is that right? If so, would there be any problem with displaying the alt text in the files table on the publish form? That would allow depositors to review the text that they entered and perhaps catch typos before publishing.

ajkiessl:
@EricDurante Correct, there's no way for users to edit the alt text after submission.  It would be helpful to display the alt text for users on the publish page for review.  Should be pretty easy to add too.  I'll add that.

PR #1721, opened 2025-05-08
First and Last Name Requirements for New Works

### LINKED ISSUE #1678: Make first and last names required (not just full name) for creators so citation gets pulled correctly

granta03:
Fixes #1678

### DIFF
--- a/app/forms/work_deposit_pathway.rb
+++ b/app/forms/work_deposit_pathway.rb
@@ -239,7 +239,7 @@ class WorkDepositPathway
         def validate_creator_names
           creators.each do |creator|
             if creator.given_name.blank? || creator.surname.blank?
-              errors.add(:creators, 'Each creator must have a given name and a surname.')
+              errors.add(:creators, I18n.t('dashboard.form.contributors.edit.incomplete_name'))
             end
           end
         end
--- a/app/forms/work_deposit_pathway.rb
+++ b/app/forms/work_deposit_pathway.rb
@@ -228,9 +228,21 @@ class WorkDepositPathway
                :contributor=,
                to: :work_version, prefix: false
 
+      validate :validate_creator_names
+
       def form_partial
         'non_instrument_work_version'
       end
+
+      private
+
+        def validate_creator_names
+          creators.each do |creator|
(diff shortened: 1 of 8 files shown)


granta03:
Validation is at the model level, which will also validate through the API. Is this what we want and if so, this makes the API not backwards compatible and we have to consider clients?

spec/forms/work_deposit_pathways/contributors_form_base_spec.rb
```diff
@@ -67,4 +67,32 @@
...
+        expect(form.errors[:creators]).to include('Each creator must have a given name and a surname.')
+      end
+    end
+
+    context 'when no given name' do
+      before { wv.creators << build(:authorship, { display_name: 'Creator 1', given_name: nil, surname: 'Doe', email: 'abc123@email.com' }) }
+
+      it 'returns an error and asks for a given name' do
+        expect(form).not_to be_valid
+        expect(form.errors[:creators]).to include('Each creator must have a given name and a surname.')
```
jlandiseigsti, line 86: I appreciate you checking for the error message as well as the validity.

jlandiseigsti (dismissed):
This all looks good to me!

app/forms/work_deposit_pathway.rb
```diff
@@ -228,9 +228,21 @@ class ContributorsFormBase < WorkVersionFormBase
...
       def form_partial
         'non_instrument_work_version'
       end
+
+      private
+
+        def validate_creator_names
+          creators.each do |creator|
+            if creator.given_name.blank? || creator.surname.blank?
+              errors.add(:creators, 'Each creator must have a given name and a surname.')
```
ajkiessl: I forgot to bring this back up while we were pairing;  We should store the text for the error message in the locales.  Probably somewhere in here: https://github.com/psu-libraries/scholarsphere/blob/main/config/locales/en.yml#L510-L511

ajkiessl (changes requested):
Just one thing.  Other than that this is good.

ajkiessl (approved):
lgtm

PR #1722, opened 2025-05-08
1639 accessibility changes

### LINKED ISSUE #1639: Accessibility Changes from 10/21/24
A route of testing with Michelle McManus on 10/21/24 revealed some things to improve for accessibility in the Scholarsphere Create Work workflow:

Details Page

- Description should be marked as required (it currently is required and the page cannot proceed without it, but the screen reader doesn't describe it as required)
- Publication date: same 
- Any other required fields unique to particular pathways
- "Submit" for the DOI section is confusing on a screen reader: should say "Submit DOI" or something similar.
- Error message that appeared did not pull focus.

Publish Page

- License should be marked as required for screen readers.
- When a required field is not filled out and publishing fails, should go to the blank field/checkbox.

jlandiseigsti:
closes #1639


ajkiessl (approved):
lgtm

PR #1723, opened 2025-05-12
Fix solr persistence issues

ajkiessl:
@granta03  and I noticed solr seemed to not persist locally when bringing down the solr container and bringing it back up.  I made these changes and tested bringing the containers down and back up _and_ shutting down orbstack and bringing the containers back up.  This seems to work, but there might be something more going on.


jlandiseigsti (approved):
I've noticed this as well – definitely down to try this fix!

ajkiessl:
@jlandiseigsti Just a heads up, you might need to delete your current solr volume, reinitialize solr, and reindex with this change.

PR #1724, opened 2025-05-13
1693 more a11y changes

### LINKED ISSUE #1693: Accessibility Manual and Automated Testing Issues Report
Here is the report that contains all errors/issues found through both manual and automated testing for ScholarSphere.

[ScholarSphere Testing Results.docx](https://github.com/user-attachments/files/19589445/ScholarSphere.Testing.Results.docx)

jlandiseigsti:
closes #1693


app/views/form_fields/_select.html.erb
```diff
@@ -5,12 +5,11 @@
 <div class="mt-1 form-group has-float-label mb-3">
   <%= form.select attribute,
                   options_for_select,
-                  { include_blank: local_assigns.fetch(:include_blank, false) },
+                  { include_blank: local_assigns[:include_blank].presence },
                   {
                     class: 'form-control custom-select',
-                    required: local_assigns[:required].presence,
```
ajkiessl, line 11: Do we want to remove this?  Does the "aria_required" do the same thing as the regular "require" where there's a popup client side?

jlandiseigsti: No "aria_required" doesn't do the client-side enforcement in the way that regular "require" does. In the case of the publish page, that's what we want...we want the enforcement to just be server-side, so that the modal will close and the message not be hidden.

I believe I removed this because it wasn't being used...that we didn't have any select fields that were passing in the "required" value. But I'm happy to add it back in so we have that option to make any of them required.

ajkiessl: Yeah let's add it back in, looks like the Grad Culminating Experiences select dropdown should be required.

ajkiessl (approved):

PR #1725, opened 2025-05-13
Do not allow users to add psu affiliated group to edit groups

ajkiessl:
Expand excluded groups in  editors form to include the psu affiliated groups.  Keep these groups from the groups options and adds validation to save method to prevent these groups from being added through the form


jlandiseigsti (approved):
LGTM!

PR #1726, opened 2025-05-13
Admin Override When Changing Work Type from non-DataAndCode to DataAndCode

### LINKED ISSUE #1677: Allow Admin to override the README validation when changing a work type to Dataset

granta03:
Fixes #1677 

Note: This PR adds paper trails to works.  This was the best way we could come up with to track if something had previously not been a data and code work type.


granta03:
Alex and I are working to move the `work_deposit_pathway_spec.rb` tests I made to the `publish_form_spec.rb` file once the CI tests pass since contexts are already defined there.

app/forms/work_deposit_pathway.rb
```diff
@@ -29,11 +29,11 @@ def contributors_form
     end
   end
 
-  def publish_form
+  def publish_form(current_user: nil)
     if scholarly_works?
       ScholarlyWorks::PublishForm.new(resource)
     elsif data_and_code?
-      DataAndCode::PublishForm.new(resource)
+      DataAndCode::PublishForm.new(resource, current_user: current_user)
```
jlandiseigsti: Think this could just be `current_user:`, as it is below.

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
```diff
@@ -256,7 +258,7 @@
         end
       end
 
-      context "when the form's work version is not published" do
+      context "when the form's work version is not published (should_validate_readme? is false)" do
```
jlandiseigsti: Context name is a little confusingly...should this be "when the form's work version is not published **and** should_validate_readme? is false"? Also, isn't the state of whether it's published now part of the should_validate_readme? method? Maybe just 'when should_validate_readme? is false', unless I'm missing something.

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
```diff
@@ -352,8 +354,10 @@
     context "when the form's work version is not otherwise valid for publishing" do
       let(:wv) { build(:work_version, :with_creators, description: nil, published_date: '2020') }
 
-      context "when the form's work version is published" do
-        before { wv.publish }
+      context "when the form's work version is published and should_validate_readme? is true" do
```
jlandiseigsti, line 357: Similar to above... `when should_validate_readme? is true', perhaps?

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
```diff
@@ -464,7 +468,7 @@
         end
       end
 
-      context "when the form's work version is not published" do
+      context "when the form's work version is not published (should_validate_readme? is false)" do
```
jlandiseigsti: Similar question to above, but also, this seems to be testing the same thing as the test that starts on line 261 is testing: that if should_validate_readme? is false, then it doesn't validate. Is there a distinction I'm not seeing?

jlandiseigsti (changes requested):
Looks good, just a few questions about the tests

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
ajkiessl: There are underlying validations within the WorkVersion model that are also ran here that are only dependent on the published status and not `should_validate_readme?`.  Since `should_validate_readme?` is always false when not published, I tried to indicate that with "(should_validate_readme? is false)", but it should probably be more descriptive.

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
ajkiessl: Like above, since the WorkVersion validations are dependent on the published status, this test should still indicate that the work version is published.  I could have added more complexity to this test and nested two more contexts "when should_validate_readme? is true" and "when should_validate_readme? is false", but I felt these were getting too complex and the tests I added below to specifically test `should_validate_readme?` cover that logic.

spec/forms/work_deposit_pathways/data_and_code/publish_form_spec.rb
ajkiessl: This one is actually different because it's nested within the context "when the form's work version is not otherwise valid for publishing" instead of "when the form's work version is otherwise valid for publishing" .

jlandiseigsti (approved):
👍

PR #1727, opened 2025-05-13
Prevent deletion of FileResource if there's an associated ThumbnailUpload

ajkiessl:
Since this set up for thumbnail uploads is not ideal, it's easier to just add a validation to the destroy callback to prevent a FileResource from being deleted if there's an associated ThumbnailUpload.

Ideally, ThumbnailUploads should just operate like a FileUploader and inherit all the Shrine functionality separate from FileResource.  Not sure why it was done this way, but I seem to remember there being a reason.  I just can't remember what.

### DIFF
--- a/spec/models/file_resource_spec.rb
+++ b/spec/models/file_resource_spec.rb
@@ -298,4 +298,25 @@ RSpec.describe FileResource do
       end
     end
   end
+
+  describe 'cannot_be_deleted_if_linked_to_thumbnail_upload' do
+    let!(:file_resource) { create(:file_resource) }
+
+    context 'when the file resource has a linked thumbnail upload' do
+      before do
+        create(:thumbnail_upload, file_resource: file_resource)
+      end
+
+      it 'does not allow the file resource to be deleted' do
+        expect { file_resource.destroy }.not_to change(described_class, :count)
+        expect(file_resource.errors[:base]).to include('FileResource cannot be deleted while associated with a ThumbnailUpload')
+      end
+    end
+
+    context 'when the file resource does not have a linked thumbnail upload' do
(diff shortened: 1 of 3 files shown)


ajkiessl:
Issue to make this more ideal: #1728

jlandiseigsti:
Do we anticipate #1728 to get done in this dev cycle?

(also sorry for the close and reopen; misclick)

spec/models/file_resource_spec.rb
```diff
@@ -298,4 +298,25 @@
...
+
+      it 'does not allow the file resource to be deleted' do
+        expect { file_resource.destroy }.not_to change(described_class, :count)
+        expect(file_resource.errors[:base]).to include('FileResource cannot be deleted while associated with a ThumbnailUpload')
+      end
+    end
+
+    context 'when the file resource does not have a linked thumbnail upload' do
+      it 'allows the file resource to be deleted' do
+        expect { file_resource.destroy }.to change(described_class, :count).by(-1)
```
jlandiseigsti, line 318: Could also check for errors to be empty, but I don't feel strongly about this.

jlandiseigsti (dismissed):
Looks good!

ajkiessl:
> Do we anticipate #1728 to get done in this dev cycle?
> 
> (also sorry for the close and reopen; misclick)

No, I just want to have an issue out there so we don't forget about it.  Even if it sits there... forever

spec/models/file_resource_spec.rb
ajkiessl: Ah yes I wanted to add that

jlandiseigsti (approved):
👍

PR #1736, opened 2025-05-20
Use file title not work_version title

ajkiessl:
Using `work_version` here displays the work version's title.  We want the file's title.

### DIFF
--- a/app/views/shared/_work_version_files.html.erb
+++ b/app/views/shared/_work_version_files.html.erb
@@ -11,7 +11,7 @@
                         title: t('resources.download', name: file.title),
                         class: 'matomo_download',
                         rel: 'nofollow' do %>
-              Download <%= work_version.title %>
+              Download <%= file.title %>
             <% end %>
           </h3>
           <p class="meta mb-0">


jlandiseigsti:
This change was actually a request from the a11y office in their report:
Screenshot 2025-05-20 at 2 27 14 PM

ajkiessl:
@jlandiseigsti There is likely a misunderstanding from the accessibility office.  Works are not "articles" with a single associated article to download.  Many of our works have numerous downloads of all types: images, pdfs, genomes, zip files, audio etc.  Displaying the work's title for every link is inconsistent with what the link is expected to convey – what file you are actually downloading. This arguably makes it more difficult for _all_ users to understand what is being downloaded.

ajkiessl:
To make this more accessible without removing the filename we could have an aria label for clear reading of the filename like:

"Download file: #{name}, file type: #{ext}"

Ultimately, even if you're using a screen reader, you still need to know the name of the file so you can find it once it's downloaded.

jlandiseigsti:
I like that solution; thanks for the clarification

jlandiseigsti (approved):
👍

PR #1739, opened 2025-05-22
1675 flexible time period reports

### LINKED ISSUE #1675: Enable users to get statistics reports for more than 1 month at a time
A user can see total downloads/views over multiple months or years.

EricDurante:
Closes #1675

We could make the UI for this slightly nicer if we think that it's worth it. We're not using a javascript date-picker anywhere else in this app yet, so I was hesitant to introduce one just for this. As is, I think it works fine with the built-in Rails date select helpers and a little bit of validation.


ajkiessl (approved):
LGTM

PR #1742, opened 2025-05-23
Send RMD uploads to airtable for curation

ajkiessl:
Was debating removing this:

https://github.com/psu-libraries/scholarsphere/blob/main/app/models/external_app.rb#L44-L48

But decided not to because:

1.  It broke a bunch of tests _and_
2. The statement `For all intensive purposes, external applications have admin rights` is still true.  External apps still have admin rights, they just aren't included in the "Don't send to airtable if published by us" logic


jlandiseigsti (dismissed):
Looks good to me!

Issue #1690, opened 2025-03-20
Scholarsphere Security patches


ajkiessl:

ajkiessl:
closed by #1754

PR #1754, opened 2025-06-03
Security vulnerability patches

ajkiessl:
This should cover everything ruby and node-wise _except_ bootstrap.  We are using both bootstrap 4 and 5.  v4 is EOL and has vulnerabilities.  When removing 4, the entire UI is blown up.  Issue for this here: https://github.com/psu-libraries/scholarsphere/issues/1755

I upgraded redoc-cli (actually migrated to redocly) and eslint in the PR as well.


ajkiessl:
closes #1690

jlandiseigsti (approved):
👍

PR #1757, opened 2025-06-10
Fix publish form bug for Grad Culminating Experiences

ajkiessl:
Move has_file_image_resource? up the inheritance chain so it gets delegated for all work types in all forms.  Adds test for confirmation


jlandiseigsti (approved):
LGTM

PR #1760, opened 2025-06-12
Allow contributors to be deleted when required fields are blank

ajkiessl:
Admins noticed an issue when deleting contributors that did not have givennames or surnames: they could not submit the form after deleting a contributor.  This was caused by `cocoon` (our frontend package that handles `has_many` associations forms) not removing the `required` attributes when marking records for destruction.

This PR adds a cocoon event to remove all required attributes from all fields in an "association" (ie contributors) form and turns off form validation for contributors if they are marked for deletion.


jlandiseigsti (approved):
Looks good to me!

PR #1765, opened 2025-07-07
Upgrade from Bootstrap 4 to 5

### LINKED ISSUE #1755: Remove bootstrap v4
v4 was EOL in 2023 and has vulnerabilities.  Only v5+ should be used. 

Lots of our styling is based off v4, so removing this breaks much of the UI.  This will take some time to convert.

granta03:
Fixes #1755

### DIFF
--- a/spec/services/rmd_publication_spec.rb
+++ b/spec/services/rmd_publication_spec.rb
@@ -4,46 +4,58 @@ require 'rails_helper'
 require 'support/vcr'
 
 RSpec.describe RmdPublication, :vcr do
-  let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
-
   context 'when publication is found' do
     describe '#title' do
+      let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
+
       it 'returns the title value for the received publication' do
         expect(rmd_publication.title).to eq 'A Scholarly Research Article'
       end
     end
 
     describe '#secondary_title' do
+      let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
+
       it 'returns the secondary_title value for the received publication' do
         expect(rmd_publication.secondary_title).to eq 'A Comparative Analysis'
       end
     end
 
     describe '#abstract' do
+      let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
+
       it 'returns the abstract value for the received publication' do
         expect(rmd_publication.abstract).to eq 'A summary of the research'
       end
     end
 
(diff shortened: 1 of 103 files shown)


jlandiseigsti (changes requested):
A few things I've noticed while testing:
1) In the header, the Nav Links aren't left-aligned with the PSU icon
2) The Administrator drop down menu is not working for me
3) The alerts can no longer be closed by clicking the x icon
4) On the Publish page, the "Access" heading is missing
5) The "Penn State only" option is misaligned.

I haven't gotten to look at every page yet; I'm having trouble publishing a work or collecting. This isn't related to the bootstrap issue though, since I'm seeing the problem locally in main as well. I'll check other pages once I figure out what's up with my local environment.

jlandiseigsti:
> A few things I've noticed while testing:
> 
>     1. In the header, the Nav Links aren't left-aligned with the PSU icon
> 
>     2. The Administrator drop down menu is not working for me
> 
>     3. The alerts can no longer be closed by clicking the x icon
> 
>     4. On the Publish page, the "Access" heading is missing
> 
>     5. The "Penn State only" option is misaligned.
> 
> 
> I haven't gotten to look at every page yet; I'm having trouble publishing a work or collecting. This isn't related to the bootstrap issue though, since I'm seeing the problem locally in main as well. I'll check other pages once I figure out what's up with my local environment.

Disregard – this is a mistake on my part. When I checked out out this branch to view the changes, I neglected to do a yarn install. Doing one corrected all these issues. Sorry about that!

spec/services/rmd_publication_spec.rb
```diff
@@ -4,46 +4,58 @@
 require 'support/vcr'
 
 RSpec.describe RmdPublication, :vcr do
-  let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
-
   context 'when publication is found' do
     describe '#title' do
+      let(:rmd_publication) { described_class.new('https://doi.org/10.1038/abcdefg1234567') }
```
jlandiseigsti, line 9: Curious why it was necessary to declare this for each test rather than just once?

jlandiseigsti (approved):
Looks good! One thing I had a question about, but it's not blocking. Nice work with this!

spec/services/rmd_publication_spec.rb
granta03: I remember that when I was running the tests with Alex some of them kept failing with just the one declaration but they passed when declaring them for each test. I'm not sure why they would.

jlandiseigsti: Probably something wasn't getting disposed of properly at the end of the test. Not a big deal, just curious.

PR #1767, opened 2025-07-11
1766 libanswer ticket change

### LINKED ISSUE #1766: Change behavior when creating an accessibility libanswers ticket when email is deactivated
In ticket #1680 , we made it so Libanswers tickets would get blocked if the submission belonged to a student who was no longer associated with the university, with the message "Depositor access no longer available."

However, the admin team _does_ want to create those tickets in some instances; they recognize that email is deactivated and the author won't respond but they still want to create a ticket in order to send to the accessibility team.

Instead of blocking the creation of the libanswers ticket, "it would be helpful to change it so we can still create a related ticket but we are also aware that the depositor no longer has access and won't respond."

This should only change for the accessibility type of tickets.

jlandiseigsti:
closes #1766

### DIFF
--- a/app/models/actor.rb
+++ b/app/models/actor.rb
@@ -95,6 +95,17 @@ class Actor < ApplicationRecord
     Collection.reindex_all(relation: created_collections)
   end
 
+  # Whether an actor/depositor is still an active part of the PSU community
+  # Staff, students, etc. will have their affiliation listed in the PsuIdentity.identity array
+  # But those with only "MEMBER" are those that are not PSU affiliated.
+  # This could be incoming students, outgoing staff, people with those "Friends of Penn State" accounts
+  def active?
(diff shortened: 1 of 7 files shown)


app/models/actor.rb
```diff
@@ -95,6 +95,17 @@ def update_index(_options = {})
     Collection.reindex_all(relation: created_collections)
   end
 
+  # Whether an actor/depositor is still an active part of the PSU community
+  # Staff, students, etc. will have their affiliation listed in the PsuIdentity.identity array
```
ajkiessl: Should this say `PsuIdentity.affiliation`?

jlandiseigsti: Yep, it should. Good catch!

ajkiessl (approved):
This all LGTM.  Thanks for doing this!

PR #1768, opened 2025-07-16
Bump resources in policy.xml

ajkiessl:
We had some thumbnail generation jobs fail due to resources limitations.  The limits set in the policy.xml were _very_ conservative, so I bumped these up a decent amount.


jlandiseigsti (dismissed):
👍

bryanlopezrr (approved):

PR #1785, opened 2025-08-19
Bug Fix: Creators' given and family names not updating in the "review and publish" step

### LINKED ISSUE #1776: Creators' given and family names not updating in the "review and publish" step
You can publish after making changes to creators' given and family names on the "review and publish" step, but it does not save the changes.

We want these changes to save to the database.

granta03:
Fixes #1776

### DIFF
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -148,6 +148,9 @@ module Dashboard
                 :actor_id,
                 :_destroy,
                 :display_name,
+                :given_name,
+                :surname,
+                :email,
                 actor_attributes: [
                   :id,
                   :email,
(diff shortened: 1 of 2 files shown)


granta03:
Another bug I found related to this is that the Access Account field still shows the old user's id after publishing. Looking at the table for authorships it doesn't look like the database updates the `actor_id`. Looking into this further.
Screenshot 2025-08-19 at 3 20 29 PM

granta03:
> Another bug I found related to this is that the Access Account field still shows the old user's id after publishing. Looking at the table for authorships it doesn't look like the database updates the `actor_id`. Looking into this further. 

Determined that this issue is more for people editing small changes (ex: Alex to Alexander) rather than replacing the full user. This can be ignored for now.

ajkiessl (changes requested):
Is there somewhere in the test suite we can add a test to make sure these values get passed through the publish controller?

ajkiessl (approved):
LGTM thanks!

PR #1792, opened 2025-09-02
added cassette for libanswers test

### LINKED ISSUE #1791: Spec Test Uses Live LibAnswers Call
In `spec/features/dashboard/work_settings_spec.rb`, one of the spec tests doesn't have a VCR cassette. It uses live API calls to LibAnswers. It would be preferred to mock this call but the test will need additional fixing too.

granta03:
FIxes #1791


granta03:
CodeClimate has been commented out because its API has been disabled and needs to be migrated to Qlty. I created a new issue about it here #1793

jlandiseigsti (approved):
LGTM!

PR #1802, opened 2025-09-09
Nanoid Security Vulnerability Fix

granta03:
Fixes nanoid security vulnerability by upgrading version 4.x.x to 5.0.9. Version 3.x.x doesn't have this vulnerability and can stay to not break node packages that use nanoid 3.


ajkiessl (approved):
LGTM

PR #1803, opened 2025-09-10
add fields needed for auto remediation workflow

### LINKED ISSUE #1801: Data model additions for pdf remediation workflow
We need a way to identify which versions are a product of automated remediation to prevent subsequent remediation of auto-remediated uploads.  We can add a boolean (maybe `auto_remediated_version`) field to `WorkVersion` to achieve this.  Then, set this to `true` when the new version is created.

We also need to store the uuid for the remediation job with the file being remediated.  We should add a string field to `FileResource` to store this.  This can be stored when the remediation is triggered, and then used to retrieve the `Work` we need to add a new version too when the remediation is complete.  It will also allow us to identify which files need removed from the new remediated `WorkVersion`.

### WDDLL

- [x] Boolean field added to `WorkVersion` to indicate if the work version is an auto-remediated version
- [x] String field added to `FileResource` to store the job UUID returned from the pdf_accessibility_api

Smullz622:
Fixes #1801

### DIFF
--- a/db/migrate/20250910123610_add_auto_remediated_version_to_work_version.rb
+++ b/db/migrate/20250910123610_add_auto_remediated_version_to_work_version.rb
@@ -1,0 +1,5 @@
+class AddAutoRemediatedVersionToWorkVersion < ActiveRecord::Migration[7.2]
+  def change
+    add_column :work_versions, :auto_remediated_version, :boolean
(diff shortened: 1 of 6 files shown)


ajkiessl (approved):
LGTM

PR #1809, opened 2025-09-17
trigger auto remediation

### LINKED ISSUE #1800: Trigger remediation
When a user clicks the download link for a file, we will trigger the remediation process _if_:

- The `WorkVersion` is the most recent published version
- The `WorkVersion` is not already remediated or in the process of being remediated
- The `WorkVersion` is not itself a remediated version
- The file being downloaded is a PDF
- The user is not the depositor, author, or an admin

To properly track if a `WorkVersion` is being remediated, we should add a `remediation_started_at` timestamp to  `WorkVersion` and store that time when a download triggers the remediation process.  This will help to prevent multiple remediation processes from starting at the same time.  Also, I think it's best to store this with the `WorkVersion` instead of the `FileResource`, since the `FileResource` will get a job UUID anyway (indicating that it has been remediated), and since files can be attached to multiple versions, this will better identify which version was remediated.

The remediation process will need triggered for all PDFs associated with the `WorkVersion`.  Each remediation process should be a background job using SideKiq.  Each job UUID returned from the pdf_accessibility_api should be stored with each corresponding `FileResource` (PDF).

Smullz622:
Fixes #1800

### DIFF
--- a/app/jobs/auto_remediation_job.rb
+++ b/app/jobs/auto_remediation_job.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'pdf_remediation/client'
+
 class AutoRemediationJob < ApplicationJob
   queue_as :auto_remediation
 
--- a/app/jobs/auto_remediation_job.rb
+++ b/app/jobs/auto_remediation_job.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class AutoRemediationJob < ApplicationJob
+  queue_as :auto_remediation
+
+  def perform(file_resource_id)
+    file_resource = FileResource.find(file_resource_id)
+
+    begin
+      remediation_job_uuid = PdfRemediation::Client.new(file_resource.file_url).request_remediation
+      file_resource.update(remediation_job_uuid: remediation_job_uuid)
+    rescue StandardError => e
+      Rails.logger.error("Failed to auto remediate #{file_resource.id}: #{e.message}")
+      raise e
+    end
+  end
+end
--- a/app/services/auto_remediate_service.rb
+++ b/app/services/auto_remediate_service.rb
@@ -11,7 +11,7 @@ class AutoRemediateService
 
   def call
     work_version.update(auto_remediation_started_at: Time.current)
-    pdfs = work_version.file_resources.where("file_resources.file_data->'metadata'->>'mime_type' = ?", 'application/pdf')
+    pdfs = work_version.file_resources.is_pdf
     pdfs.each do |pdf|
       AutoRemediationJob.perform_later(pdf.id) if pdf.remediation_job_uuid.blank?
     end
--- a/app/services/auto_remediate_service.rb
+++ b/app/services/auto_remediate_service.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class AutoRemediateService
+  attr_reader :work_version, :admin, :download_is_pdf
+
+  def initialize(work_version_id, is_admin, download_is_pdf)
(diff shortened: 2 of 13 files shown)


app/jobs/auto_remediation_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
ajkiessl, line 1: To get this to work for me locally, I needed to add:
```
require 'pdf_remediation/client'
```

ajkiessl (changes requested):
I managed to get this whole thing working end-to-end entirely in docker locally.  I don't recommend doing it this way, though.  It was not easy.  In the future I'll probably just hook ScholarSphere up to either the QA or Prod instance of the pdf_accessibility_api and take the hit on the Adobe tokens.  There was one issue I encountered which I mention below.  So far everything else is looking good, but I haven't gotten through the tests yet.  I wanted to get this review in before the weekend, though.

db/schema.rb
```diff
@@ -275,6 +275,7 @@
     t.boolean "mint_doi_requested"
     t.boolean "accessibility_remediation_requested"
     t.boolean "auto_remediated_version"
+    t.datetime "remediation_started_at"
```
ajkiessl: Given our discussion in kickoff about displaying something onscreen while the new version is being manually remediated, I'm thinking we might want to tweak the name of this field to make it more specific to the auto-remediation process.  Maybe `auto_remediation_started_at`?

app/services/auto_remediate_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(work_version_id, is_admin, download_is_pdf)
+    @work_version = WorkVersion.find(work_version_id)
+    @admin = is_admin
+    @download_is_pdf = download_is_pdf
+  end
+
+  def call
+    work_version.update(auto_remediation_started_at: Time.current)
+    pdfs = work_version.file_resources.where("file_resources.file_data->'metadata'->>'mime_type' = ?", 'application/pdf')
```
ajkiessl: Not sure if this refactoring would work, but could you make this a scope in `FileResource`?  And use that scope in the existing `:needs_accessibility_check` scope.  Then, we'd have the mime type defined in the `FileResource` class and it removes the mime type filtering duplication.

spec/jobs/auto_remediation_job_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    context 'when an error occurs' do
+      let(:error_message) { 'Something went wrong' }
+
+      before do
+        allow(client).to receive(:request_remediation).and_raise(StandardError, error_message)
+      end
+
+      it 'creates or updates the AccessibilityCheckResult with the error message and raises the error' do
```
ajkiessl: I think the `AccessibilityCheckResult` part of this needs removed.

ajkiessl (changes requested):
I spotted a potential refactoring and a minor test naming issue.  Everything else looks good!

app/services/auto_remediate_service.rb
ajkiessl: I was thinking something like this in 'FileResource':

```
scope :is_pdf, -> {
    where("file_data->'metadata'->>'mime_type' = ?, PDF_MIME_TYPE)
   }

scope :needs_accessibility_check, -> {
    left_joins(:accessibility_check_result)
    .is_pdf
    .where("accessibility_check_results.id IS NULL")
  }
```

ajkiessl (approved):
LGTM

PR #1815, opened 2025-09-29
remediation triggered popup notification

### LINKED ISSUE #1808: Remediation process started notification
After the user clicks the download link and the remediation process is triggered, a pop up should appear on the screen notifying the user that an accessible version is being created.  It should also inform the user that this could take up to an hour to complete and to return to this work later to access the new version.  The notice should also indicate that AI will be used to generate alt text for images embedded in the PDF.

Dependent on #1800

Smullz622:
Fixes #1808 

side note- looks like the trouble I had with handling `helpers.current_user` in the test isn't new (similar examples [here](https://github.com/psu-libraries/scholarsphere/blob/d160a038ef586a62c0eb740d01f1b8975b134cd9/spec/components/collection_metadata_component_spec.rb#L9) and [here](https://github.com/psu-libraries/scholarsphere/blob/d160a038ef586a62c0eb740d01f1b8975b134cd9/spec/components/files_visibility_detail_component_spec.rb#L15)). What's there now is obviously working, but might be something to create a ticket for if we ever spend some time on a refactor/interns are looking for something to do

### DIFF
--- a/app/views/shared/_remediation_popup.html.erb
+++ b/app/views/shared/_remediation_popup.html.erb
@@ -2,17 +2,16 @@
   <div class="modal-dialog modal-lg" role="document">
     <div class="modal-content">
       <div class="modal-body">
-        <p><strong>Accessible Version in Progress</strong></p>
-        <p>We’re generating an accessible version of this file to meet ADA Title II requirements. This process may take up to one hour. Please return later to access the accessible copy once it’s ready.</p>
+        <p><strong><%= t('remediation_popup.title') %></strong></p>
+        <p><%= t('remediation_popup.description') %></p>
         <ul>
-          <li><strong>What’s happening:</strong> An accessible PDF is generated using Adobe with AI used to generate alternative text (alt text) for images in the PDF.</li>
-          <li><strong>Quality check:</strong> The Libraries’ Adaptive Technology and Services team will review and correct the accessible version manually, but that review may occur after the file becomes available.</li>
+          <li><strong><%= t('remediation_popup.whats_happening_label') %></strong> <%= t('remediation_popup.whats_happening_text') %></li>
+          <li><strong><%= t('remediation_popup.quality_check_label') %></strong> <%= t('remediation_popup.quality_check_text') %></li>
         </ul>
         <br>
-        <p>For questions or to request a different format, contact the Adaptive Technology and Services team at <a href="mailto:ul-adaptive-tech-svcs@lists.psu.edu">ul-adaptive-tech-svcs@lists.psu.edu</a>.</p>
-      </div>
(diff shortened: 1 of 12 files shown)


app/views/shared/_remediation_popup.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <div class="modal-body">
+        <p><strong>Accessible Version in Progress</strong></p>
+        <p>We’re generating an accessible version of this file to meet ADA Title II requirements. This process may take up to one hour. Please return later to access the accessible copy once it’s ready.</p>
+        <ul>
+          <li><strong>What’s happening:</strong> An accessible PDF is generated using Adobe with AI used to generate alternative text (alt text) for images in the PDF.</li>
+          <li><strong>Quality check:</strong> The Libraries’ Adaptive Technology and Services team will review and correct the accessible version manually, but that review may occur after the file becomes available.</li>
+        </ul>
+        <br>
+        <p>For questions or to request a different format, contact the Adaptive Technology and Services team at <a href="mailto:ul-adaptive-tech-svcs@lists.psu.edu">ul-adaptive-tech-svcs@lists.psu.edu</a>.</p>
+      </div>
```
ajkiessl: We should move the text here to the en.yml so it can be localized.

ajkiessl (changes requested):
This looks great.  I think for readability,  we should lower the pop up a bit more to be closer to center and increase the width of the box.  Maybe more like this:

Screenshot 2025-09-30 at 2 10 44 PM

Also, if it's possible, I think we should have a feature test to test clicking the download link and then confirming the modal.

ajkiessl:
This LGTM.  Not sure why those tests are failing in CI, they work for me locally.  I'll look into it further if I get some time.

spec/features/resources_spec.rb
```diff
@@ -114,6 +114,87 @@
...
+              current_count = ViewStatistic.where(
+                resource_type: 'FileResource',
+                resource_id: file.file_resource.id
+              ).count
+
+              break if current_count == initial_count + 1
+
+              sleep(0.1)
+            end
+          end
```
ajkiessl, line 190: @Smullz622 Was the issue that there was a delay after clicking 'Download' for the ViewStatistic count to update?

Smullz622: Yeah, it seemed like that was contributing. When everything was contained within the expect block, it was still failing with a sleep I think because of how those blocks are evaluated. It finally worked when I took it out of the expect block and I set up the timeout loop to poll if it had enough time to process vs just putting in a `sleep(10)`

ajkiessl (approved):
LGTM

PR #1819, opened 2025-10-06
Build Auto Remediated Work Version Webhook

### LINKED ISSUE #1773: pdf_accessibility_api webhook endpoint
This will receive the following json payload from the pdf_accessibility_api:

```
{
 uuid: "uuid",
 status: "status",
 output_url: "output_url"
}
```
The output_url being the presigned_url for downloading the remediated file.

The endpoint will need some basic authentication to check the authentication token sent from the pdf_accessibility_api.  We can simply store this as an environment variable.  The token will be passed along in the 'X-API-Key' header.

There are some things we will need to consider when creating the new work versions:

1.) What if a user started a draft before remediation finishes?  Is it possible to just publish a new version and leave the draft?  Then the user can return to the draft.  _Or_ do we move the version number of the draft to the most recent version after publishing the remediated version?

2.) Do we publish immediately after creation of the remediated version? Or do we want to wait until all the files for the new version are remediated before publishing (so keep it as a draft until all PDFs are remediated)?  There could be multiple files being remediated for a single version.  Is it possible to have more than one draft?

There should be a method somewhere to build a new version from the previous version (since our workflow already does this), so it should be easy to use that to create the new version.  Then, we just need the endpoint process to replace the old files with the new remediated files.  

So, generally the process will go like this: when a file comes back from remediation, we can use the job UUID to identify the `FileResource` and the most recent published version it is associated with.  Then, if that version is not a remediated version, create a new `WorkVersion`; tag it as a remediated version (via the new boolean from #1801); and replace the `FileResource` with the new remediated file.  If the the most recent published version is already tagged as a remediated version, then just replace the identified file with the new remediated file - no reason to create a new version.

ajkiessl:
closes #1773


ajkiessl:
Preview: #1824

PR #1821, opened 2025-10-08
add admin under accessibility review checkbox

### LINKED ISSUE #1817: Allow admins to set a Work as "Under accessibility review"
Works with more complicated documents (big files, non-OCR'd files) will require manual accessibility review.  These works require more time before becoming accessible.  We want admins to be able set a work as "Under manual accessibility review".

In the "Work Settings" page we can add a toggle button or checkbox that only admins can see that indicates the work is under review by the accessibility team.  We will need a boolean field added to the works table for this.

While this boolean is `true`, there should be a notice or info box on the work's view page indicating that it is under review (for accessibility) and that the most recent published version's documents may not be fully accessible.  We should get more precise language from stakeholders.

Also, while this boolean is `true`, the trigger for auto accessibility remediation should be prevented.

Smullz622:
Fixes #1817 

@bdezray I put the following placeholder text for what we want to display on the work page while under manual review. Please edit as needed:
This work is currently under manual accessibility review and may not be fully accessible at this time. An accessible version is being generated and will be available soon.

### DIFF
--- a/app/views/dashboard/shared/_manual_review_form.html.erb
+++ b/app/views/dashboard/shared/_manual_review_form.html.erb
@@ -1,0 +1,23 @@
+<p><%= t('.explanation') %></p>
+
+<%- manual_review_form_id = "edit_manual_review_#{dom_id manual_review_form.resource}" %>
+<%= form_for manual_review_form,
+             url: url,
+             method: :patch,
+             remote: false,
+             html: { id: manual_review_form_id, class: 'edit-resource-manual-review' } do |form| %>
+
+  <%= hidden_field_tag 'form_id', manual_review_form_id %>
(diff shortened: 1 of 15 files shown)


app/views/dashboard/shared/_manual_review_form.html.erb
```diff
@@ -0,0 +1,23 @@
+<p><%= t('.explanation') %></p>
```
jlandiseigsti, line 1: I didn't know about this way of writing which translation to use, that's cool!

jlandiseigsti (approved):
Looks great! Very clean, comprehensive tests.

ajkiessl (approved):

PR #1824, opened 2025-10-13
Build Auto Remediated Work Version Webhook

### LINKED ISSUE #1773: pdf_accessibility_api webhook endpoint

ajkiessl:
closes #1773

This implements the webhook logic for receiving a file from the PDF Accessibilty API and replacing the non-accessible file with that file.  It will create a new version of the work.  If there is more than one file awaiting remediation, the process will create a draft version and replace each file until all files are replaced.  Then, it will publish the version.  

Also, this PR adds the PdfAccessibilityApi ExternalApp.  This ExternalApp is stored in the PaperTrail versioning as "whodunnit" when an accessible version is being created so the versioning at the bottom of the work show page indicates the PdfAccessibilityApi made the changes.  Also, this ExternalApp db record is linked to work versions it creates.

Some things we still need to figure out: #1828 + #1827 

preview: https://scholarsphere-pdf-webhook.dsrd.libraries.psu.edu/

Note: the preview sends live requests to the remediation app, so it does use tokens.


app/jobs/build_auto_remediated_work_version_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class BuildAutoRemediatedWorkVersionJob < ApplicationJob
+  queue_as :auto_remediation
+
+  def perform(remediation_job_uuid, remediated_file_url)
+    # In the extremely rare case that multiple remediation jobs are kicked off
+    # for the same file, the older jobs should fail here since the
+    # remediation_job_uuid will no longer be associated with the file.
```
jlandiseigsti, line 9: Appreciate the comment here.

jlandiseigsti (approved):
Looks good! Seems like you thought through all the edge cases.

PR #1830, opened 2025-10-17
Fix flaky test

ajkiessl:
This seems to have fixed the flaky test - or at least it passed 4 times in CI so I'm pretty confident.  Finding the dropdown and clicking it makes the following `select` statement more robust.  It seemed like the `select` statement was trying to select an option in the dropdown before the dropdown actually dropped.


jlandiseigsti (approved):
👍

PR #1833, opened 2025-10-21
Filter PDFs that are already auto remediated from auto remediation trigger

### LINKED ISSUE #1816: Filter `FileResources` out of remediation if already remediated
While I was working on #1773, I realized we'll need to consider how to handle work versions created by users _after_ a remediated version is created.  So, if someone creates a new version, how do we know which `FileResource`s are already remediated?  I plan to add an `auto_remediated_version` boolean to `FileResource` in #1773 and store that as `true` when the remediated file is stored with ScholarSphere.  However, we'll need to add some filtering during the remediation trigger to:

1.) Not trigger remediation if a user clicks the download link on an already remediated PDF
2.) Not trigger remediation on PDFs associated with the work version that remediation has been triggered for

This will prevent remediated PDFs from being remediated again, _and_ still allow new PDFs to be remediated.

Depends on #1773

ajkiessl:
closes #1816

### DIFF
--- a/app/components/file_download_link_component.rb
+++ b/app/components/file_download_link_component.rb
@@ -59,8 +59,8 @@ class FileDownloadLinkComponent < ViewComponent::Base
       @file_version_membership.file_resource.image?
     end
 
-    def pdf?
-      @file_version_membership.file_resource.pdf?
+    def can_remediate?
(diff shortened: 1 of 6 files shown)


Smullz622 (approved):
LGTM

PR #1836, opened 2025-10-27
Separate auto_remediation queues

### LINKED ISSUE #1826: Auto remediation SideKiq processes/queues
Should the auto remediation jobs run in their own process?  Given these jobs involve pushing and pulling files around the internet, I'm concerned they'll clog up our "main" process that already has like 9 queues.  Also, there are two of this "main" process, both are single threaded.  I'm not too sure why it is this way.

ajkiessl:
closes #1826 

Still need to get these queues and processes configured in the helm charts

### DIFF
--- a/app/jobs/auto_remediation_job.rb
+++ b/app/jobs/auto_remediation_job.rb
@@ -3,7 +3,7 @@
 require 'pdf_remediation/client'
 
 class AutoRemediationJob < ApplicationJob
-  queue_as :auto_remediation
+  queue_as :auto_remediation_out
 
   def perform(file_resource_id)
     file_resource = FileResource.find(file_resource_id)
--- a/app/jobs/build_auto_remediated_work_version_job.rb
+++ b/app/jobs/build_auto_remediated_work_version_job.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class BuildAutoRemediatedWorkVersionJob < ApplicationJob
-  queue_as :auto_remediation
+  queue_as :auto_remediation_in
 
   def perform(remediation_job_uuid, remediated_file_url)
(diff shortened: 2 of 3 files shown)


ajkiessl:
@bryanlopezrr `auto_remediation_out` and `auto_remediation_in` are the two Sidekiq queues we are adding.  We'd like these to run with the main process that already has like 9 queues, but we'd like to bump the threads to 5.

Or maybe we just add the two queues, leave the main process running with two separate replicas, and wait to see if we need to make changes.  I'm open to suggestions.  One of the these queues will be running very simple and quick jobs.  They won't need a lot of time or processing.  The other might take up a little more resources and time.  It downloads a file, stores that file, and adds a handful of records to the database.  Neither of these jobs are particularly time sensitive, though.

bryanlopezrr:
At the moment, the manifest has this:
```
          reservedWorkers:
            - name: thumbnails
              replicas: 1
              concurrency: 1
              queues:
                - thumbnails
          queues:
            - default
            - doi
            - indexing
            - mailers
            - shrine
            - statistics
            - signature
            - accessibility_check
            - webhooks
 ```
            
link: https://github.com/psu-libraries/scholarsphere-config/blob/9111c0bb9fa81f9a2475d2acf673e9ad5bd1e232/clusters/dev/manifests/scholarsphere/main.yaml#L47

I can add those two new queues to the manifest in dev and then increase the replicas to 2 with a higher concurrency of 5. To be honest, the benefit I see to bumping up values would be performance improvements

ajkiessl:
@bryanlopezrr Sounds good, but maybe just bump the concurrency to 2.  Since we'll have 2 replicas running.  That will give us a little room for concurrency in the same process, but not too much.  Then when we're ready we can add it to prod.

PR #1837, opened 2025-10-27
Fetches params in mailer so an error will raise if a param is missing

ajkiessl:
I noticed while testing in QA that the auto remediation email notification was not getting any "to" emails.  Digging through the code I realized I was passing the wrong param key to set up the "to" emails.  I added `.fetch`es to grab the params in the mailer class so an error raises if params are missing.


jlandiseigsti (approved):
looks good to me

PR #1839, opened 2025-10-28
1828 accessible file naming

### LINKED ISSUE #1828: PDF Accesssibility API incoming file naming
We don't want the remediated files named: "COMPLIANT_jsah897q3bo7391b938fg....".  We should replace the "COMPLIANT" and random ID with "ACCESSIBLE_VERSION".  In other words: prepend "ACCESSIBLE_VERSION_" to the original filename.

jlandiseigsti:
closes #1828


app/services/build_auto_remediated_work_version.rb
```diff
@@ -30,26 +30,26 @@ def self.call(file_resource, remediated_file_url)
...
       end
     end
   end
 
-  def self.on_publish(built_work_version)
-    built_work_version.publish!
-    lib_answers = LibanswersApiService.new
-    lib_answers.admin_create_ticket(built_work_version.work.id, 'work_remediation')
-    AutoRemediationNotifications.new(built_work_version).send_notifications
+  def self.build_file_resource(work_version, replacement_tempfile, original_filename)
```
ajkiessl: It's probably best to keep this `private` since it's only used in this class.

ajkiessl (changes requested):
Looks good, just one suggestion.

ajkiessl (dismissed):
LGTM

ajkiessl (approved):

PR #1843, opened 2025-10-30
Remediation failure handling

### LINKED ISSUE #1840: Auto remediation failure handling
When we receive a failed job response from the PDF Accessibility API (`job.failed`) we should create a ticket in LibAnswers for the work, and set the work to "Under Manual Review".  We send anything that was successful to LibAnswers anyway, and we don't currently have a way to rerun anything that fails, so this seems like the logical thing to do.  Also, we still need to remediate it, so this will let the Accessibility Team know they need to do it manually.

We should have something in the LibAnswers ticket that let's them know this one failed.

ajkiessl:
closes #1840


jlandiseigsti (approved):
LGTM!

PR #1844, opened 2025-10-31
Prevent auto remediation in production

### LINKED ISSUE #1841: Prevent automated remediation features in production
We still need to finalize our contract with Adobe, _so..._

We don't want the automated accessibility remediation features running in production, but we still want to release our changes.  So, we'll need to make sure the remediation process is not triggered and the popup on download does not show.  

I think the simplest way to handle this without breaking a lot of tests and preventing us from at least being able to use the new features locally, would be to wrap the triggers for the remediation process and popup in a `unless Rails.env.production?` block.

We might be able to package this logic up in the `able_to_auto_remediate?` function in the `AutoRemediateService` class.

ajkiessl:
Check rails environment in #able_to_auto_remediate?.  If `production` -> return false.  This turns off the remediation features (both the trigger and popup) in production.

closes #1841

### DIFF
--- a/app/services/auto_remediate_service.rb
+++ b/app/services/auto_remediate_service.rb
@@ -18,6 +18,10 @@ class AutoRemediateService
   end
 
   def able_to_auto_remediate?
+    # The following line can be removed when
+    # auto-remediation is enabled in production
+    return false if Rails.env.production?
+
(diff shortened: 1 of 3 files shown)


jlandiseigsti (approved):
👍

PR #1845, opened 2025-10-31
Revert "Prevent auto remediation in production"

### LINKED ISSUE #1846: Turn Auto Remediation On
The plan is to do this sometime early in the Spring semester.

ajkiessl:
Reverts psu-libraries/scholarsphere#1844

Leaving this here for when we want to turn the auto remediation feature on.  We can deploy a preview of this when we approach that date.

closes #1846


ajkiessl:
Converting to feature flag instead

PR #1852, opened 2025-11-21
Ignore Bot Requests to /resources/{uuid} That Trigger Noise & Bugsnag Alerts

### LINKED ISSUE #1851: Ignore Bot Requests to /resources/{uuid} That Trigger Noise & Bugsnag Alerts
We’re receiving frequent bot/scanner requests to URLs in the format:

```
/resources/{uuid}
```

When the UUID does not correspond to an existing record, the application logs errors which get reported to Bugsnag. This creates noise and increases our Bugsnag event volume (which affects costs).

We should implement safe handling for these requests so they do not log errors or notify Bugsnag. 

### Potential solutions:

- Return a simple 404 without raising an exception
   Ensure controller/model lookups fail silently or in a controlled way.
   I think this is probably the most robust way to handle this.

- Add logic to mark these as “ignored” errors in Bugsnag
   using Bugsnag.before_notify or some other modification in the Bugsnag config.

- Identify bots and block them via cloudflare/bot challenge.

granta03:
Fixes #1851


config/initializers/bugsnag.rb
```diff
@@ -5,4 +5,5 @@
   config.release_stage = ENV.fetch('BUGSNAG_RELEASE_STAGE', 'development')
 
   config.discard_classes << 'ActionController::UnknownFormat'
+  config.discard_classes << 'ActiveRecord::RecordNotFound'
```
ajkiessl: I'm not certain this is a safe way to handle this.  We can handle `ActionController::UnknownFormat` like this because that error only raises when a user sends us junk requests.  There are many ways we could get a `ActiveRecord::RecordNotFound` that aren't when a user sends us a junk request.  I think we need to find a way to isolate this error to when it comes from a junk request.  So, maybe some logic in here to identify the request path, or controller.  Or, have something in the controller to ignore these.

ajkiessl (changes requested):

granta03:
Merge done due to having issues at the Bundle stage of CircleCI
Edit: this did not help. I changed the CircleCI config to what it suggested in the CI output.

ajkiessl (approved):

Issue #1848, opened 2025-11-07
Flag files that have already been remediated


binkylush:
After files are uploaded to SS, they go to the curation team for review. The curation team then passes them on to the Accessibility Team for manual remediation. The Accessibility team remediates them and then sends them back to the curation team to upload into SS. 

We need to flag all of the files that have been manually remediated. 

The curation team adds "AccessibleCopy_" to the front of every file that has been remediated so far.

ajkiessl:
There will still be instances where manual review is necessary even after autoremediation is turned on (ie failures, file too large, etc), ~so we should add a field to the Settings screen for admins to flag a version as remediated.~

Edit: Actually, this won't be possible from the Settings page.  It needs to be flagged per version.  So, we'll need something in the upload workflow that only admins can see to do this.

This is along with creating a script to flag remediated versions with a file that has "AccessibleCopy_" in the filename.

PR #1872, opened 2026-02-03
Autoremediation release

### LINKED ISSUE #1846: Turn Auto Remediation On

ajkiessl:
closes #1846 

Separating this from #1848 , since it's not necessary to at least get this running in QA. 

I removed the webhook token environment variable, and changed the webhook auth to use the token generated by `ExternalApp`.  Since Scholarsphere is managing those tokens, I think this is a better way of doing it.  However, the naming is a little confusing, so I came up with a temporary non-ideal method for grabbing webhook tokens.


spec/models/external_app_spec.rb
```diff
@@ -37,6 +37,7 @@
     subject(:app) { described_class.pdf_accessibility_api }
 
     it { is_expected.to be_a(described_class) }
+    its(:token) { is_expected.to eq(app.api_tokens.first.token) }
```
jlandiseigsti, line 42: Maybe overkill but we could add a similar test for `webhook_token`

docker-compose.yml
```diff
@@ -26,10 +26,10 @@ x-web_env: &web_env
       SOLR_COLLECTION: scholarsphere
       POSTGRES_HOST: db
       AZURE_GRAPH_GROUPS: ${AZURE_GRAPH_GROUPS}
+      ADMIN_GROUP: ${ADMIN_GROUP}
```
jlandiseigsti, line 29: This Env Var isn't used for anything related to pdf-remediation, right?

jlandiseigsti (changes requested):
Just one minor thing.

spec/models/external_app_spec.rb
ajkiessl: Oh yes, I'll add that.

docker-compose.yml
ajkiessl: Right.  It just makes local setup easier, since you can configure the admin group to something already seeded in the db.  I'm not sure why I didn't add it earlier.

jlandiseigsti (approved):
👍

PR #1875, opened 2026-02-06
Handles Tempfile uploads in "page_count" shrine step

ajkiessl:
Grabbing the page counts was failing when storing the auto remediated files because we pass Shrine a Tempfile, not a cached S3 location (like the Uppy upload does).  I fixed this and added a shrine plugin for determining mime types (backup in case no content type is specified in downloaded files).  Adds tests for determining page counts.


jlandiseigsti (approved):
Looks like a good fix

PR #1879, opened 2026-02-10
Directory restructure for PDF remediation

ajkiessl:
This is just a refactoring.  I moved classes for PDF remediation into their own directories, so:

- app/jobs/pdf_remediation/**
- app/services/pdf_remediation/**

And all the necessary changes in the tests and implementations.


jlandiseigsti (approved):
Looks good! Thanks for doing the organization.

PR #1880, opened 2026-02-10
Flag manually remediated work versions/files

### LINKED ISSUE #1848: Flag files that have already been remediated

ajkiessl:
closes #1848 

The most notable thing in this PR is that I changed "auto_remediation" vernacular in a lot of places to just "remediation".  Originally, I assumed almost everything would be done via auto-remediation, but since we're factoring in a decent amount of manual remediation, some places needed a wording change to make sense.  A lot of this PR is the wording changes.

Other than that, there is a rake task for flagging existing remediated work versions and files as remediated, and a checkbox on the publish screen for admins to flag uploads as remediated.


lib/tasks/flag_existing_remediated_files_and_work_versions.rake
```diff
@@ -0,0 +1,21 @@
...
+  task flag_existing_remediated_files_and_work_versions: :environment do
+    scope = FileResource.is_pdf.where("file_data->'metadata'->>'filename' LIKE ?", 'AccessibleCopy_%')
+
+    puts "Found #{scope.count} FileResource records with AccessibleCopy_ filename prefix"
+
+    scope.find_each do |file_resource|
+      file_resource.update(remediated_version: true)
+
+      latest_work_version = file_resource.work_versions.order(:version_number).last ||
+        file_resource.work_versions.order(:created_at).last
```
jlandiseigsti, line 14: Related to my previous comment...should a WorkVersion be marked as remediated if it contains a remediated file resource, or if _all_ the file resources attached are remediated? I guess it comes down to what we are using work_version.remediated_version to indicate.

app/models/work_version.rb
```diff
@@ -398,7 +398,13 @@ def has_large_pdf_file_resource?
   end
 
   def has_remaining_auto_remediation_jobs?
-    file_resources.exists?(['remediation_job_uuid IS NOT NULL AND auto_remediated_version = ?', false])
+    file_resources.exists?(['remediation_job_uuid IS NOT NULL AND remediated_version = ?', false])
+  end
+
+  def mirror_remediated_version_to_files!
+    file_resources.is_pdf.find_each do |file_resource|
+      file_resource.update!(remediated_version: remediated_version)
```
jlandiseigsti, line 406: Just want to clarify something...since a work version can have multiple file resources, I could see a situation where one file is remediated and one is not. Do the admins know that marking a work version as accessible will mark all the files within as accessible?

jlandiseigsti (commented):
Two clarifying questions before approving:

app/models/work_version.rb
ajkiessl: I don't think there's a better way to handle this.  You could display a per-file checkbox for remediation, but that would get overlooked in the UI and make things more complicated and less intuitive.  Admins (and users in general) approach things on the WorkVersion level (mostly).  I would expect an admin doing remediation to remediate the whole WorkVersion, and the files associated to the remediated version to act like remediated files from there-on.

ajkiessl: We can let admins know, but we should also observe how they react to it.  If they start reporting that they wish they could fine-tune which files are flagged, then we can consider changing it.

lib/tasks/flag_existing_remediated_files_and_work_versions.rake
ajkiessl: Like my other comment, admins should be remediating all files associated with a WorkVersion.  They then prepend "AccessibleCopy_" to them.  It probably would be _fine_ to just flag all associated files as remediated, but the expectation would be that they get flagged anyway.  If there really is an associated file without "AccessibleCopy_", then that's not really a big deal.  It could trigger an auto-remediation down the road, and then we have a sort of "6 in one, half dozen in the other" situation where we got a file remediated that needed to be remediated, but the other is now auto-remediated after being manually remediated.  So, maybe that's more of a "5 in one, half dozen in the other" because the end result actually involves more accessibility coverage.

jlandiseigsti (approved):
👍

PR #1887, opened 2026-02-17
Use update_columns so callbacks are skipped

ajkiessl:
We don't want solr reindexing on all of these, it's not needed


Smullz622 (approved):
lgtm

PR #1889, opened 2026-02-18
Hotfix - reindex work versions after thumbnail generation

ajkiessl:
Since thumbnail generation happens at the same time the work version is created in the accessibility remediation pathway, the work versions were not being reindexed after thumbnail creation.  This was preventing thumbnails from showing in the catalog index/list pages.

### DIFF
--- a/app/jobs/shrine/thumbnail_job.rb
+++ b/app/jobs/shrine/thumbnail_job.rb
@@ -9,9 +9,16 @@ class Shrine::ThumbnailJob < ApplicationJob
       attacher.create_derivatives :thumbnail
       record.save
     end
-    # If the created record is a thumbnail uploaded by a user,
-    # then the associated resource needs to be reindexed
-    record.thumbnail_upload.resource.update_index if record.thumbnail_upload.present?
+
+    if record.thumbnail_upload.present?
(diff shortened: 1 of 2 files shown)


app/jobs/shrine/thumbnail_job.rb
```diff
@@ -9,9 +9,16 @@ def perform(record)
...
       record.save
     end
-    # If the created record is a thumbnail uploaded by a user,
-    # then the associated resource needs to be reindexed
-    record.thumbnail_upload.resource.update_index if record.thumbnail_upload.present?
+
+    if record.thumbnail_upload.present?
+      # If the created record is a thumbnail uploaded by a user,
+      # then the associated resource needs to be reindexed
+      record.thumbnail_upload.resource.update_index if record.thumbnail_upload.present?
```
jlandiseigsti: You can remove this second if record.thumbnail_upload.present?

jlandiseigsti (changes requested):
Just one change.

app/jobs/shrine/thumbnail_job.rb
ajkiessl: Oh yes good catch

jlandiseigsti (approved):
👍

PR #1894, opened 2026-02-25
Adds clarifying line to remediation popup

### LINKED ISSUE #1893: Remediation popup wording fix
Indicate that the file downloaded after accepting the popup will be the original (non-accessible) version

ajkiessl:
Also, adds healthcheck to solr in docker compose setup to stop the annoying race condition where the web pod fails because solr isn't totally up.

closes #1893


jlandiseigsti (approved):
👍

PR #1895, opened 2026-02-26
Depositors and creators cannot trigger auto remediation

ajkiessl:
Adds logic to `able_to_auto_remediate?` to return false if current user is the depositor or a creator.  Adds tests for this.

Part of #1800


jlandiseigsti (approved):
👍

PR #1912, opened 2026-03-19
ScholarSphere hotfixes

ajkiessl:
There's a couple things in this PR:

- app/controllers/webhooks/pdf_accessibility_api_controller.rb: the strong params were not being used here, so I changed the param call to use strong params
- I realized that an admin could upload a new work version _after_ a remediated version, and the `remediated_version` values for `FileResource`s could get reset if it's not flagged during upload that it's a remediated version.  So, I decided to add the `auto_remediated_version` field back to the `FileResource` data model.  That stays separate from `remediated_version`, so it's not affected by admins' manual remediation.
- I changed up how we define private methods here: app/services/pdf_remediation/build_auto_remediated_work_version.rb .  I think it's just a more idiomatic way of doing it when there are multiple private class methods.  Functionality is the same except that `auto_remediated_version` is set when the file is created (as a part of the above bullet point).

### DIFF
--- a/app/services/pdf_remediation/build_auto_remediated_work_version.rb
+++ b/app/services/pdf_remediation/build_auto_remediated_work_version.rb
@@ -46,7 +46,8 @@ class PdfRemediation::BuildAutoRemediatedWorkVersion
   private_class_method def self.build_file_resource(work_version, replacement_tempfile, original_filename)
     new_resource = work_version.file_resources.create!(
       file: replacement_tempfile,
-      remediated_version: true
+      remediated_version: true,
+      auto_remediated_version: true
     )
 
     new_resource.file_data['metadata']['filename'] = "ACCESSIBLE_VERSION_#{original_filename}"
--- a/app/services/pdf_remediation/build_auto_remediated_work_version.rb
+++ b/app/services/pdf_remediation/build_auto_remediated_work_version.rb
@@ -8,6 +8,7 @@ class PdfRemediation::BuildAutoRemediatedWorkVersion
   def self.call(file_resource, remediated_file_url)
     wv_being_remediated = file_resource.latest_remediation_work_version_candidate
     original_filename = file_resource.file_data['metadata']['filename']
+
     # In the rare case a new work version is published after the remediation
     # job started, we stop the creation of a new remediated version here.
     unless wv_being_remediated.latest_published_version?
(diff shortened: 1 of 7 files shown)


Smullz622 (approved):
lgtm

app/services/pdf_remediation/build_auto_remediated_work_version.rb
```diff
@@ -43,20 +44,25 @@ def self.call(file_resource, remediated_file_url)
...
+  class << self
+    private
 
-    new_resource.file_data['metadata']['filename'] = "ACCESSIBLE_VERSION_#{original_filename}"
-    new_resource.save!
-  end
+      def build_file_resource(work_version, replacement_tempfile, original_filename)
+        new_resource = work_version.file_resources.create!(
+          file: replacement_tempfile,
+          remediated_version: true,
```
jlandiseigsti, line 53: Just to make sure I'm understanding correctly...remediated_version just means it has been remediated in some way, either by humans or the automatic process?

jlandiseigsti: Although maybe not: I see in a test an example of where remediated_version is false, and autoremediated_version is true. So does remediated_version just refer to manual remediation? If so, should it be left false here?

ajkiessl: Kind of, but I guess it doesn't always work that way in practice.  For instance, an admin could upload a newer work version, that is not a remediated work version, and have this file still attached.  Then, the `remediated_version` on the `FileResource` would be set to `false`, but the `auto_remediated_version` would still be `true`.  I was more-so thinking that `remediated_version` on `FileResource` should always reflect the most recent `WorkVersion` it is associated to's `remediated_version` value, but there isn't anything explicitly indicating that in the code.  So, I'm not sure if that's the best way to approach it.

ajkiessl: >> there isn't anything explicitly indicating that in the code.

I should clarify that the code _does_ work this way, though.  It's just in two separate places.  Through the GUI, the `WorkVersion` and `FileResource`'s `remediated_version`s are made to mirror each other, and through the auto remediation webhook, they are made to mirror each other.

ajkiessl: We may have to consider fine-grained management of manually remediated files.  I figured that would be too tedious, but it might be necessary to cover every edge-case.

jlandiseigsti (approved):

app/services/pdf_remediation/build_auto_remediated_work_version.rb
ajkiessl: I added a ticket for this: #1916 .  We'll see how this works out in production.  I'd like to see how admins actually interact with it before making any changes.

PR #1913, opened 2026-03-20
bump dependencies for security fixes (aws-sdk-s3, preact, shakapacker)

karthikreddy-02:
This draft PR updates several dependencies to address security vulnerabilities identified in recent scans.

Changes Made
Ruby (Gemfile.lock): Updated aws-sdk-s3 from 1.217.0 to latest secure version (includes core updates to aws-sdk-core 3.244.0 and related gems).
JavaScript (package.json & [yarn.lock]):
Bumped preact from ^10.5.13 to 10.28.2.
Bumped shakapacker from 7.2.3 to 9.5.0.

Closes #1855, Closes #1856, Closes #1857, Closes #1864, Closes #1878, Closes #1881, Closes #1898, Closes #1899, Closes #1900, Closes #1903, Closes #1906, Closes #1914


karthikreddy-02:
Please take a look at this document before approving the changes:https://docs.google.com/document/d/1mFLNyNj7aqXfg-YLJcIq4P1A7IlTvG2VH86yJsOFoBk/edit?tab=t.0

Thanks!

db/schema.rb
```diff
@@ -155,8 +154,8 @@
     t.uuid "uuid", default: -> { "uuid_generate_v4()" }
     t.string "remediation_job_uuid"
     t.boolean "remediated_version", default: false, null: false
-    t.boolean "auto_remediated_version", default: false, null: false
     t.datetime "auto_remediation_failed_at"
+    t.boolean "auto_remediated_version", default: false, null: false
```
jlandiseigsti, line 158: In this case, I would recommend discarding schema changes, since you didn't make a migration or changes to models. These changes are likely the result of a slightly different local database configuration and don't need to be saved.

app/javascript/controllers/uppy_controller.js
```diff
@@ -1,8 +1,8 @@
 import { Controller } from 'stimulus'
 import Uppy from '@uppy/core'
-import AwsS3Multipart from '@uppy/aws-s3-multipart'
+import AwsS3 from '@uppy/aws-s3'
```
jlandiseigsti, line 3: When I run this branch locally and attempt to upload a file with uppy, it fails with the message 
 `Failed to upload test.pdf Could not POST /s3/multipart`

In routes.rb, line 29, there is the line `mount Shrine.uppy_s3_multipart(:cache) => '/s3/multipart'`. And in the file shrine.rb, it uses the plugin Shrine.plugin :uppy_s3_multipart (line 14). I suspect these will have to be changed to complete the transition from AwsS3Multipart to AwsS3.

.circleci/config.yml
```diff
@@ -318,6 +318,7 @@ jobs:
           yarn test
     - run:
         name: "RSpec"
+        no_output_timeout: 10m
```
jlandiseigsti, line 321: Do we still need this?

spec/features/dashboard/work_version_form_spec.rb
```diff
@@ -1002,20 +1002,15 @@
       it 'requires alt text to be submitted and works' do
         visit dashboard_form_files_path(work_version)
 
-        # Wait for Uppy to load
-        while page.has_no_selector?('.uppy-Dashboard-AddFiles')
-          sleep 0.1
-        end
+        expect(page).to have_css('.uppy-Dashboard-AddFiles', wait: 30)
```
jlandiseigsti, line 1005: Nice decision to get rid of the while loops. There's nothing worse than an entire test suite to be brought to a halt by a single element not appearing.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM harbor.k8s.libraries.psu.edu/library/ruby-3.4.9-node-22:20260317 AS base
+FROM harbor.k8s.libraries.psu.edu/library/ruby-3.4.1-node-22:20260320 AS base
```
jlandiseigsti: I had to change this to ruby-3.4.9 in order to get my docker image to build. As of Justin's last update, we're on 3.4.9 for this application.

package.json
```diff
@@ -11,11 +11,11 @@
     "@rails/actioncable": "^6.1.4",
     "@rails/activestorage": "^6.1.4",
     "@rails/ujs": "^6.1.4",
-    "@uppy/aws-s3-multipart": "^3.3.0",
+    "@uppy/aws-s3": "^4.0.0",
     "@uppy/core": "^4.4.1",
     "@uppy/dashboard": "^4.2.0",
     "autocomplete.js": "^0.38.0",
-    "axios": "^1.12.0",
+    "axios": "1.13.5",
```
jlandiseigsti: Unless there's a really specific reason not to, let's keep using the caret that makes the version a little more flexible.

From stack overflow: 

> ^version “Compatible with version”, will automatically update you to all future minor/patch versions that are backwards-compatible, without incrementing the major version. ^1.2.3 will use releases from 1.2.3 to < 2.0.0.

package.json
```diff
@@ -25,38 +25,47 @@
     "css-loader": "^6.8.1",
     "css-minimizer-webpack-plugin": "^5.0.1",
     "d3": "^7.8.2",
+    "dompurify": "3.3.3",
```
jlandiseigsti: I actually don't think you need to specify any of these dependency packages. I removed them all, deleted the yarn.lock, ran yarn install, and they were all up-to-date. In general, we don't want to add specific versions of the dependency packages; we want our essential packages to determine which versions of their dependencies they need. Sometimes a "parent" package will need to updated in order to update the child, but we generally shouldn't get too granular in spelling out these versions.

Gemfile
```diff
@@ -55,7 +55,7 @@ gem 'rexml'
 gem 'rsolr', '>= 1.0', '< 3'
 gem 'rubyzip', '~> 2.3.0'
 gem 'seedbank'
-gem 'shakapacker', '~> 7.1'
+gem 'shakapacker', '9.5.0'
```
jlandiseigsti: Let's keep the ~> so it can grab minor updates as it needs to.

jlandiseigsti (changes requested):
Appreciate your tenacity on this one! Unfortunately, Uppy is still causing problems. When I ran it locally, I couldn't get the new set-up to work.

Also, I believe we want 3.4.9 in the Dockerfile, and fewer changes in the package.json.

Let me know if you want any clarification on any of my comments

package.json
```diff
@@ -25,38 +25,47 @@
...
     "postcss-loader": "^7.3.3",
     "postcss-preset-env": "^9.3.0",
+    "preact": "10.28.2",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "sass": "^1.69.5",
     "sass-loader": "^16.0.5",
     "select2": "^4.0.13",
-    "shakapacker": "7.2.3",
+    "shakapacker": "9.5.0",
```
jlandiseigsti, line 46: This is a good exception to the "^" rule. The version of shakapacker have to be exactly the same, so lets have this one be precise.

app/javascript/controllers/uppy_controller.js
karthikreddy-02: is minio running when you test? And are you running inside Docker or with a local Rails server?

jlandiseigsti: I was running inside Docker, with minio running. Is there a different setup I should try?

PR #1920, opened 2026-03-31
Flag non PDF files as remediated

ajkiessl:
I spoke with the ScholarSphere admins and discovered that when they are doing manual accessibility remediation, they are uploading accessible copies for more than just PDFs.  So, this PR changes the existing script that backfills "remediated_version" records to update _any_ file with an "AccessibleCopy_" prefix.  I also changed the code that sets the "remediated_version" for files when an admin checks the checkbox on the publish page to set that field for _all_ files (not just PDFs).

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -402,7 +402,7 @@ class WorkVersion < ApplicationRecord
   end
 
   def mirror_remediated_version_to_files!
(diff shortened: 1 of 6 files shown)


jlandiseigsti (approved):
Looks good to me!

PR #1921, opened 2026-04-01
Record usage for webhook token on successful webhook authentication

ajkiessl:
I realized we were not recording usage of the APIToken on the webhook requests.  It's probably worth doing this since most of the functionality is already there.  So, if a request gets past auth now, the APIToken (in this case the "webhook token") updates it's last used timestamp.


jlandiseigsti (approved):
LGTM

PR #1923, opened 2026-04-06
Skip validations for some of the values set at remediation start

ajkiessl:
`update` was being used in a few locations, which was validating old records and failing.  Model validations have changed so much that we have a lot of legacy "invalid" data.  Validations only really need to happen when editing and publishing works, so we can skip validations here with `update_column`.


Smullz622 (approved):
LGTM

jlandiseigsti (commented):
LGTM

PR #1925, opened 2026-04-07
Adds bot challenge to downloads controller

ajkiessl:
Also, adds logic to bot challenge page initializer to not challenge logged in users.

It appears we do have crawlers on our downloads endpoints.  This did not show in matomo analytics.  It may be because these are very simple bot crawlers, and do not trigger JS.  If that's the case, then this PR will dramatically reduce traffic, or it will do nothing because these bots may be "good" bots.


jlandiseigsti (approved):
👍

PR #1927, opened 2026-04-07
Skip validations in build auto remediated work version

ajkiessl:
We have legacy data that is incomplete and fails validations, so we skip validations when creating an auto remediated work version.

Adds test for this.  Also, when publishing incomplete work versions, test that they remain as drafts after auto remediation.


Smullz622 (approved):
LGTM

PR #1932, opened 2026-04-08
Preview/dependencies bump

karthikreddy-02:
New PR with the latest docker image being used for preview branch


spec/support/faker.rb
```diff
@@ -3,12 +3,12 @@
 RSpec.configure do |config|
   config.before :suite do
     if ENV.key?('FAKER_SEED')
-      puts "Running tests with Faker seed #{ENV['FAKER_SEED']}"
+
       Faker::Config.random = Random.new(ENV['FAKER_SEED'].to_i)
     end
   end
 
   config.after :suite do
```
jlandiseigsti: If we're not doing anything after the suite, can we remove this block?

db/schema.rb
jlandiseigsti, line 1: These changes can be reverted.

jlandiseigsti (changes requested):
Just two tiny things! Almost there

spec/support/faker.rb
karthikreddy-02: This was a rubocop update, I can revert it and see if it breaks.

jlandiseigsti: I think we can just tell Rubocop to ignore this line if we want to keep the logging.

app/javascript/controllers/uppy_controller.js
```diff
@@ -1,8 +1,8 @@
 import { Controller } from 'stimulus'
 import Uppy from '@uppy/core'
-import AwsS3Multipart from '@uppy/aws-s3-multipart'
+import AwsS3 from '@uppy/aws-s3'
 import Dashboard from '@uppy/dashboard'
-import { generateUploadedFileData, simulateEditAndUpload } from './uppy_utils'
```
ajkiessl, line 5: Is `simulateEditAndUpload` not being used anymore?  If so, can we delete it?

jlandiseigsti: Looks like it was deleted to me

ajkiessl: It is still defined here: https://github.com/psu-libraries/scholarsphere/blob/preview/dependencies-bump/app/javascript/controllers/uppy_utils.js#L14

spec/support/feature_helpers.rb
```diff
@@ -14,19 +14,12 @@ def retry_click(count: 0)
...
+
     end
   end
 
-  def print_html(selector)
-    puts "HTML for #{selector}:"
-    puts page.find(selector)['innerHTML']
-    puts ''
-  end
+  def print_html(selector); end
```
ajkiessl: Why was this method reduced to nothing?  Can it be removed?

app/javascript/controllers/uppy_controller.js
jlandiseigsti: Oh, I understand. Yeah, I think that can be removed.

spec/support/feature_helpers.rb
karthikreddy-02: This was a rubocop fix, I think I will revert these changes back and ask rubocop to ignore them.

app/javascript/controllers/uppy_controller.js
karthikreddy-02: Will do!

spec/support/faker.rb
karthikreddy-02: Will do!

jlandiseigsti (approved):
Looks good to me!

PR #1936, opened 2026-04-09
Token verification on download links

### LINKED ISSUE #1928: Add Signed Tokens to Protect Download Endpoint
We want to further protect our downloads endpoints from bots. Cloudflare has reduced most of the automated traffic to these endpoints.  However, we may not want to _only_ rely on Cloudflare here.

We could generate a signed token to attach to the download URL in the download links of a work version's show page.  (https://api.rubyonrails.org/v7.1.0/classes/ActiveRecord/SignedId.html#method-i-signed_id)

Then use that token to validate the request in the downloads controller, before triggering the accessibility remediation process.

This would guarantee that the request being sent to the downloads controller was sent by a user that rendered the show page.  The token should expire fairly quickly (like 30 seconds), and the user should be redirected to the show page if token matching fails.

ajkiessl:
Replace `download=true` process on file download links/controller with a signed token verification process.  I set the token to expire after 8 minutes, this may need adjusted but I think that's long enough for most real users.  If the token expires, the user will still be directed to the "View", where they can still download.  

This will confirm that users are downloading and triggering remediation only when the show page has been rendered in a browser.

closes #1928

### DIFF
--- a/app/components/file_download_link_component.rb
+++ b/app/components/file_download_link_component.rb
@@ -25,11 +25,13 @@ class FileDownloadLinkComponent < ViewComponent::Base
   end
 
   def download_path(download: true)
+    query_params = { resource_id: work_version_uuid }
+    query_params[:download_token] = download_token if download
+
     Rails.application
       .routes.url_helpers
       .resource_download_path(@file_version_membership.id,
-                              resource_id: work_version_uuid,
-                              download: download)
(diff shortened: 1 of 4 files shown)


jlandiseigsti (approved):
Looks good! I learned about a new Rails Feature with this one

PR #1939, opened 2026-04-13
Hotfix scholarsphere

ajkiessl:
Faraday was having issues in uldev when sending requests to the LibAnswers API.  There was something wrong with the query params in the url, so I explicitly configured Faraday to use the `url_encoded` setting.  It is working now.  I regenerated some of the VCR mocks, too.  They didn't really change, but at least they indicate the `application/x-www-form-urlencoded` content type now.

I checked all the other Faraday implementations, and they work, so nothing else should need fixed.

I also noticed the db schema got messed up, so i fixed that.  And there was a bug in the `requests_alternate_format` method in the `LibanswersApiService` when setting the queue ID that I fixed.


jlandiseigsti (approved):
Looks good to me

PR #1940, opened 2026-04-15
Do not create LibAnswer ticket when quota error is received and rake task for resetting failed remediations

### LINKED ISSUE #1935: Do not create LibAnswers ticket if auto-remediation fails from exceeding quota
We don't want the accessibility team being notified about these, because they didn't actually fail remediation.  They just need to be reset and re-ran once we make adjustments to get around the quota.

We receive the error message from the PDF API, so we can match the pattern in the error message to determine if a LibAnswers ticket should be created.  This _is_ flimsy (we should really add a whole new response type from the PDF API in the future), but this works for now.

ajkiessl:
closes #1935 

There's also a rake task in here for resetting auto-remediation timestamps/fields for failed remediation attempts.  So they can go through the workflow again.


jlandiseigsti (approved):
Looks good to me

PR #1956, opened 2026-05-05
add open access subpathway to the scholarly work pathway

### LINKED ISSUE #1945: Create Open Access Scholarly Work pathway
### WDLL

- [x] A new `work_deposit_pathway` (Open Access Scholarly Work) that has all the same fields as the existing Scholarly Work pathway.  There may be a way to do this without creating an entirely new pathway.  If so, we should do that.
- [x] When a user selects a Scholarly Work from the "Work Type" dropdown, a checkbox will appear (will need Javascript for this) asking "Is this an Open Access submission?".  If this is checked, a boolean should be stored with the `work_version` indicating it is Open Access.  This boolean will determine if the user is sent through the OA Scholarly Work pathway.
- [x] Restrict users from editing: publisher statement, embargo date, and license (this will be retrieved for them).
- [x] The "Review and Publish Page" will have a section for "Open Access Version" below the "Files" section with two radio button options: "Accepted Version" and "Published Version".

The other two parts of this (the version guesser and retrieving/populating the open access policy info), can be worked on separately and will be defined in separate tickets.

This acceptance criteria is derived from this user story: https://github.com/psu-libraries/researcher-metadata/issues/1156#issuecomment-4193287980

Smullz622:
Fixes #1945 

Rather than create a new pathway, I kept it within the scholarly work pathway & just added open access as a boolean to work version to control field visibility. Briana also requested that once something is set to open access, only admins can change that.

### DIFF
--- a/app/views/dashboard/form/publish/_version.html.erb
+++ b/app/views/dashboard/form/publish/_version.html.erb
@@ -1,0 +1,10 @@
+<div class="form-group">
+  <div class="form-check ms-5">
+    <%= form.radio_button :open_access_version, 'accepted', class: 'form-check-input' %>
+    <%= form.label :open_access_version, 'Accepted Version', value: 'accepted', class: 'form-check-label' %>
+  </div>
+  <div class="form-check ms-5">
+    <%= form.radio_button :open_access_version, 'published', class: 'form-check-input' %>
+    <%= form.label :open_access_version, 'Published Version', value: 'published', class: 'form-check-label' %>
+  </div>
+</div>
--- a/app/controllers/dashboard/form/publish_controller.rb
+++ b/app/controllers/dashboard/form/publish_controller.rb
@@ -118,6 +118,8 @@ module Dashboard
               :subtitle,
               :rights,
               :version_name,
+              :open_access,
+              :open_access_version,
               :published_date,
               :depositor_agreement,
               :psu_community_agreement,
--- a/app/controllers/dashboard/form/work_version_type_controller.rb
+++ b/app/controllers/dashboard/form/work_version_type_controller.rb
@@ -35,6 +35,7 @@ module Dashboard
             .require(:work_version)
             .permit(
               :title,
+              :open_access,
               work_attributes: [
                 :id,
                 :work_type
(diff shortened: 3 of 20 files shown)


app/views/dashboard/form/publish/_version.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="form-group">
+  <div class="form-check ms-5">
+    <%= form.radio_button :open_access_version, 'accepted', class: 'form-check-input' %>
+    <%= form.label :open_access_version, 'Accepted Version', value: 'accepted', class: 'form-check-label' %>
+  </div>
+  <div class="form-check ms-5">
+    <%= form.radio_button :open_access_version, 'published', class: 'form-check-input' %>
+    <%= form.label :open_access_version, 'Published Version', value: 'published', class: 'form-check-label' %>
+  </div>
+</div>
```
ajkiessl: Styling the top div with "form-wrapper" and adding some padding below the radio buttons centers this form a little better:

Image

```suggestion
<div class="form-wrapper">
  <div class="form-check mb-2">
    <%= form.radio_button :open_access_version, 'accepted', class: 'form-check-input' %>
    <%= form.label :open_access_version, 'Accepted Version', value: 'accepted', class: 'form-check-label' %>
  </div>
  <div class="form-check mb-2">
    <%= form.radio_button :open_access_version, 'published', class: 'form-check-input' %>
    <%= form.label :open_access_version, 'Published Version', value: 'published', class: 'form-check-label' %>
  </div>
</div>
```

ajkiessl: Also, we should probably call this file `_open_access_version.html.erb` to avoid confusion with other "version" wording.

ajkiessl (changes requested):
This looks good to me.  There were just a couple small things below. Also, I'm curious — since this will be the starting point for a decent sized feature — is there a place you think this could be feature flagged?  Then we could merge this and continue deploying as we get other things completed.

Smullz622:
> is there a place you think this could be feature flagged?

@ajkiessl I think we'd be covered by just putting the checkbox itself behind one. If that's hidden, nothing ever gets set as open access & everything functions as usual.

ajkiessl (approved):
LGTM

PR #1958, opened 2026-05-06
prevent work versions from re entering Airtable after removal

### LINKED ISSUE #1953: Review Airtable/Curation process
Curators have been noticing when they delete something from AirTable, it often gets re-uploaded.  In briefly reviewing the code, I've noticed we have a scope to determine "recently published" work versions in the work model class:

```
scope :recently_published, -> {
    joins(:versions)
      .where(work_versions: { aasm_state: 'published' })
      .where(work_versions: { sent_for_curation_at: nil })
      .distinct
  }
```

This is used to determine which work versions should be sent to AirTable.  I'm not sure if it was intentional to build this scope off of the works model.  I believe an issue with this is the most recent version gets flagged with the `sent_for_curation_at` timestamp.  So, if the work has older versions, it continues to be pulled into the `recently_published`.  This isn't an issue until that most recent version's record is deleted in AirTable.  It just gets re-imported because of this.  Was this intentional?

Also, curators do not need to curate auto-remediated work versions from the PDF accessibility API.  Should we find a way to filter these out of the AirTable export?

Smullz622:
Fixes #1953 

I left the scope on Work because ultimately we want to be able to check Airtable for all versions/make new versions as updated version & having it on Work better facilitates that. Instead I added a condition to exclude versions older than 30 days as well as an additional check on the sync service to make sure the version about to be sent has not been curated already.

This also prevents remediated works from being sent to Airtable

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -53,12 +53,14 @@ class Work < ApplicationRecord
 
   accepts_nested_attributes_for :versions
 
-  scope :recently_published, -> {
+  scope :send_for_curation_sync, -> {
     joins(:versions)
       .where(work_versions: { aasm_state: 'published' })
       .where(work_versions: { sent_for_curation_at: nil })
+      .where(work_versions: { created_at: 30.days.ago.. })
       .distinct
   }
+
   validate :embargoed_until_is_valid_date
   has_paper_trail versions: :paper_trail_versions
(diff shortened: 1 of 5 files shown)


ajkiessl (approved):
LGTM, thanks!

PR #1967, opened 2026-05-11
1957 draft creation endpoint

### LINKED ISSUE #1957: Allow draft creation through the API `/ingest` endpoint
ScholarSphere previously allowed draft creation through the API.  This commit can give some context:

https://github.com/psu-libraries/scholarsphere/commit/587467e100ed27f489d8ace604acba642853e73c

I believe this was used in the ScholarSphere 3 migration.  We removed this feature when we added the RMD integration thinking we would not want to create drafts.  Since we want to create a draft to move the RMD Open Access upload workflow into ScholarSphere, we should add this feature back in with some differences.

We still want the `/ingest` endpoint to default to the existing "publish or fail" behavior.  So, we can modify the endpoint to allow an optional `publish` parameter that can be `true` or `false`.  If `true` is passed, then the API works as it does now.  If nothing is passed, then it also works as it does now.  If `false` is passed, then instead of publishing it should just create a draft.  If there is an error on creation, the response should be like the response when publishing raises an error.  If creation is successful, the response should be a `200` and return the URL for the work (like publish does).  However, this response should also include an `edit_url`.  The `edit_url` will contain the URL to the "Files" upload form in ScholarSphere for the work version.

Important note: the [scholarsphere-client](https://github.com/psu-libraries/scholarsphere-client) will need modified to handle this new parameter.  Also, we should modify the API docs to include this new parameter.

jlandiseigsti:
closes #1957

### DIFF
--- a/spec/controllers/api/v1/ingest_controller_spec.rb
+++ b/spec/controllers/api/v1/ingest_controller_spec.rb
@@ -233,7 +233,7 @@ RSpec.describe Api::V1::IngestController do
       end
     end
 
-    context 'with missing false AND publish: false', vcr: VCRHelpers.depositor_cassette do
+    context 'with missing file AND publish: false', vcr: VCRHelpers.depositor_cassette do
       before do
         post :create, params: {
           metadata: {
--- a/spec/controllers/api/v1/ingest_controller_spec.rb
+++ b/spec/controllers/api/v1/ingest_controller_spec.rb
@@ -203,6 +203,60 @@ RSpec.describe Api::V1::IngestController do
       end
     end
 
+    context 'with publish: false', vcr: VCRHelpers.depositor_cassette do
+      before do
+        post :create, params: {
+          metadata: {
+            title: metadata[:title],
+            work_type: Work::Types.default,
+            description: metadata[:description],
+            published_date: metadata[:published_date],
+            creators: [creator],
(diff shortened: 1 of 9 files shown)


spec/controllers/api/v1/ingest_controller_spec.rb
```diff
@@ -203,6 +203,60 @@
...
+          "{\"message\":\"Work was successfully created\",\"url\":\"/resources/#{Work.last.uuid}\"," +
+          "\"edit_url\":\"/dashboard/form/work_versions/#{Work.last.latest_version.id}/files\"}"
+        )
+        expect(Api::V1::WorkCreator).to have_received(:call).with(
+          a_hash_including(external_app: api_token.application)
+        )
+      end
+    end
+
+    context 'with missing false AND publish: false', vcr: VCRHelpers.depositor_cassette do
```
ajkiessl: Should this be "with missing file"?

ajkiessl (changes requested):
There was just one minor thing I found.  Otherwise this looks good!

spec/controllers/api/v1/ingest_controller_spec.rb
jlandiseigsti: Good catch!

ajkiessl (approved):

PR #1969, opened 2026-05-12
Open Access Version Guesser

### LINKED ISSUE #1950: Implement article version guesser
When the "Files" form is submitted in the Open Access Scholarly Works pathway, the article version guesser should kick-off in the background. ~We should add a metadata field for Scholarly Works work versions to store the `article_version`.  The `article_version` can only be `accepted_version` or `published_version`.~  The user should not be allowed to publish until the version guesser is done.  This may conflict with the Accessibility Checker.  Since we now have an accessibility solution for published PDFs, we should be okay removing the Accessibility Check requirement.  The version guessing will take a minute anyway, so the Accessibility Check should finish in that time, too.  When the guess is completed in the background, it should update the work version's `open_access_version` and live-update the publish screen with the selected article version. If no version is determined, the user should be prompted to select a version.

~_Note: we may want to come up with a better name for the `article_version` field.  This may be confusing since ScholarSphere has its own versioning._~ Edit: we went with `open_access_version`
______________________________________

The code for the version guesser can be found in RMD in the standard OA workflow pathway code.  The bulk of the code exists here: 

https://github.com/psu-libraries/researcher-metadata/blob/main/app/models/file_version_checker.rb 

We call it the `FileVersionChecker`, but I think we'll need to modify the naming in ScholarSphere to not confuse it with other classes/concepts.  It uses this CSV: 

https://github.com/psu-libraries/researcher-metadata/blob/main/config/file_version_checking_rules.csv 

to match patterns to help identify if a PDF is an accepted or published version.  So, we'll need that too.  

There is some polling done in RMD: 

https://github.com/psu-libraries/researcher-metadata/blob/main/app/controllers/open_access_publications_controller.rb#L73-L115 
https://github.com/psu-libraries/researcher-metadata/blob/main/app/javascript/profile/file_version_polling.js  

This may not be needed in the ScholarSphere implementation, since we can just live-update the screen when the database record is updated after the version is determined.

ajkiessl:
closes #1950 

Implements the open access version guessing feature from RMD.  The implementation is a little different, but functionality is pretty much the same.  The most notable difference, but still minor, is that we don't store journal info in ScholarSphere, and journal matching in document metadata is a published version signal.  We do hold publisher info, so I am using that instead.

Also, I added `unknown` as an open access version value.  This is useful to store if the version cannot be determined.  It doesn't need to be an option users see, though.


app/services/open_access_version/guesser.rb
```diff
@@ -0,0 +1,79 @@
...
+
+module OpenAccessVersion
+  class Guesser
+    # The OpenAccessVersion::Guesser attempts to determine whether a WorkVersion's
+    # PDF or docx file(s) indicate that the WorkVersion is an accepted, published, or unknown
+    # version.  It does this with three steps in order of precedence:
+    #   1. Check for version signals in the PDFs' or docx files' EXIF metadata
+    #   2. Check for an arXiv watermark in the PDFs' content
+    #   3. Check for version signals in the PDFs' content and filename using the
+    #      OpenAccessVersion::ScoreCalculator.  The score is accumulated across all PDF files
```
jlandiseigsti, line 13: Appreciate this summary.

jlandiseigsti (approved):
Everything looked good to me – matched the logic that had been in RMD, with good documentation to explain the process.

PR #1971, opened 2026-05-13
auto populate open access permissions

### LINKED ISSUE #1951: Import Open Access Permissions information
In the Open Access Scholarly Works pathway, when the "Files" page is submitted, a request should be sent out to OA.Works to find the Open Access Permissions information for this work.  This requires the work's DOI (we may need to make the DOI a required field earlier in the workflow).  There is a ruby client in RMD that can do this: 

https://github.com/psu-libraries/researcher-metadata/blob/main/app/services/oaw_permissions_client.rb 

We can then store the permissions data in a hidden field in the rendered HTML on the publish page.  Once the article version is determined, we can auto populate the permissions info (embargo, set statement, license) using javascript.  The auto populated data should change with the article version field, so if the user changes from "accepted version" to "published version", the permissions data should automatically change.

There is also logic in RMD to indicate to the user that they should upload a different version if the article version does not match up with what is retrieved from OA.Works, or if nothing is found.  This service class holds that logic: 

https://github.com/psu-libraries/researcher-metadata/blob/main/app/services/oaw_permissions_service.rb

For example, the user may upload the published version, but OA.Works does not have any permissions information for a published version — only the accepted version.  That means the user needs to upload the accepted version instead.  This logic will need to be implemented in javascript in ScholarSphere if we want the publish page to work as described above.  This component class in RMD has the logic for what renders on the page; depending on the correctness of the determined article version, and the pulled permissions data: 

https://github.com/psu-libraries/researcher-metadata/blob/main/app/components/scholarsphere_deposit_form_component.rb .

Depends on #1950 and #1945 (at least for the javascript part)

Smullz622:
Fixes #1951

@ajkiessl pushing this up in case it helps because I think we'll have some overlap around the version buttons (I assume you're using action cable to update them when version is determined & will have a js controller for that)
I think I've got all the behavior how we want it working, but it's not ready for full review because I want to see if I can refine some of it & need to add tests


app/services/oaw_permissions_client.rb
```diff
@@ -0,0 +1,87 @@
...
+
+class OawPermissionsClient
+  class InvalidVersion < StandardError; end
+  attr_reader :doi, :version
+
+  def all_permissions
+    @all_permissions ||= JSON.parse(permissions_response)['all_permissions']
+  rescue JSON::ParserError
+    nil
+  end
```
ajkiessl: I think the reason it takes so long to load when there is no DOI is because `all_permissions` does not memoize in the rescue block.  This should fix it:

```
def all_permissions
    return @all_permissions if defined?(@all_permissions)

    @all_permissions = JSON.parse(permissions_response)['all_permissions']
  rescue JSON::ParserError
    @all_permissions = nil
  end
```

app/models/work_version.rb
```diff
@@ -7,9 +7,9 @@ class WorkVersion < ApplicationRecord
   include GeneratedUuids
   include UpdatingDois
 
-  OPEN_ACCESS_VERSIONS = %w[
-    accepted
-    published
+  OPEN_ACCESS_VERSIONS = [
+    I18n.t('dashboard.works.edit.open_access_version.accepted_version'),
+    I18n.t('dashboard.works.edit.open_access_version.published_version')
   ].freeze
```
ajkiessl: We should remove these from the `I18n` config and just store them as hardcoded constants here.  This is a problem in RMD that we need to fix.  We should not be using `I18n` to store machine-used strings.  `I18n` should be reserved for text displayed to users in views so it can be translated to different languages.

app/services/oaw_permissions_client.rb
```diff
@@ -0,0 +1,87 @@
...
+        ['Attribution-ShareAlike 4.0 International (CC BY-SA 4.0)', 'https://creativecommons.org/licenses/by-sa/4.0/'],
+        ['Attribution-NonCommercial 4.0 International (CC BY-NC 4.0)', 'https://creativecommons.org/licenses/by-nc/4.0/'],
+        ['Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)', 'https://creativecommons.org/licenses/by-nd/4.0/'],
+        ['Attribution-NonCommercial-NoDerivatives 4.0 International (CC BY-NC-ND 4.0)', 'https://creativecommons.org/licenses/by-nc-nd/4.0/'],
+        ['Attribution-NonCommercial-ShareAlike 4.0 International (CC BY-NC-SA 4.0)', 'https://creativecommons.org/licenses/by-nc-sa/4.0/'],
+        ['Public Domain Mark 1.0', 'http://creativecommons.org/publicdomain/mark/1.0/'],
+        ['CC0 1.0 Universal', 'http://creativecommons.org/publicdomain/zero/1.0/'],
+        ['All rights reserved', 'https://rightsstatements.org/page/InC/1.0/']
+      ]
+    end
```
ajkiessl: Instead of defining all the rights here, could we pull and map to the authority file (`config/authorities/licenses.yml`) using the `Qa` gem?

app/services/oaw_permissions_client.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def oaw_permissions_base_url
+      'https://bg.api.oa.works/permissions/'
+    end
+
+    def accepted_version
+      if all_permissions.present?
+        all_permissions
+          .select { |perm| perm if perm['version'] == 'acceptedVersion' } # will need to add to translation
```
ajkiessl: Looks like you were planning to pull the version strings from the translation file.  Let's pull them from the constants in the `WorkVersion` model.

app/controllers/dashboard/form/publish_controller.rb
```diff
@@ -20,6 +21,7 @@ def update
         authorize(@work_version)
 
         @resource.attributes = work_version_params
+        @submitted_permissions = oaw_permissions_params
```
ajkiessl, line 24: I having trouble understanding what this is doing.  Does this hold the permissions selections if there's a validation error and the screen reloads?

ajkiessl (changes requested):
I have some change requests below 👇 


This might just be my environment, so if it's not happening for you then its okay, but the background card on the publish page doesn't reach the bottom of the form:

Image

app/controllers/dashboard/form/publish_controller.rb
Smullz622: Yes it does. It was erroring out otherwise. I stored them separately because it stores everything from the permissions service, not just selected values, because full permissions from each version needs to be carried over in case the user changes the version after the validation error

app/services/oaw_permissions_client.rb
Smullz622: I was able to use the qa gem to replace the hardcoding in `#rights_options` but the mapping in `#map_licence` is still needed because it maps what is returned from OAWorks specifically (not sure if you were suggesting to replace that as well, so just clarifying)

ajkiessl (dismissed):
@Smullz622 I merged in `main` and resolved the merge conflicts.  I also did a little refactoring.  Lmk if any of the changes don't look right, but everything seems to be working so I think this is good to go.

Smullz622:
@ajkiessl the tab header issue I noticed is actually on main, so this is good to go! Thank you!

ajkiessl (approved):
LGTM

PR #1973, opened 2026-05-16
1959 - send webhook to RMD when open access work is published

### LINKED ISSUE #1959: Open Access Scholarly Work publish webhook
When an Open Access Scholarly Work is published, a webhook request should be sent to RMD with a payload:

```
{
  scholarsphere_work_url: "/resources/{work_uuid}"
}
```

It will need to authenticate with RMD's webhook endpoint, so we'll also need to send a webhook token in the request.

karthikreddy-02:
When an Open Access work is published in ScholarSphere, send a webhook to RMD so it can create an OpenAccessLocation for the matching deposit.

Changes:

Add WorkPublishedWebhook service that POSTs /webhooks/scholarsphere/open_access_work_published to RMD with the work's URL path
Add WorkPublishedWebhookJob to enqueue the webhook asynchronously
Trigger the job via after_commit on the publish AASM event in WorkVersion, only when the work is open access
Depends on: psu-libraries/researcher-metadata#1232

Closes #1959


app/models/work_version.rb
```diff
@@ -227,12 +227,17 @@ def label(id)
...
   aasm timestamps: true do
     state :draft, intial: true
     state :published, :withdrawn, :removed
 
     event :publish do
+      after_commit do
+        if work.open_access?
+          WorkPublishedWebhookJob.perform_later(work.uuid)
+        end
+      end
```
ajkiessl: Is this valid syntax?  Did you test it?  There are no specs for this.

app/services/work_published_webhook.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class WorkPublishedWebhook
```
ajkiessl: We should modularize our webhooks services and move them into the `services/webhooks` directory.

spec/support/shoulda_matchers.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'shoulda/matchers'
```
ajkiessl: Why was this added?

app/models/work_version.rb
```diff
@@ -243,7 +248,9 @@ def label(id)
                     self.reload_on_index = true
                   }
     end
+    # rubocop:enable Metrics/BlockLength
```
ajkiessl: Rather than disabling, can this be refactored?  If it can't it might be better to just disable this cop for this file in the rubocop.yml.

ajkiessl (changes requested):

app/models/work_version.rb
```diff
@@ -243,7 +248,9 @@ def label(id)
...
 
+    # Fields that can contain multiple values...
```
ajkiessl: Remove comment

karthikreddy-02:
spec/support/vcr.rb
Replaced the broad ignore_hosts addition (which accidentally disabled VCR cassette playback for all RMD API requests) with a targeted ignore_request block that only bypasses VCR for the specific webhook path. This ensures existing RMD-related specs continue to use their cassettes normally.

spec/services/work_published_webhook_spec.rb
Set ENV['RMD_HOST'] in the before block so the blank-host guard in notify doesn't exit early — allowing the Faraday stub to be exercised locally where RMD_HOST is not set. Cleaned up with after { ENV.delete('RMD_HOST') }.

app/services/work_published_webhook.rb (deleted)
Stale leftover from before the service was moved into the Webhooks:: namespace. Removed to avoid confusion with the canonical app/services/webhooks/work_published_webhook.rb.

spec/support/vcr.rb
```diff
@@ -8,16 +8,15 @@
...
+# Allow connections to Docker images and webdriver update urls
 VCR.configure do |c|
   c.cassette_library_dir = 'spec/fixtures/vcr_cassettes'
   c.hook_into :webmock
   c.configure_rspec_metadata!
   c.allow_http_connections_when_no_cassette = true
   c.ignore_localhost = true
-  c.ignore_hosts *allowed_hosts
+  c.ignore_hosts 'selenium', 'minio', 'solr', 'docker.for.mac.localhost'
+  c.ignore_request { |req| req.uri.include?('/webhooks/scholarsphere/open_access_work_published') }
```
ajkiessl, line 21: I'm assuming this request was sending out every time a work was being published in the test suite?  So, in order to not generate a bunch of VCRs you've ignored it?  If so, I think that is a good idea.  It might be worth adding a comment above that line explaining why that request needs to be ignored. 

Also, the comment on line 11 doesn't make sense anymore because that array was moved to line 18.  We'll need to either move that comment down to above line 18, or revert that change back to the original `allowed_hosts` variable.

app/models/work_version.rb
```diff
@@ -446,6 +441,21 @@ def perform_update_index
...
+    def after_publish
+      work.try(:update_deposit_agreement)
+      if work.professional_doctoral_culminating_experience? || work.masters_culminating_experience? || work_type == 'instrument'
+        set_publisher_as_scholarsphere
+      end
+      self.reload_on_index = true
+    end
+
+    def enqueue_published_webhook
+      return unless @work_just_published && work.open_access?
```
ajkiessl: This is using the wrong method `.open_access?`.  This method looks at the work's permissions (access controls) to determine open access.  This isn't what we want, we want to check the `.open_access` boolean on `WorkVersion`.

app/models/work_version.rb
```diff
@@ -446,6 +441,21 @@ def perform_update_index
...
+
+      @work_just_published = false
```
ajkiessl, line 455: I think we could end up with this instance variable becoming stale if the above logic `returns`, or if there's a database rollback.  Can you set the instance variable to `false` above the logic?  And then use a method variable to store what's needed for the method like:

```
just_published = @work_just_published
@work_just_published = false
return unless just_published && work.open_access
```

Then you can handle rollbacks with a rollback hook:

```
after_rollback :clear_publish_flag

...

def clear_publish_flag
    @work_just_published = false
end
```

ajkiessl (changes requested):

ajkiessl (approved):
LGTM, thanks!

PR #1977, opened 2026-05-22
Upgrades Sidekiq to v8

ajkiessl:
We noticed some errors in the `ulprod` deployment indicating a gem version mismatch between `connection_pool` and `sidekiq`.  Upgrading Sidekiq to v8 fixes that.

Also, this removes the general redis ttl configuration and instead uses key-based ttl where needed.  Sidekiq no longer accepts this config value for its redis client.

Rubocop suddenly doesn't like `unprocessable_entity` as an http status and prefers `unprocessable_content`.  Those changes are included in this PR.  I created an issue to further review `unprocessable_entity` wording here: #1978 

Preview is here: https://scholarsphere-p-sidekiq-upgrade.uldev.k8s.libraries.psu.edu/admin/sidekiq/


jrpatterson (approved):
Thanks Alex for getting this working.  LGTM.   Never is just a version number incrementing.

PR #1980, opened 2026-05-26
add airtable label for open access uploads

### LINKED ISSUE #1979: Export open access indicator to AirTable
We want curators to know when a `WorkVersion` was uploaded through the open access scholarly work pathway.  We should add a field to AirTable for this, and when we know what this field is, add it to the AirTable exporter in ScholarSphere.

Smullz622:
Fixes #1979 

The label has already been added within Airtable


ajkiessl (approved):
LGTM

PR #1982, opened 2026-05-26
Use Sidekiq::Job instead of ActiveJob for MetdataListener::Job

ajkiessl:
There appears to be a thing with Sidekiq 8 where it does not allow the `ActiveJob` adapter queue to work with non-Rails apps.  Since metadata-listener is not a Rails app, it was failing to process its Sidekiq jobs.  This PR changes the metadata-listener job to use the base Sidekiq module instead of `ActiveJob`.  This module uses the `perform_async` method instead of `perform_later`.  The changes needed for metadata-listener are here: https://github.com/psu-libraries/metadata-listener/pull/130

### DIFF
--- a/app/jobs/metadata_listener/job.rb
+++ b/app/jobs/metadata_listener/job.rb
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 module MetadataListener
-  class Job < ApplicationJob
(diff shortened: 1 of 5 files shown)


jlandiseigsti (approved):
LGTM

Issue #1984, opened 2026-05-29
Review Scholarly OA Work upload help text and wording


ajkiessl:
Having a PO walk through the workflow and look for places where help text could be added or improved would be helpful.  We want to help the user understand what exactly the "Open Access" upload is and the parts of it that operate differently from a regular upload.

bdezray:
### Work Type Tab
**Help Text for Checkbox:** I am depositing this work to comply with [Penn State’s Open Access Policy (AC02)](https://policy.psu.edu/policies/ac02), and this work has a publisher-provided DOI. 
**Text to display below the checkbox or as hover/tool tip text:** 
Selecting this check box  and entering the publisher-provided DOI will initiate the following automated tools on your behalf:
1.	**Version Check** – The system will attempt to identify which version of the article you have uploaded to determine publisher sharing information. 
2.	**Publisher Sharing Information** – Using the identified version, the system will attempt to retrieve the publisher’s default sharing policy. It will then notify you regarding your permission to share the uploaded version as well as assign the appropriate license, embargo (if applicable), and publisher statement (if applicable). 
3.	**Metadata Import** – If information about this scholarly work is available in the [Penn State Researcher Metadata Database](https://metadata.libraries.psu.edu), it will be automatically imported into the appropriate fields. You may retain or edit imported information.   

**_Please note:_** You will not be able to modify the selection of this box or the DOI after completing a search. If you find that either the selection of the DOI is incorrect, please delete the draft deposit and begin a new submission.

### Autocomplete for Scholarly Works
•	Only display when check-box is selected and make required. 
•	Remove the language in the second paragraph: “Please note that if you are working to make your work open per [AC02](https://policy.psu.edu/policies/ac02), we recommend using the process available directly within the [Researcher Metadata Database](https://openaccess.psu.edu/deposit/rmd/) for deposit when possible.”

### Open Access Version
•	Displayed above radio buttons:
The system has attempted to identify which version of the article you have uploaded. If a version could not be detected, no option will be pre-selected and you will be asked to choose manually. If you believe the detected version is incorrect, you may override it by selecting the appropriate option below. More information about the version options can be found in the [AC02 FAQ](https://psu.libanswers.com/faq/279953).

**_Please note:_** The version you select will be used to look up the publisher’s default sharing policy and complete the publishing details section including license and embargo. Selecting the correct version helps ensure accurate implementation of policy requirements.

ajkiessl:
completed in #1990

PR #1993, opened 2026-06-03
Require DOI for open access sub pathway

### LINKED ISSUE #1990: require doi for open access sub pathway
The open access sub pathway doesn't add value without a doi to pull permissions data. Making it required will also resolve the issue of works without a doi being blocked on publish because permissions data cannot be found. There are two approaches we discussed with this:

1) Make publisher identifier (DOI) the required field. This field can have multiple values, some of which may not be doi, so we would have to add validation logic to make sure at least one value is a valid doi & that is the value used to pull permissions.

2) Make the doi in the top section `Autocomplete for Scholarly Works' required. The only complication here is the value is not saved to identifier unless the pull from RMD is successful so we would want to set the identifier first. 

@bdezray & I discussed & decided option 2 would provide the best user experience by making it very clear where to enter DOI, that it is required, and cleans up a lot of the current help text to treat it as one process. Additionally, that autocomplete section is only relevant for open access works in the sub pathway anyway, so it should only be displayed if the work version is an open access submission anyway.

Smullz622:
RMD autocomplete form becomes a centralized doi submission form for the open access subpathway. The RMD form is only applicable to the open access sub pathway anyway so it makes sense to combine functionality for a better user experience. The form will still attempt to pull from RMD, but it will now save doi to identifier regardless of whether the RMD pull is successful. This form will now be required (enforced by a check that `imported_metadata_from_rmd` is not nil, ie the form has been submitted) to ensure any works going through the OA sub pathway have a DOI since auto populating permissions (the main feature of this sub pathway) cannot happen without DOI.

Note - it was a deliberate decision to leave the help text for the oa upload checkbox visible on the publish page because users coming from RMD land on the files tab & won't have seen it yet.

Fixes #1990 and #1984 

Also addresses some small issues in the OA sub pathway:
- allow publishing when no permissions are returned
- make open access upload version required when in OA sub pathway
- upload version help text was breaking when the form returned an error because versionFound was being set as a string instead of JSON

### DIFF
--- a/app/controllers/api/v1/ingest_controller.rb
+++ b/app/controllers/api/v1/ingest_controller.rb
@@ -88,6 +88,7 @@ module Api::V1
             :publisher_statement,
             :doi,
             :open_access,
+            :open_access_version,
             :imported_metadata_from_rmd,
             keyword: [],
             resource_type: [],
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -170,6 +170,8 @@ class WorkVersion < ApplicationRecord
               in: OpenAccessVersion::VersionValues::ALL
             }
 
+  validate :open_access_version_required, if: :published?
+
   validates :rights,
             presence: true,
             inclusion: {
@@ -467,6 +469,12 @@ class WorkVersion < ApplicationRecord
           open_access_version: open_access_version
         }
       )
+    end
+
+    def open_access_version_required
+      if open_access && OpenAccessVersion::VersionValues::KNOWN.exclude?(open_access_version)
+        errors.add(:open_access_version, :blank)
+      end
     end
 
     def strip_blanks_from_array(arr)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -441,7 +441,7 @@ en:
             heading: Withdraw A Version
         open_access_version:
           permissions_not_found: >
-             We could not find sharing rules for your work. Please contact us for assistance.
+             We could not find sharing rules for your work. Please fill out any missing information.
           other_version_preferred: >
              We could not find sharing rules for the %{this_version} of this work, only the %{other_version}. To upload the %{other_version}, please return to the file upload page and upload your %{other_version} there. If you are sure you want to upload the %{this_version}, please contact us for assistance.
       update:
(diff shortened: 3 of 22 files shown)


app/controllers/api/v1/ingest_controller.rb
```diff
@@ -88,6 +88,7 @@ def metadata_params
             :publisher_statement,
             :doi,
             :open_access,
+            :open_access_version,
```
Smullz622, line 91: required for validation (see comment on work_version)

app/models/work_version.rb
```diff
@@ -170,6 +170,8 @@ def label(id)
               in: OpenAccessVersion::VersionValues::ALL
             }
 
+  validate :open_access_version_required, if: :published?
+
```
Smullz622, line 174: I tried adding this validation as just a UI form validation, but with it being a sometimes hidden field it got tricky & model validation was cleaner. This does impact the API, but I think that's actually how we want it to work because if someone is publishing a work marked as open access through the API, we would want the upload version as well

ajkiessl: Yes, I agree with this.

config/locales/en.yml
```diff
@@ -278,6 +278,9 @@ en:
         manufacturer: Organization or individual that built the instrument.
         rights: The license you select determines how your work can be used by others. [Read more about choosing a license](https://docs.scholarsphere.psu.edu/guides/licenses/)
         contributor: If applicable, please include a brief acknowledgment of any funding sources that supported your research. This helps ensure proper attribution and increases the visibility of your funder’s contribution.
+        open_access_version: >
+          The system has attempted to identify which version of the article you have uploaded. If a version could not be detected, no option will be pre-selected and you will be asked to choose manually. If you believe the detected version is incorrect, you may override it by selecting the appropriate option below. More information about the version options can be found in the <a href="https://psu.libanswers.com/faq/279953">AC02 FAQ</a>.
```
ajkiessl: Since there will be times when the user gets to the file version form while the checking is still happening, and if no version is found we don't show them any message, I think changing the first two sentences a little will help the user.  Maybe like:

"ScholarSphere attempts to identify which version of the article you have uploaded. If a version could not be detected, no option will be pre-selected and you will need to choose manually."

ajkiessl (changes requested):
This all LGTM.  Just one minor wording suggestion below 👇

ajkiessl (approved):

PR #1994, opened 2026-06-03
Live update open access version

### LINKED ISSUE #1972: Live update `open_access_version` on work version publish page
When a `WorkVersion`'s `open_access_version` is updated, the boolean on the publish page should update via `ActionCable`.

When this happens after the user uploads their files, the open access version boolean field on the publish page should indicate that the files are processing and that the user cannot make a selection.  Users also should not be able to publish during this time.  

Once the version is updated through the model, a live update to the screen should set the value and explain that our system believes the user uploaded an accepted or published version.  If no version is determined, there should be a message saying our system could not determine a version and the user needs to select it themselves.  Also, if a version is determined, we should let the user know they can change it, but this will affect their permissions data.

ajkiessl:
closes #1972 

I built this using Rails ActionCable broadcasting and then realized it doesn't work if the open access version is determined very quickly.  This is because the browser doesn't have enough time to create a websocket connection on page load before the open access version is determined.  So, I implemented a Rails endpoint that also relays the open access version.  The JS on the publish page checks this on page load _and_ sets up a websocket connection for slower open access version processing.  Technically, I could remove all the ActionCable setup now and just do JS polling, but I figure at this point its not worth it.  The ActionCable setup isn't that complicated anyway.

A spinner shows while the page waits for an open access version update:
Screenshot 2026-06-04 at 2 52 11 PM

### DIFF
--- a/app/javascript/controllers/open_access_version_controller.js
+++ b/app/javascript/controllers/open_access_version_controller.js
@@ -54,19 +54,19 @@ export default class extends Controller {
 
   startTimer() {
     setTimeout(() => {
-        const spinnerVisible = this.hasLoadingTarget && !this.loadingTarget.classList.contains('d-none')
-        const controlsHidden = this.hasControlsTarget && this.controlsTarget.classList.contains('d-none')
+      const spinnerVisible = this.hasLoadingTarget && !this.loadingTarget.classList.contains('d-none')
+      const controlsHidden = this.hasControlsTarget && this.controlsTarget.classList.contains('d-none')
 
-        if (spinnerVisible && controlsHidden) {
-          if (this.hasControlsTarget) this.controlsTarget.classList.remove('d-none')
-          if (this.hasLoadingTarget) this.loadingTarget.classList.add('d-none')
-        }
+      if (spinnerVisible && controlsHidden) {
+        if (this.hasControlsTarget) this.controlsTarget.classList.remove('d-none')
+        if (this.hasLoadingTarget) this.loadingTarget.classList.add('d-none')
+      }
 
-        // then unsubscribe from the channel
-        if (this.subscription && this.subscription.unsubscribe) {
-          try { this.subscription.unsubscribe() } catch (e) { void e }
-        }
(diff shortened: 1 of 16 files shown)


app/javascript/controllers/open_access_version_controller.js
```diff
@@ -1,13 +1,85 @@
...
+      this.subscription.unsubscribe()
+    }
+  }
+
+  createSubscription(id) {
+    this.subscription = consumer.subscriptions.create(
+      { channel: 'OpenAccessVersionChannel', id: id },
+      {
+        received: (data) => {
+          const targetId = this.data.get('id')
```
jlandiseigsti: I may be missing something, but is there a reason we can't use `id`, given that it also is defined as `this.data.get('id')` (line 14)?

app/javascript/controllers/open_access_version_controller.js
```diff
@@ -1,13 +1,85 @@
...
+    setTimeout(() => {
+      const spinnerVisible = this.hasLoadingTarget && !this.loadingTarget.classList.contains('d-none')
+      const controlsHidden = this.hasControlsTarget && this.controlsTarget.classList.contains('d-none')
+
+      if (spinnerVisible && controlsHidden) {
+        if (this.hasControlsTarget) this.controlsTarget.classList.remove('d-none')
+        if (this.hasLoadingTarget) this.loadingTarget.classList.add('d-none')
+      }
+
+      // then unsubscribe from the channel
```
jlandiseigsti: We can lose this comment

app/javascript/controllers/open_access_version_controller.js
ajkiessl: Ah good catch!

jlandiseigsti (changes requested):
Just one other minor thing I saw. I also don't know jest very well, so I'm not sure if there's anything to be changed there; I'll put learning about that testing framework on my todo list.

jlandiseigsti (approved):
👍

PR #1995, opened 2026-06-04
1991 - makes open access checkbox ro

### LINKED ISSUE #1991: Hide "Is this an Open Access Submission?" on publish page
This checkbox shows up for non-admins as editable on the publish page if it is not checked.  Can we remove this from the publish page?  Or at least make it read-only so it can't be changed at the publish step.

jlandiseigsti:
closes #1991

If the type choice is also read-only. This will always be true on the publish screen, and be true on the view screen if the work is imported from rmd.


ajkiessl (approved):
LGTM

PR #1996, opened 2026-06-05
Rename open_access to open_access_upload on WorkVersion

karthikreddy-02:
Renames the open_access boolean column on WorkVersion to open_access_upload to avoid conflicts with the existing open_access? method on the Permissions concern, which checks access controls and is unrelated to this field.

Changes:

Migration to rename the column
Updated all references across models, forms, controllers, and views
Added explicit open_access_upload? helper method on WorkVersion
Updated all affected specs
Note: The open_access? method in app/models/concerns/permissions.rb was intentionally left unchanged as it checks ACL-based visibility and is unrelated to this column.

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -448,7 +448,7 @@ class WorkVersion < ApplicationRecord
     def enqueue_published_webhook
       just_published = @work_just_published
       @work_just_published = false
-      return unless just_published && open_access
+      return unless just_published && open_access_upload
 
       WorkPublishedWebhookJob.perform_later(work.uuid)
     end
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -170,6 +170,8 @@ class WorkVersion < ApplicationRecord
               in: OpenAccessVersion::VersionValues::ALL
             }
 
+  validate :open_access_version_required, if: :published?
+
   validates :rights,
             presence: true,
             inclusion: {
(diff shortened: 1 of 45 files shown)


db/schema.rb
```diff
@@ -10,9 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2026_04_28_182459) do
+ActiveRecord::Schema[7.2].define(version: 2026_06_04_000000) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
ajkiessl, line 15: `enable_extension "pg_stat_statements"` needs added back in.

db/schema.rb
```diff
@@ -232,7 +231,8 @@
   end
 
   create_table "versions", force: :cascade do |t|
-    t.string "item_type", null: false
+    t.string "item_type"
+    t.string "{null: false}"
```
ajkiessl: This needs reverted too.

ajkiessl (changes requested):
This looks good.  The only issue here is that the db/schema got a little messed up.  This tends to happen locally for some reason.  I explain what needs reverted below 👇

app/models/work_version.rb
```diff
@@ -393,6 +393,10 @@ def has_publisher_doi?
     !!identifier.find { |id| Doi.new(id).valid? }
   end
 
+  def open_access_upload?
+    open_access_upload == true
+  end
```
ajkiessl: I actually think Rails automatically gives us this method for booleans.  So, you can remove this and either implement `open_access_upload?` in all the places like: `work_version.open_access_upload`, or just leave the implementations as they are.

ajkiessl (changes requested):

ajkiessl (approved):
LGTM

PR #2005, opened 2026-06-08
Adds JS form validation on publish button 

### LINKED ISSUE #2002: Cannot publish when "Given Name" and/or "Last Name" are empty
Not being able to publish is expected behavior, however, no error pops up and there's no message or any indication of what needs added.  The button simply doesn't do anything.  This is particularly a problem for uploads from RMD, since there is no way to get "Given Name" and "Last Name" to fill in from RMD.

This needs to be addressed before releasing the new OA Scholarly Workflow feature, since this will be very confusing for users coming from RMD.

ajkiessl:
To direct users to incomplete parts of the form _before_ popup covers the screen.

This was a much easier fix than I predicted.  We don't have model level validations for `given_name` and `last_name` for contributors, because those names usually are a bit messy.  However, we do have front-end requirements for them, to encourage users to add them.  The problem was that the popup was blocking the frontend validation errors.

https://github.com/user-attachments/assets/1b52eacc-9dd8-4e3a-94d3-7b4d6b37e4fd

Closes #2002

### DIFF
--- a/spec/features/dashboard/work_version_form_spec.rb
+++ b/spec/features/dashboard/work_version_form_spec.rb
@@ -1227,6 +1227,21 @@ RSpec.describe 'Publishing a work', with_user: :user do
   end
 
   describe 'The Publish tab' do
+    context 'when required contributor fields are blank', :js do
+      let(:work_version) { create(:work_version, :article, :able_to_be_published) }
+      let(:user) { work_version.work.depositor.user }
+
+      it 'does not open the publish dialog and keeps focus on form validation' do
+        visit dashboard_form_publish_path(work_version)
+
+        fill_in 'work_version_creators_attributes_0_given_name', with: ''
+
+        click_on I18n.t!('dashboard.form.actions.publish.button')
+
+        expect(page).to have_no_css('dialog#hidden-publish-dialog[open]')
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


spec/features/dashboard/work_version_form_spec.rb
```diff
@@ -1227,6 +1227,21 @@
...
+      let(:user) { work_version.work.depositor.user }
+
+      it 'does not open the publish dialog and keeps focus on form validation' do
+        visit dashboard_form_publish_path(work_version)
+
+        fill_in 'work_version_creators_attributes_0_given_name', with: ''
+
+        click_on I18n.t!('dashboard.form.actions.publish.button')
+
+        expect(page).to have_no_css('dialog#hidden-publish-dialog[open]')
```
EricDurante, line 1241: If it works and is no more difficult to implement than this, then I'd prefer a test that actually tries to interact with the elements that were being blocked before as opposed to a test that just checks for the presence of the thing that was blocking them.

EricDurante: Two reasons for that ☝️:

1. That's what the user would actually be doing.
2. The check for the blocking element is brittle. If the the naming of anything gets changed, then the test could give a false negative.

EricDurante (approved):

PR #2006, opened 2026-06-08
Adds a notice message on files upload page for when users are directed here from an external location

### LINKED ISSUE #2003: Entry to ScholarSphere from RMD popup
Is there some way we can add a flash message for users coming from RMD letting them know they are in ScholarSphere and they should start by uploading their files and finish up reviewing the metadata on the next page?

ajkiessl:
This is determined by a query param, which will need added to the external request (from RMD).

Screenshot 2026-06-08 at 10 49 53 AM

Closes #2003


jlandiseigsti (approved):
LGTM

PR #2008, opened 2026-06-08
Do not show OA checkbox if the work version is not the initial draft

### LINKED ISSUE #2004: Only allow Open Access Scholarly Work Pathway for first version of a `Work`?
Should this only be allowed for the first `WorkVersion` in a `Work`?

It seems like this would be unnecessary for subsequent versions of the same work, but I am not sure.

ajkiessl:
There's no reason to allow an Open Access upload for anything beyond the first version, so this will help to avoid confusion.

closes #2004


jlandiseigsti (approved):
👍

PR #2010, opened 2026-06-09
Moves webhook rmd publish callback to publish controller

ajkiessl:
In QA, the webhook callback in the model was running even when the WorkVersion was not being published.  I moved the webhook implementation to the publish controller.  This gives us more reliable behavior, even if it's only tied to GUI interactions.  This webhook does not _need_ to be tied to model commit/update callbacks, so this should work for us.


ajkiessl:
@karthikreddy-02 Just a heads up.  I moved some of this webhook logic around.  In QA, it was triggering when a draft was created.  Idk why because the code logic indicates it should only happen on publish.  I moved the webhook into the publish controller.  We should have more control over it there.

PR #2012, opened 2026-06-09
Hacky "auto-login"

ajkiessl:
Some major oversight on my part: 

When testing the RMD/ScholarSphere upload workflow in QA I realized if I wasn't already logged into ScholarSphere, I got redirected to ScholarSphere's home page.  We don't have any _real_ auto-login functionality in ScholarSphere right now.  So, I implemented a kind of hacky login that redirects to a page with a hidden "Login" button that uses Javascript to automatically click the button.  This does the Azure login and redirects back to the original page (in our case the files/edit page).

We should investigate real auto-login in the future: #2013


app/views/users/auth_redirect/show.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= form_with url: user_azure_oauth_omniauth_authorize_path,
+              method: :post,
+              local: true,
+              id: 'azure-oauth-login-form' do |form| %>
+  <noscript>
```
jlandiseigsti, line 5: Good thinking to include this just in case.

jlandiseigsti (approved):
Looks good to me!

PR #2030, opened 2026-06-23
Handle open access workflow when user changes `work_type`

ajkiessl:
Context: A user changed their work type _after_ checking the Open Access checkbox and saving, making the upload an Open Access upload despite not being eligible.  This was breaking things.

Fix: Adds helper method for determining if a work version's type is eligible for the open access workflow, and using that, a helper method for determining if the open access flow should be active for a work version.  Also, reset open_access_upload and open_access_version if the user changes the work type to something that is not eligible.  This will prevent issues when users change work types.


Smullz622 (approved):
LGTM

PR #2031, opened 2026-06-23
Disable open access workflow checkbox if open_access_upload is already set

ajkiessl:
This will make sure the checkbox cannot be unchecked if the user returns to the "Work Type" page after selecting the open access workflow.


Smullz622 (approved):
lgtm!

PR #2032, opened 2026-06-24
Send most recent published version that is not remediated for curation

ajkiessl:
When a work version is created _and_ remediated in the same day, nothing gets sent to AirTable because the most recent published version is a remediated version.  These changes make it so the most recent published version that is not remediated is sent to AirTable.


jlandiseigsti (approved):
Looks good to me!

PR #2033, opened 2026-06-30
use best permission data when that version matches the current version

Smullz622:


ajkiessl (dismissed):
LGTM

Maybe for a future refactor: I think the `all_permissions` method could be made private, and then we can use `OawPermissionsService#permissions_found?` instead of `@permissions&.all_permissions.present?` publicly.  But I just noticed this at a glance, so there may be more to it.

ajkiessl (approved):
LGTM

PR #2034, opened 2026-07-10
Use boolean cast on publish param to handle string and boolean types

ajkiessl:
I noticed the Activity Insight Open Access Uploads were failing in RMD.  They actually weren't completely failing, they were just creating drafts instead of publishing the works.  Looks like the way the `publish` param is being handled is the issue.  When `true`, the param gets pulled out as `true` not `'true'`, so the boolean check was returning `false` instead of `true`.  These changes fix that so it can handle booleans and strings acting as booleans.


jlandiseigsti (approved):
LGTM

PR #2039, opened 2026-07-13
Adds ExternalApp registry for Researcher Metadata Database

ajkiessl:
There were two issues with the Scholarsphere --> RMD webhook interaction for OA uploads.  These are fixed in this PR and in this PR in the RMD repo: https://github.com/psu-libraries/researcher-metadata/pull/1290.  Issues:

1.) We sometimes send publications from RMD to Scholarsphere in the OA workflow that don't have DOIs.  These aren't flagged as `open_access_uploads` in Scholarsphere since we can't get permissions data for these.  I had the export from Scholarsphere to RMD filtering out things that are not `open_access_uploads` so these were not updating in RMD.  I fixed this by changing the logic to send the update back to RMD if the work version was uploaded by RMD.

2.) Users will sometimes delete the draft in Scholarsphere after it has been created by RMD.  We need a way to let RMD know the `ScholarsphereWorkDeposit` needs to be deleted.  Most of the work for that is in https://github.com/psu-libraries/researcher-metadata/pull/1290.

I added RMD to the `ExternalApp` registry in this PR.  It probably should have been that way from the start, but now that these to applications are much more coupled, this make things more concrete.  Also, I refactored the webhook services to make them share some of their code.  This involved changing the 'published' export so it sends full urls (instead of the path).  Changes were made in https://github.com/psu-libraries/researcher-metadata/pull/1290 to accommodate this.


app/services/webhooks/base_webhook.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Webhooks
+  class BaseWebhook
```
jlandiseigsti, line 4: Appreciate this refactoring!

jlandiseigsti (approved):
Looks good to me!

PR #2041, opened 2026-07-20
Upgrade to rails v8 and bot challenge v1

### LINKED ISSUE #2040: Upgrade to Rails v8
v8.1.3 is the newest version.  Note that apps using bot_challenge_page will likely need to upgrade to bot_challenge_page v1 to be compatible with Rails 8

ajkiessl:
closes #2040 

Some noteworthy changes in Rails 8.1 that affect ScholarSphere.  I actually think both of these were changes in v7.1, but we were still loading v7.0 config defaults before moving to v8.1:

- Rails decodes dates stored in Postgres to `Date` objects instead of strings: https://guides.rubyonrails.org/configuring.html#config-active-record-postgresql-adapter-decode-dates
- When multiple ActiveRecord commit callbacks are called for the same record, Rails will run the "freshest" rather than the first callback.  This prevents stale callbacks: https://guides.rubyonrails.org/configuring.html#config-active-record-run-commit-callbacks-on-first-saved-instances-in-transaction


jlandiseigsti (approved):
This LGTM

PR #2053, opened 2026-07-29
Fix memoization bug in changed work version validator

ajkiessl:
This was causing the seed command to fail.

@jrpatterson @AndrewGearhart The `bundle exec rake db:seed` command can be used to seed the uldev preview and main deployments from an empty database.  This command will also create placeholder files for the records in S3.  These records should also reindex solr as they are created, so no need to run the whole solr reindex.

I think this is pretty close to what we do with ETDA Explore.

I also added in the changes to the mailer configuration.  Replacing the environment variable setup with the out-of-the-box Rails 8 defaults.

### DIFF
--- a/app/validators/changed_work_version_validator.rb
+++ b/app/validators/changed_work_version_validator.rb
@@ -17,6 +17,7 @@ class ChangedWorkVersionValidator < ActiveModel::Validator
   private
 
     def identical?
+      previous_version = find_previous_version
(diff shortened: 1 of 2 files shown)


ajkiessl:
Ticket to enrich seed data in the future: #2054

jlandiseigsti (approved):
LGTM

PR #2060, opened 2026-07-30
Allow regular users to edit permissions data for open access uploads

### LINKED ISSUE #2037: Depositor should be able to override auto-filled embargo and related data during deposit process
# Bug Report

## Description
When the deposit form for OA policy articles was in RMD, the user had the opportunity to override the auto-filled embargo date, license, "Publisher Statement", and possibly other fields as well (though definitely not DOI). The user should still have that override ability during the deposit process in ScholarSphere. Currently only ScholarSphere admins have that ability.

A key insight from our Teams chat courtesy of @ajkiessl 

> Allowing regular users to edit those autopopulated fields could lead to confusion if they change the open access version and suddenly their selections are overwritten.

ajkiessl:
closes #2037 

Preview branch isn't working, but here's a demo:

https://github.com/user-attachments/assets/27caff7d-043e-44a2-b63d-597f7d740a2a

The permissions fields are editable by non admins now, and when the fields are updated from the open access version changing, a message tells the user which fields changed and the form boxes are highlighted.

Also, I refactored `‎app/javascript/controllers/open_access_version_controller.js` since it was getting large.


app/javascript/controllers/open_access_version/form_updater.js
```diff
@@ -0,0 +1,137 @@
...
+    if (embargoChanged) {
+      changedFields.push(this.data.get('embargoLabel'))
+      this.#markFieldAsUpdated(embargoInput)
+    }
+
+    this.#showFieldUpdates(changedFields)
+    this.#updateVersionMessage(version)
+  }
+
+  #updateInputValue(input, value) {
```
jlandiseigsti: This function method name makes it sound like the function itself is changing the values, rather than a check that returns returns a boolean. Maybe something like #checkInputValue or #isValueChanged?

jlandiseigsti (approved):
I have one naming suggestion for a method, but otherwise this looks good to me.

jlandiseigsti (approved):
LGTM!

PR #2074, opened 2026-08-13
Security Vulnerabilities and remove GraphQL

### LINKED ISSUE #1752: Do we still need GraphQL?
Can this be removed?  Is anyone using it?

Investigate why this was added, who was using it, and who still uses it.

### LINKED ISSUE #2073: security vulns

ajkiessl:
closes #1752 
closes #2073

The `graphiql` packages were a major contributor to security vulnerabilities.  As far as I know, nobody uses the GraphQL features – both the `/api/public` programmatic endpoint and the `/graphiql` GUI.  Nobody has used these for years.  I decided to remove this feature in the PR because it is now more trouble to maintain than its usage.  I checked with the Scholarsphere/RMD teams to see if any of them use it.  So far I have one "no", but waiting for the rest.  None of our internal systems use it.

Redocly/cli also had a handful of vulnerabilities, but I wasn't able to get all of them cleared up.  This is a dev dependency, though, and doesn't actually integrate with anything in our app.  So, it's low priority, but I think I got most of the dependency patches for it.

I also removed a couple of yarn packages that had vulnerabilities, but were not being used.

Also, another note: I was able to clear up all the ruby vulns with dependabot.  So, there aren't any gem updates in this PR.  However, I did remove the `graphql` gem since I'm removing that feature.


jlandiseigsti (approved):
Looks good to me!

PR #2081, opened 2026-09-16
Prevent Open Access Version Guessing if version is already determined

ajkiessl:
Adds logic to `OpenAccessVersionGuesserJob` to prevent updating the open access version when the version is already determined.  This will prevent overwrites if the guessing takes longer that the GUI selection process.

I also increased the timeout for the spinner on the publish page that shows while the open access version is being determined.  I bumped it to 30 seconds.

Also, the minio docker image is no longer hosted on Dockerhub. We need to pull it from quay.io now. Quay.io is a reputable docker image repository managed by RedHat. I changed the minio definitions in the docker-compose and circle ci configs and updated some things due to deprecations.


jlandiseigsti (approved):
LGTM

